PR #18, opened 2019-06-26
specify initial version and PyPI name

steven-murray:
Set version to 3.0.0dev (to be bumped to 3.0.0 when we are happy with the overall API) and set PyPI name to 21cmFAST.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -16,8 +16,11 @@
 
 **A semi-numerical cosmological simulation code for the radio 21cm signal.**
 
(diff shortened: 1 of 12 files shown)


andreimesinger:
sounds good

steven-murray:
@andreimesinger you'll need to click the "Review" button and click "approve", otherwise it doesn't let me merge! It's GitHub's way of making sure I don't submit something stupid.

PR #24, opened 2019-09-23
Photon cons

### LINKED ISSUE #33: undefined f_coll_min in IonisationBox.c
undefined f_coll_min in IonisationBox.c. I remember we were calculating it using Z_HEAT_MAX in the older version. Please check if this feature needs to be removed or added back.

BradGreig:
Merging in of photon non-conservation feature branch.
Closes #33 .


src/py21cmfast/src/IonisationBox.c
```diff
@@ -98,6 +103,77 @@ LOG_SUPER_DEBUG("defined parameters");
...
+        }
+        
+        // keep the original sampled redshift
+        stored_redshift = redshift;
+        
+        // This redshift snapshot now uses the modified redshift following the photon non-conservation correction
+        redshift = adjusted_redshift;
+    }
+    
+
```
steven-murray: Can you momve all this to its own function?

src/py21cmfast/src/IonisationBox.c
```diff
@@ -666,7 +758,13 @@ LOG_ULTRA_DEBUG("while loop for until RtoM(R)=%f reaches M_MIN=%f", RtoM(R), M_M
             rec = 0.;
         
             xHI_from_xrays = 1;
-            Gamma_R_prefactor = pow(1+redshift, 2) * (R*CMperMPC) * SIGMA_HI * global_params.ALPHA_UVB / (global_params.ALPHA_UVB+2.75) * N_b0 * ION_EFF_FACTOR / 1.0e-12;
+            if(flag_options->PHOTON_CONS) {
+                // Used for recombinations, which means we want to use the original redshift not the adjusted redshift
+                Gamma_R_prefactor = pow(1+stored_redshift, 2) * (R*CMperMPC) * SIGMA_HI * global_params.ALPHA_UVB / (global_params.ALPHA_UVB+2.75) * N_b0 * ION_EFF_FACTOR / 1.0e-12;
```
steven-murray: try reducing duplicate code

src/py21cmfast/wrapper.py
```diff
@@ -2139,6 +2186,73 @@ def run_lightcone(
         else max_redshift
     )
 
+    # Perform the calibration and initialisation for the photon non-conservation correction
+    if flag_options.PHOTON_CONS is True:
```
steven-murray: just `flag_options.PHOTON_CONS:`

src/py21cmfast/wrapper.py
```diff
@@ -2139,6 +2186,73 @@ def run_lightcone(
...
+
```
steven-murray: Let's move this to its own function

src/py21cmfast/wrapper.py
```diff
@@ -903,6 +901,71 @@ def calc_zstart_photon_cons():
     return lib.ComputeZstart_PhotonCons()
 
 
+
+def access_photon_nonconservation_data():
```
steven-murray: I think the name for this function should be `get_photon_nonconservation_data`... "access" sounds like its being stored globally somewhere and you're just providing a window onto it, however this function actually calculates and returns the data.

This also needs a docstring!

src/py21cmfast/wrapper.py
```diff
@@ -2009,6 +2085,12 @@ def __init__(
         node_redshifts=None,
         global_xHI=None,
         global_brightness_temp=None,
+        analytic_z = None,
```
steven-murray: can we put  all of these into a dictionary called something like `photon_nonconservation_data`? Also name the variables with `snake_case`, not `CamelCase`.

src/py21cmfast/wrapper.py
```diff
@@ -2415,3 +2448,80 @@ def _get_lightcone_redshifts(
             for d in lc_distances
         ]
     )
+
+def _CalibratePhotonCons(
```
steven-murray: can we rename this `calibrate_photon_cons`?

src/py21cmfast/wrapper.py
BradGreig: Well technically these are now globals as it'll allow easier debugging of the gsl errors. Nevertheless, I have renamed the function and added the docstring

src/py21cmfast/wrapper.py
BradGreig: Concatenated this into a single dictionary. Note, this now adds an extra returned argument to the run_coeval function which will need to be appropriately added to the examples etc.

src/py21cmfast/wrapper.py
BradGreig: Done

BradGreig:
Testing for this should only be an extension of one of the tests you were making. Thus, I'll leave that for you to add

PR #34, opened 2019-10-03
Relvel

JulianBMunoz:
Added a new transfer function option (POWER_SPECTRUM = 5), which generates density initial conditions from a CLASS input table, and tests for it.


src/py21cmfast/src/ps.c
```diff
@@ -360,41 +545,49 @@ double init_ps(){
...
     TFset_parameters();
-    
+
     sigma_norm = -1;
-    
+
     R = 8.0/cosmo_params_ps->hlittle;
-    kstart = 1.0e-99/R;
-    kend = 350.0/R;
+    kstart = FMAX(1.0e-99/R, KBOT_CLASS);
```
steven-murray: let's wrap that in an if statement

src/py21cmfast/src/ps.c
```diff
@@ -406,65 +599,86 @@ double init_ps(){
...
     return R_CUTOFF;
 }
 
 
+
+
+//function to free arrays related to the power spectrum
+void free_ps(){
+
+	double temp_var;
```
steven-murray: let's remove the variable as it is unused

JulianBMunoz:
I joined both transfer functions for CLASS into TF_class, with an extra flag_dv = 0 for outputting densities and = 1 for relative velocities.

PR #44, opened 2019-10-07
Relvel

JulianBMunoz:
Finalized adding the relative velocity boxes, which now exist in both the c code and the python wrapper.

Wrote a test to make sure the rms of vcb is ~30 km/s at recombination and that the average is smaller than the rms, as expected.


src/py21cmfast/src/GenerateICs.c
```diff
@@ -170,6 +196,70 @@ int ComputeInitialConditions(unsigned long long random_seed, struct UserParams *
...
+    fftwf_execute(plan);
+
+  //and filter each box:
+    if (user_params->DIM != user_params->HII_DIM){
+      filter_box(HIRES_box_vcb_x, 0, 0, L_FACTOR*user_params->BOX_LEN/(user_params->HII_DIM+0.0));
+      filter_box(HIRES_box_vcb_y, 0, 0, L_FACTOR*user_params->BOX_LEN/(user_params->HII_DIM+0.0));
+      filter_box(HIRES_box_vcb_z, 0, 0, L_FACTOR*user_params->BOX_LEN/(user_params->HII_DIM+0.0));
+    }
+
+  //to save into a lowres python box
```
steven-murray: this is not really a "python" box -- it's a C struct. But it is the thing we can easily access in python :-)

JulianBMunoz:
The lowrest_vcb box was being showed in Fourier space, which is nonsense. Now it's fixed and in real space. Plot shows the hires and lowers vcb in a box with HII_DIM=100 and L=300 Mpc.
Screen Shot 2019-10-08 at 2 32 09 PM

PR #47, opened 2019-10-22
fixed 2LPT bug

### LINKED ISSUE #22: [BUG] 2LPT has preferred direction
**Describe the bug:**

I am unclear as to the precise nature of this bug -- @jaehongpark00 can you comment on this?

jaehongpark00:
Fixed 2LPT bug.

Fixes #22


BradGreig:
The FFT normalisation was incorrectly applied for each component of the 2LPT component. It should only be applied to the final sum. This was resolved in Pisa, I don't think there is any need for any accompanying figures.

PR #54, opened 2019-11-20
Memleak

steven-murray:
An in-depth fix for memory leaks, checked with valgrind. Mostly the problem was free'ing of FFTW plans. 

Fixes https://github.com/21cmfast/21CMMC/issues/4


steven-murray:
Thanks @BradGreig , you're right -- I fixed it now.

Issue #52, opened 2019-11-04
method for accessing coeval data different and not documented correctly [BUG]


catherinewatkinson:
Not so much a bug as something I think needs rethinking/revising.

I am just trying to run using run_coeval using some existing code with the most up-to-date 21cmFAST and I am having to work out how to access the boxes all over again by trawling through the code itself. The new method using a Coeval class containing a load of other classes is returning an Class object containing classes of classes for each coeval type. Unlike before where it returned the individual coeval classes, each containing a list of the coeval at each redshift were returned. 

I have a couple of issues with this:
1. Could we perhaps decide on the higher level interfacing ASAP and stop changing it until the next major release (and even then only if absolutely essential). It is incredibly irritating to have to waste time debugging established code. This made worse by the fact the documentation doesn't appear to be keeping up with these high-level tweaks . I also had something similar with the power spectrum noisefile changing what string it used to access the dictionary contents. Which I adjusted code to allow for and now it has changed back to the old method so I have to do this all over. #wingeover ;)

2. It seems quite convoluted to have to use nested classes to get to the coeval data. i.e. 
coevals = p21.run_coeval( redshift = redshifts, user_params = UParams, astro_params=AParams, flag_options=FOptions, write=False)

box = coevals[0].brightness_temperature.brightness_temp

I've just also realised, 21cmMC is not compatible with the new coeval class method.

steven-murray:
Thanks @catherinewatkinson. I admit that the `Coeval` class addition was my fault. I do happen to think it's better than what was before, but you are quite right to highlight the irritating nature of backwards-incompatible API changes. 

Going forward, the versioning scheme should make this issue disappear -- API-breaking changes are _only_ allowed for major version changes (i.e 2.x.x --> 3.0.0). However, we are working towards our 3.0.0 currently (we're on 3.0.0alpha), and so we need to decide on the high-level API before we switch to 3.0.0. Until then, unfortunately, the API will not be set in stone (we need it to be relatively fluid now, otherwise we won't be able to get it as right as we want for that first proper release).

In that vein, what do you suggest for the API here? What is the most obvious output for you? For me, the downside of the previous output was that if we wanted to pass some other data back from `run_coeval`, it would instantly break the API again. Now, if whatever we pass back is part of the `Coeval` object, it doesn't break anything to break the internals of that object. 

The `21cmMC` calls are fixed in the current `memory-leak` branch over there, which should be merged very soon.

[...]

PR #57, opened 2019-11-27
Make Coeval nicer to work with

steven-murray:
An attempt to make the new `Coeval` class a bit nicer to work with.
Addresses #52 .

Inviting feedback on the API!


steven-murray:
I'm going to wait for @catherinewatkinson to lend her approval before moving on this :-)

PR #64, opened 2020-01-03
More rigorous exceptions

### LINKED ISSUE #16: [BUG] GSL Error for Specific Parameters
This is a copy of part of https://github.com/BradGreig/Hybrid21CM/issues/30.

**Describe the bug:**

@BellaNasirudin reported receiving the following error:

```
2019-04-08 11:24:12 | INFO    | UsefulFunctions.c | writeAstroParams:553 [pid=31775] | AstroParams: [HII_EFF_FACTOR=13.039708, ION_Tvir_MIN=46699094016.000000, X_RAY_Tvir_MIN=46699094016.000000, R_BUBBLE_MAX=15.000000, L_X=3.532312e+38, NU_X_THRESH=273.539978, X_RAY_SPEC_INDEX=1.000000, F_STAR10=0.050119, F_STAR=0.500000, N_RSD_STEPS=0.000000]
2019-04-08 11:24:12 | INFO    | UsefulFunctions.c | writeFlagOptions:559 [pid=31775] | AstroParams: [USE_MASS_DEPENDENT_ZETA=0, SUBCELL_RSD=0, INHOMO_RECO=0, USE_TS_FLUCT=0]
2019-04-08 11:24:12 | SUPER-DEBUG | IonisationBox.c | ComputeIonizedBox:69 [pid=31775] | defined parameters
2019-04-08 11:24:12 | SUPER-DEBUG | IonisationBox.c | ComputeIonizedBox:136 [pid=31775] | erfc interpolation done
2019-04-08 11:24:12 | SUPER-DEBUG | IonisationBox.c | ComputeIonizedBox:179 [pid=31775] | density field calculated
2019-04-08 11:24:12 | SUPER-DEBUG | IonisationBox.c | ComputeIonizedBox:210 [pid=31775] | minimum source mass has been set: 1086022819126444032.000000
2019-04-08 11:24:12 | SUPER-DEBUG | IonisationBox.c | ComputeIonizedBox:216 [pid=31775] | sigma table has been initialised
2019-04-08 11:24:12 | ULTRA-DEBUG | ps.c            | FgtrM_General:862 [pid=31775] | integration range: 41.529054 to 46.134224
gsl: qag.c:261: ERROR: could not integrate function
Default GSL error handler invoked.
/var/spool/slurm/job2979243/slurm_script: line 21: 31775 Aborted      
```
 for a run with the following parameters:
```
HII_EFF_FACTOR = 13.03970787 
ION_Tvir_MIN = 10.66930846
L_X = 38.54805908
NU_X_THRESH = 273.53996706
```

**To Reproduce:**
Bella obtained the original error during an MCMC run, but got the error also in a single-box run for these particular parameters. Other parameters of the run were not given.

**Expected behavior:**

Either a correct result, *or* a more useful error message.

steven-murray:
This PR adds exception handling to C. 

The basic premise is that any non-API function (i.e. any function that's not in `21cmFAST.h`) should never just return an int to signal that it broke. Instead, it should do a `Throw <exception_kind>;`. 
The new `cexception.h` file includes the `Throw` macro. If this happens within a `Try {} Catch {}` block, the execurtion will jump to the `Catch` (at any level of the call stack). So these really are like `Try; Except; Raise` in Python. 

Any API-level function in C should be pretty much globally wrapped in a `Try Catch`, and if the `Catch` is activated, should return the error code to Python, which will then deal with it accordingly.

Closes #16.

This will go a long way to solving issues like #16 . In fact, this may solve that issue entirely, though I'd like @BradGreig to have a look first and see if something wrong is not happening in the code. If it is just that we don't expect those parameters ever to work, then this fixes that issue.

### DIFF
--- a/docs/notes_for_developers.rst
+++ b/docs/notes_for_developers.rst
@@ -26,7 +26,11 @@ below about running with valgrind). When changing C code, before
 testing, ensure that the new C code is compiled into your environment by running::
 
     $ rm -rf build
-    $ pip install -e .
+    $ pip install .
+
+Note that using a developer install (`-e`) is not recommended as it stores compiled
+objects in the working directory which don't get updated as you change code, and can
+cause problems later.
 
 There are two main purposes you may want to write some C code:
 
@@ -89,6 +93,23 @@ changes in the Python are necessary. However, you may want to consider adding th
 parameter to relevant ``_filter_params`` lists for the output struct wrapping classes in
 the wrapper. These lists control which global parameters affect which output structs,
 and merely provide for more accurate caching mechanisms.
+
+C Function Standards
+~~~~~~~~~~~~~~~~~~~~
+The C-level functions are split into two groups -- low-level "private" functions, and
+higher-level "public" or "API" functions. All API-level functions are callable from
+python (but may also be called from other C functions). All API-level functions are
+currently prototyped in `21cmFAST.h`.
+
+To enable consistency of error-checking in Python (and a reasonable standard for any
+kind of code), we enforce that any API-level function must return an integer status.
+Any "return" objects must be modified in-place (i.e. passed as pointers). This enables
+Python to control the memory access of these variables, and also to receive proper
+error statuses (see below for how we do exception handling). We also adhere to the
+convention that "output" variables should be passed to the function as its last
+argument(s). In the case that _only_ the last argument is meant to be "output", there
+exists a simple wrapper `_call_c_simple` in `wrapper.py` that will neatly handle the
+calling of the function in an intuitive pythonic way.
 
 Running with Valgrind
 ~~~~~~~~~~~~~~~~~~~~~
--- a/docs/notes_for_developers.rst
+++ b/docs/notes_for_developers.rst
@@ -134,3 +134,50 @@ Producing Integration Test Data
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 There are bunch of so-called "integration tests", which rely on previously-produced
 data. To produce this data, run `python tests/produce_integration_test_data.py`.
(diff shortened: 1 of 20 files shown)


andreimesinger:
having a virial temperature of 10^11 K is off the scale…  this is larger that superclusters, and i am sure whatever interpolation is done in the code will break on these scales.

steven-murray:
I just saw @BradGreig 's comment on #16 to the effect that those parameters are just too extreme, and we don't expect to be able to handle them. Oh, and @andreimesinger 's. 

I'll add a warning into the Python part when setting the parameter value to make it easier to diagnose for users. And then I think this solves that issue as well.

qyx268:
This is a great PR! It will make debugging much easier...

I just have a minor point. It happened to me when 21cmmc crashed and returned

`BrokenProcessPool exception (most likely an unrecoverable crash in C-code).`
  `Due to the nature of this exception, it is impossible to know which of the following parameter vectors were responsible for the crash. Running your likelihood function with each set of parameters in serial may help identify the problem. `                                                          

I think it would be great if this exception can also tell the user about the parameters that is causing it.

steven-murray:
@qyx268 : thanks for the report. It would definitely be nice to know the parameters, but unfortunately in that case it is quite impossible. It happens when running with multiple processors and a Really Bad Error happens in C, which Python can't do anything about, and the whole thing crashes. Then all the parameters that are currently being run (in different processes) _could_ be the culprit, and it can't know which, hence the error message.

The only solution to it is to eliminate all possible sources of Bad Error. I.e. catch the errors with the code I've included here, and pass back the errorcodes to Python. But that means catching any potential error by knowing when they might occur.

PR #65, opened 2020-01-03
Contributing

steven-murray:
This PR makes the `CONTRIBUTING.rst` a bit better. 
Now it leads the new developer through the exact process required to get setup (and also serves as a good memory-jog for us others). It also points directly to the developer documentation for more details (those docs also link back the to the contributing). 

I figure it would be good to get @andreimesinger to have a quick read through, as someone who hasn't used the new setup very much, to see if it makes sense and suggest changes.

Fixes #65

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -10,9 +10,9 @@ Bug reports/Feature Requests/Feedback/Questions
 It is incredibly helpful to us when users report bugs, unexpected behaviour, or request
 features. You can do the following:
 
-    * `Report a bug <https://github.com/21cmFAST/21cmFAST/issues/new?template=bug_report.md>`
-    * `Request a Feature <https://github.com/21cmFAST/21cmFAST/issues/new?template=feature_request.md>`
-    * `Ask a Question <https://github.com/21cmFAST/21cmFAST/issues/new?template=question.md>`
+    * `Report a bug <https://github.com/21cmFAST/21cmFAST/issues/new?template=bug_report.md>`_
+    * `Request a Feature <https://github.com/21cmFAST/21cmFAST/issues/new?template=feature_request.md>`_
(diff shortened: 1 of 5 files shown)


CONTRIBUTING.rst
```diff
@@ -44,27 +47,50 @@ pre-commit hooks will run that enforce the ``black`` coding style. If you do not
...
+
+   or if using `conda` (suggested)::
+
+    conda env update -f environment.yml
+
+4. Install pre-commit hooks::
+
+    pre-commit install
+
+3. Create a branch for local development::
```
andreimesinger: step 3 should be step 5

CONTRIBUTING.rst
```diff
@@ -44,27 +47,50 @@ pre-commit hooks will run that enforce the ``black`` coding style. If you do not
...
 
     git checkout -b name-of-your-bugfix-or-feature
 
-   Now you can make your changes locally. **Note: you _must_ do this step. If you
```
andreimesinger: this step -> step 5

CONTRIBUTING.rst
```diff
@@ -44,27 +47,50 @@ pre-commit hooks will run that enforce the ``black`` coding style. If you do not
...
-   make changes on master, you will _not_ be able to push them, as a member**.
+   Now you can make your changes locally. **Note: as a member, you _must_ do this step. If you
+   make changes on master, you will _not_ be able to push them**.
 
 3. When you're done making changes, run all the checks, doc builder and spell checker
```
andreimesinger: numbering seems off

CONTRIBUTING.rst
steven-murray: Ahhh, thanks. Now fixed.

CONTRIBUTING.rst
steven-murray: Yup, screwed that up. Fixed now.

steven-murray:
All ready for quick re-review when you're ready, @andreimesinger

PR #68, opened 2020-01-03
Config

### LINKED ISSUE #10: [Feature Req.] Global Configuration Settings
This is a copy/restatement of https://github.com/BradGreig/Hybrid21CM/issues/6

While at this point, the only configuration option residing in `.21cmfast/config.yml` is `boxdir`, it is possible that in the future, several 'global' options may be set there. In this case, it would be nice to have a way of over-riding these options at a per-session level, without changing the actual file.

I'm not sure of the best way to do this -- but libraries like `matplotlib` do it, and so we should be able to get an idea there.

steven-murray:
A lot of changes to implement configuration management. The way it works is documented in the `misc.rst` file, but basically, you can set global changes to the config (stuff like where the cache is), or use a context manager, i.e. do stuff like 

```python
with p21.config.use(direc="my_new_cache"):
    p21.run_coeval(...)
```

This will set the option back to the default after exiting from the context. You can also write the config back to file to make it permanent.

Close #10 

This is branched off the `doc-updates` branch, so let's get that pulled in before we try to pull this in.

### DIFF
--- a/src/py21cmfast/cli.py
+++ b/src/py21cmfast/cli.py
@@ -666,7 +666,7 @@ def _query(direc=None, kind=None, md5=None, seed=None, clear=False):
             print()
 
         else:
-            direc = direc or path.expanduser(_cfg.config["boxdir"])
+            direc = direc or path.expanduser(_cfg.config["direc"])
             remove(path.join(direc, file))
 
 
--- a/src/py21cmfast/inputs.py
+++ b/src/py21cmfast/inputs.py
@@ -321,7 +321,7 @@ class UserParams(StructWithDefaults):
         2: Watson (Watson FOF)
         3: Watson-z (Watson FOF-z)
     USE_RELATIVE_VELOCITIES: int, optional
-        Flag to decide whether to use relative velocities, default True.
+        Flag to decide whether to use relative velocities.
         If True, POWER_SPECTRUM is automatically set to 5. Default False.
     POWER_SPECTRUM: int or str, optional
         Determines which power spectrum to use, default EH (unless `USE_RELATIVE_VELOCITIES`
--- a/src/py21cmfast/inputs.py
+++ b/src/py21cmfast/inputs.py
@@ -331,7 +331,7 @@ class UserParams(StructWithDefaults):
         2: EFSTATHIOU
         3: PEEBLES
         4: WHITE
-        5: CLASS (single output)
+        5: CLASS (single cosmology)
     """
 
     _ffi = ffi
--- a/src/py21cmfast/wrapper.py
+++ b/src/py21cmfast/wrapper.py
@@ -92,6 +92,7 @@ import numpy as np
 from astropy import units
 from astropy.cosmology import z_at_value
 
+from ._cfg import config
 from ._utils import StructWrapper
 from .c_21cmfast import ffi
 from .c_21cmfast import lib
@@ -295,6 +296,7 @@ def _process_exitcode(exitcode):
 
 
 def _call_c_func(fnc, obj, direc, *args, write=True):
+    logger.debug("Calling {} with args: {}".format(fnc.__name__, args))
     exitcode = fnc(
         *[arg() if isinstance(arg, StructWrapper) else arg for arg in args], obj()
     )
@@ -308,6 +310,14 @@ def _call_c_func(fnc, obj, direc, *args, write=True):
         obj.write(direc)
 
     return obj
+
+
+def _get_config_options(direc, regenerate, write):
+    return (
+        config["direc"] if direc is None else direc,
+        config["regenerate"] if regenerate is None else regenerate,
+        config["write"] if write is None else write,
+    )
 
 
 # ======================================================================================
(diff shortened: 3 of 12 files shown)


src/py21cmfast/inputs.py
```diff
@@ -0,0 +1,614 @@
...
+    """
+
+    _ffi = ffi
+
+    _defaults_ = {
+        "SIGMA_8": 0.82,
+        "hlittle": Planck15.h,
+        "OMm": Planck15.Om0,
+        "OMb": Planck15.Ob0,
+        "POWER_INDEX": 0.97,
```
JulianBMunoz: Why don't we set SIGMA_8 and and POWER_INDEX to the Planck 2015 values?

src/py21cmfast/inputs.py
```diff
@@ -0,0 +1,614 @@
...
+        Determines which halo mass function to be used for the normalisation of the
+        collapsed fraction (default Sheth-Tormen). If string should be one of the
+        following codes:
+        0: PS (Press-Schechter)
+        1: ST (Sheth-Tormen)
+        2: Watson (Watson FOF)
+        3: Watson-z (Watson FOF-z)
+    USE_RELATIVE_VELOCITIES: int, optional
+        Flag to decide whether to use relative velocities, default True.
+        If True, POWER_SPECTRUM is automatically set to 5. Default False.
```
JulianBMunoz: Default true or false? Pick one.

src/py21cmfast/inputs.py
```diff
@@ -0,0 +1,614 @@
...
+    POWER_SPECTRUM: int or str, optional
+        Determines which power spectrum to use, default EH (unless `USE_RELATIVE_VELOCITIES`
+        is True). If string, use the following codes:
+        0: EH
+        1: BBKS
+        2: EFSTATHIOU
+        3: PEEBLES
+        4: WHITE
+        5: CLASS (single output)
```
JulianBMunoz: Maybe it's more accurate to say single cosmology? It's the same though, and only for now!

src/py21cmfast/inputs.py
```diff
@@ -0,0 +1,614 @@
...
+    """
+
+    _ffi = ffi
+
+    _defaults_ = {
+        "BOX_LEN": 150.0,
+        "DIM": None,
+        "HII_DIM": 50,
```
JulianBMunoz: This is a discussion for another time, but I've always found it dubious that the default BOX_LEN is 150 Mpc, since that can be too small to capture all the matter fluctuations (k>0.05 Mpc^-1).

src/py21cmfast/wrapper.py
```diff
@@ -1096,8 +628,8 @@ def perturb_field(
     user_params=None,
     cosmo_params=None,
     random_seed=None,
-    regenerate=False,
-    write=True,
+    regenerate=None,
+    write=None,
```
JulianBMunoz: Does this overwrite line 592?

src/py21cmfast/wrapper.py
```diff
@@ -1228,144 +761,147 @@ def ionize_box(
     init_boxes=None,
     cosmo_params=None,
     user_params=None,
-    regenerate=False,
-    write=True,
+    regenerate=None,
+    write=None,
```
JulianBMunoz: Same as before.

docs/usage.rst
```diff
@@ -1,130 +0,0 @@
...
-following will not re-calculate the init box::
-
-    $ 21cmfast init
-    $ 21cmfast perturb redshift=8.0
-
-This means that almost all of the functionality provided in the library is accessible
-via the CLI.
-
-
-Library
```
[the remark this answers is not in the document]
steven-murray: I removed the usage section about the library because there are numerous tutorials and also full API documentation, which makes it redundant and in danger of growing out of date.

src/py21cmfast/cli.py
```diff
@@ -510,8 +591,28 @@ def coeval(ctx, redshift, config, regen, direc, z_step_factor, z_heat_max, seed)
 def lightcone(
     ctx, redshift, config, regen, direc, max_z, z_step_factor, z_heat_max, seed
 ):
-    """
-    Efficiently generate coeval cubes at a given redshift.
+    """Efficiently generate coeval cubes at a given redshift.
```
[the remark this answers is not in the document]
steven-murray: This is a typo, but it was fixed in what is now master.

src/py21cmfast/inputs.py
steven-murray: We can do that, but it should be in another PR. I'll make an issue for it now.

src/py21cmfast/inputs.py
steven-murray: Woops! Fixed it up.

src/py21cmfast/inputs.py
steven-murray: Yup, changed to "cosmology"

src/py21cmfast/inputs.py
steven-murray: I think that's a fair point. The default is supposed to be really quite small, so that if you fire up a session and just run with defaults, it doesn't take a bazillion years. But I'll pull up a separate issue for it.

src/py21cmfast/wrapper.py
steven-murray: Yes -- these are passed through to each recursive level.

Issue #49, opened 2019-10-23
[BUG] Tests for Photon Conservation


steven-murray:
**Describe the bug:**

The Photon Conservation feature is not currently tested. It should be added to the integration tests suite.

BradGreig:
Closing this as tests were added in #71

PR #71, opened 2020-01-09
Multithreading

### LINKED ISSUE #13: [Feature Req.] Multithreading
This is a copy/restatement of https://github.com/BradGreig/Hybrid21CM/issues/28.

In the current version, there is no multi-processing/threading in `21cmFAST`, which is due to this version being based on `Hybrid21cm`, which favored multi-processing in the MCMC, rather than on a per-simulation basis. This option should be added back in, though.

It should be possible to set an argument `nthreads` (or something like it) to dynamically change the number of threads used in the `C` code. This could probably go into `user_params`.

### LINKED ISSUE #36: [BUG] FFTW wisdom creation in generateICs.c
**Describe the bug:**

In lines 159-183 of generateICs.c, when it creates the FFTW wisdom it erases the HIRES_box, so it has to be copied from the _saved version. However, HIRES_box was filtered and HIRES_box_saved was not filtered, which seems inconsistent.

**To Reproduce:**
1. ...
2. ...

**Expected behavior:**

**Details:**
 - Python Version: [e.g. 22]
 - Relevant compiler versions:
 - Relevant package versions:

**Additional context**

### LINKED ISSUE #53: Filter_box in UsefulFunctions.c produces segfault running under intel compiler [BUG]
In attempting to get the code to run on the HPC at Imperial, I experienced segfaults having compiled with the Intel ICC compiler.

The source of the segfault is the for loops in filter_box (UsefulFunctions.c), which take the form (L100):
(n_x=dimension; n_x--;)

Replacing this and the corresponding one for (n_y=dimension; n_y--;) with:
(n_x=(dimension-1);  n_x>=0; n_x--)

resolves the problem.

BradGreig:
An intermediate stage for the multiprocessing branch (complete, but will need to be updated with the #37 branch before merging).

Fixes #13 
Also fixes #53 
Fixes #36


src/py21cmfast/src/GenerateICs.c
```diff
@@ -47,44 +47,82 @@ int ComputeInitialConditions(unsigned long long random_seed, struct UserParams *
...
     // Removed all references to threads as 21CMMC is always a single core implementation
 
+    int checker;
+    
+    checker = 0;
     // seed the random number generators
-    r = gsl_rng_alloc(gsl_rng_mt19937);
-    gsl_rng_set(r, random_seed);
+    for (thread_num = 0; thread_num < user_params->N_THREADS; thread_num++){
+        switch (checker){
```
freddavies: Rather than cycle through the random number generators in this linear way, which could lead to some residual correlation between threads that use the same generator, a safer way to do this is to generate the individual thread seeds with a random number generator from the initial seed and shuffle them. See https://rpubs.com/Jouni_Helske/225931 for details. 

My rough implementation in the old 21cmFAST looked like this:
```
int num_int = INT_MAX/16;
unsigned int *many_ints = (unsigned int *)malloc((size_t)(num_int*sizeof(unsigned int))); // some large number of possible integers
for (i=0; i<num_int; i++) {
    many_ints[i] = i;
}
gsl_rng * rseed = gsl_rng_alloc(gsl_rng_mt19937);
gsl_rng_set(rseed, RANDOM_SEED);
unsigned int seeds[NUMCORES];
gsl_ran_choose(rseed, seeds, NUMCORES, many_ints, num_int, sizeof(unsigned int));
gsl_ran_shuffle(rseed, seeds, NUMCORES, sizeof(unsigned int));
for (thread_num=0; thread_num < NUM_RNG_THREADS; thread_num++) {
    r[thread_num] = gsl_rng_alloc(gsl_rng_ranlxs2);
    gsl_rng_set(r[thread_num],seeds[thread_num]);
    fprintf(stderr, "Thread #%i will use a ranlxs2 RNG seeded with %u\n", thread_num, seeds[thread_num]);
}

**Comment by andreimesinger** [comment]
i think the correlations only might appear in very large boxes, but i agree in principle it is better. @freddavies, you could maybe make a quick PR for this at some point

File: src/py21cmfast/src/GenerateICs.c (hunk 1 above)
Comment (BradGreig) [reply src/py21cmfast/src/GenerateICs.c]: Thanks @freddavies! I agree this is a nice way of dealing with this. I originally toyed with having each RNG being seeded by a random product of random numbers. 

I'm not sure what you mean by residual correlations between threads that use the same generator. Each thread uses its own differently seeded RNG so no generator has the same seed.

Comment (freddavies) [reply src/py21cmfast/src/GenerateICs.c]: The blog post shows that if the set of initial seeds is too simply structured (e.g., a linear function of thread ID, as done here), then there are residual correlations between the values generated by the various streams.  The first set of figures shows that there is a non-negligible correlation between the first random values drawn using the set of seeds from 1 to 100000 for a few common generators.

It's likely that this doesn't matter too much, but it's probably better to be safe than sorry. I can put in a PR later as @andreimesinger suggested.

**Comment by steven-murray** [comment]
I've set aside some time to review this on Thursday (Phoenix time), so it would be great to get the conflicts resolved with all the new changes in master (I can also try to do that myself if it's not done before Thursday). 

@freddavies : I think it makes sense to implement your idea as a separate PR which can be applied after this is merged. Can you create a specific issue for it, to remind us that it has to be done? Once this is merged, we'll probably forget all about it :-)

File: src/py21cmfast/src/PerturbField.c [hunk 2]
```diff
@@ -102,103 +120,123 @@ int ComputePerturbField(float redshift, struct UserParams *user_params, struct C
...
+                        zi = zf;
+                        if (xi >= (user_params->HII_DIM)){ xi -= (user_params->HII_DIM);}
+                        if (xi < 0) {xi += (user_params->HII_DIM);}
+                        if (yi >= (user_params->HII_DIM)){ yi -= (user_params->HII_DIM);}
+                        if (yi < 0) {yi += (user_params->HII_DIM);}
+                        if (zi >= (user_params->HII_DIM)){ zi -= (user_params->HII_DIM);}
+                        if (zi < 0) {zi += (user_params->HII_DIM);}
+                    
+                        *( (float *)LOWRES_density_perturb + HII_R_FFT_INDEX(xi, yi, zi) ) +=
+                            (1 + init_growth_factor*(boxes->hires_density)[R_INDEX(i,j,k)]);
```
freddavies: One more thing (sorry @steven-murray !)...

I believe this line is technically a race condition if `LOWRES_density_perturb` is written to in the same location in memory by multiple threads at the same time (i.e. multiple hires cells being displaced to the same lowres location at the same time). I don't remember exactly what the consequences were (segfaults, probably?), but I do remember it caused me many headaches in the past in my old parallel implementation of 21cmFASTv1.

My workaround was to place the `omp  for` pragma at the `j` loop instead, store the computed `HII_R_FFT_INDEX(xi, yi, zi)` (at fixed `i`) in a separate 2D array since that is the most computationally intensive part, and then (serially) add the displaced mass to the 3D field. I'm sure there is a better way to do it with OpenMP structures, though.

steven-murray: Perhaps this could be fixed by an `atomic` statement before the line?

src/py21cmfast/src/GenerateICs.c
BradGreig: @freddavies ah, sorry. I didn't actually get a chance to look at the linked article.

src/py21cmfast/src/PerturbField.c
BradGreig: @steven-murray Yes, I think an atomic statement would suffice, but I can think about it a little further. 

Nice catch @freddavies! In this instance, no segfaults would arise, I think just the cells that incur the behaviour would have less density than they should.

BradGreig:
@steven-murray if you are intending to merge master back into this branch, then it makes sense for me just to add in these fixes (seeding and race condition) after merge.

src/py21cmfast/src/PerturbField.c
BradGreig: @steven-murray an atomic statement does fix this issue. Theoretically though this can cause rounding errors, owing to the fact that it is a float operation. However, you need to be down-sampling quite significantly to incur this. Down-sampling by > 20 did it, though its only at the 3rd-4th decimal place. As soon as it goes to the nominal factor of 3-4 its completely gone. 

The only true solution would to have this down-sampled grid allocated as a double and then convert to float. It would increase the memory, but not substantially. However, I am not sure its worth doing.

Actually, I might just go ahead and do it.

src/py21cmfast/src/GenerateICs.c
```diff
@@ -47,44 +47,82 @@ int ComputeInitialConditions(unsigned long long random_seed, struct UserParams *
...
-    gsl_rng * r;
+    gsl_rng * r[user_params->N_THREADS];
+    
+    omp_set_num_threads(user_params->N_THREADS);
+    fftwf_init_threads();
+    fftwf_plan_with_nthreads(user_params->N_THREADS);
 
     // ************  INITIALIZATION ********************** //
 
     // Removed all references to threads as 21CMMC is always a single core implementation
```
steven-murray: Can probably remove this comment ;-)

steven-murray:
Note to all involved in this PR: I've rebased it onto master, so it's ready for adding those extra things. However, you'll  have to do `git pull --rebase` in order to work on it. 

I'll review again after changes have been implemented.

steven-murray:
@BradGreig what's the status on this? I think this is the next big thing to pull in.

BradGreig:
Hi @steven-murray, this has been waiting for the `mini-halos` branch to be fully tested and merged. As far as I understand `mini-halos` was re-opened as there were still issues requiring resolving. If it is all ready to go, then I can to a merge and update accordingly

steven-murray:
@BradGreig I think `mini-halos` is needing to do a full rebase onto master (which might be a bit painful -- I'm happy to help @qyx268 out with this). @qyx268 has said that he's finding it hard to find time to do. 

In that case, if this is ready to go, should we just merge this, and deal with the extra bit of rebasing required for `mini-halos`? 

There were a few extra things that @freddavies suggested adding: have you put these in?

BradGreig:
Hey @steven-murray, ok, I can update this and fix a couple of the issues. Hopefully will get time to do it before the end of the week.

qyx268:
Hey guys, sorry for making you guys wait. I'm not sure if I can finish minihalo by next week. So if you guys want to merge this, please do so. I'll deal with the conflict later (very likely with your help though)...

steven-murray:
@BradGreig I think this is now the next thing to be merged, but you'll have to do a rebase first (I think a rebase is better than a merge in this case).

BradGreig:
Ok, I'll try and get to this as soon as I can. I wasn't worried about the tests yet, since it needs all the changes in master for using the `write` keyword properly.

steven-murray:
Yup, I think once it's rebased it'll all pass.

BradGreig:
@steven-murray, the integration tests are failing. I'll need to explore whats changed between the two versions (i.e. between the rebased version and the version where the test data was originally computed). I suspect it could be to do with the initial conditions, now that the seeding has changed. But, I'll explore that remotely.

BradGreig:
I should say that at this point I've only done the rebase. Need to pull in all the mini-halo stuff etc.

steven-murray:
The rebase should have pulled in the mini-halo stuff? Did you `git pull` on
master before rebasing?

qyx268:
no, the mini-halo didn't rebase, i simply merged...

steven-murray:
Mini-halos is now merged into master. So this branch has to be rebased onto master:

```
git checkout master
git pull
git checkout Multithreading
git rebase master
```

Then follow the series of conflicts that will arise and fix them. If you've never done a rebase before, it can be kind of strange. But basically, you just fix the conflicts (marked in the usual git fashion), `git add` the file you've fixed, then do `git rebase --continue` and it will bring up more conflicts (it just applies each of the commits in the Multithreading branch on top of the current master successively).

I was going to try it myself, but I don't know enough of the details of the mini-halos and multithreading that I'd be sure to get the mix of them right.

BradGreig:
Ok, some weirdness seemed to have happened when attempting the rebase last week. Its picked up all the changes today (may have missed an error at some point). 

Still work in progress, its a bit painful trying to merge the two branches.

BradGreig:
Re-basing is now finalised, I need to go through and add in additional multiprocessing tools now for extra mini-halo stuff.

@steven-murray, this will fail all integration tests. This is purely driven by the differing initial conditions. I have verified that all tests are passed on my local machine (including the mini-haloes) when I revert to the old random seed usage. Therefore, I assume that once I am finished with this branch, I generate all integration test data locally and push all those files in addition to the code changes. Is this correct?

qyx268:
@BradGreig let me know if I can help with the minihalo stuff. I've changed many structures in terms of redefining stuff in python.

steven-murray:
@BradGreig yes, if all the seeds are changed, we'll need to update all the
integration test files. You can just run them locally, as you said.

BradGreig:
@steven-murray, one of the integration tests that previously passed has now failed (without no changes on my end). This has me a little worried, so I'll need to keep digging

steven-murray:
This is really strange. It's failing on py37 but not on py38 or py36. Also the linting is failing -- have you done the `pip install pre-commit; pre-commit install`? It seems to be just the one test that is failing though.

BradGreig:
This also resolves #49.

The issue ended up being an initialisation flag within the Photon non-conservation correction that wasn't being reset upon repeated runs within the Python environment. It would then just use the first set of data. This is all now resolved.

BradGreig:
Also, the tests probably need to be reduced. The `PHOTON_CONS = True` tests do take a while.

PR #73, opened 2020-01-09
Global params

### LINKED ISSUE #60: [Feature Req.] A context manager for global params

I've been thinking for a while about how to best manage the global params. Mostly it is pretty simple: just change a value of an attribute. However, for those used to using python, the concept of truly "global" parameters is a bit alien. You usually want to set parameters just for a given call, and not have them affect anything else. This could be kind of surprising for python users.

One simple way could be to define a context manager on the global params class, which would enable something doing like this:

```python
with global_params.set(ALPHA_UVB=5.5):
    run_coeval(...)
```

After exiting from the context, the parameters are all set back to their default values. 

Another option would be to add keywords to each of the functions, which could change the global parameters _for that function_, and then remembers to set them back to their default value at the end. But that means changing a lot of code, so the context manager is to be preferred.

steven-murray:
Adds a context manager to `global_params`, so you can do stuff like

```
from py21cmfast import global_params, run_lightcone
with global_params.use(Z_HEAT_MAX=20):
    run_lightcone(redshift=7)
```

and afterwards, your `Z_HEAT_MAX` is back to 35. 

Fixes #60 .

Requires waiting for #61 to be merged before merging.


steven-murray:
I have one question for both @BradGreig and @catherinewatkinson before I feel that this is done. 

We talked at some point about whether `z_heat_max` and `zprime_step_factor` should be used in `run_lightcone` etc. Obviously they have to be available to change, but at the moment, they can be changed as keywords in those functions, but the functions themselves simply then change the global params, and remember to change it back at the end. 

However, since this process is now super easy with this context manager, it might make sense to remove those parameters entirely from the functions and just make the user use the context manager. 
This makes everything that little bit cleaner.

The downside to this is that it lowers the visibility of these quite important parameters. If they're in the globals, they're less likely to be known about by new users, and they won't be seen upon reading the more commonly used input parameter structs. Perhaps this is something that can be defeated with documentation though.

What do you think? Should I go ahead and remove them from the functions (and all related documentation)? Or should I first deprecate them and we can remove in v3.1 (really shouldn't, because it's breaking API at that point -- that's why I want to remove them now)? Or leave them in for good?

caw11:
My personal feeling is that core user parameters should all be adjustable in the same way, and I think this is most cleanly done in the call to the function. 

From what I can tell the only downside to having these is that the code internally sets these globally and then reverts them to the default. Given this is a couple of lines of code that once written never need be changed, and that presumably these two actions are near instantaneous, then I see no advantage to making the user experience less straightforward by removing this functionality. 

If my stated understanding of what you are asking is correct, then I think leave them in the function call forever.

steven-murray:
@caw11 thanks for your thoughts! 

I think there are a couple more downsides to leaving them in that I didn't mention. At the moment, the only two we let users manage through the function call itself are `Z_HEAT_MAX` and `ZPRIME_STEP_FACTOR` (because I found that I was changing those ones reasonably often). However, that choice is kind of arbitrary. There are lots more parameters in there that, to be consistent, we should probably also add to the functions. But then we have to also document them, and update them whenever we add more, or remove one. 

One way to perhaps get around this would be to add a `**global_kwargs`` argument to all low- and high-level functions, which would simply pass on its arguments to the context manager, so the user could just do

```
lightcone = run_lightcone(redshift=7, Z_HEAT_MAX=20)
```

In the documentation for `run_lightcone`, we can then _explicitly_ point to the `GlobalParams` documentation, so people know exactly what parameters they can set. This removes a single line from the user's code (the `with global_params.use(...)` line), and means we have to add a single line for each low-level function ourselves, once.

This might be the best compromise, do you think? It would entail changing user's code now to use the capitalized names that are in global_params. 

One other idea would be to use my proposal in #9 to completely eliminate all global params (but that's slated for v4.0.0 as it would seriously break API).

caw11:
Treat them the same as all the other user-defined variables for now I would say. I think simplicity is the most important thing here. You can leave #9 to v4.

BradGreig:
Apologies for the slow response. I would tend towards these being removed as keywords. In my opinion these variables would rarely need to be used by users thus can be hidden in the globals struct. I agree all this should be explicit in the documentation.

Do you want me to go through this PR, or wait for any more changes?

steven-murray:
Thanks @BradGreig. I think using a single `**kwargs` argument that can be explicitly routed to global params probably makes sense. This gives us the chance to naturally document what those parameters are and point users towards them, without us having to actually add them explicitly. 

Let me make a few more changes to this effect then I'll ping you to review, thanks!

steven-murray:
@BradGreig and @catherinewatkinson this is now ready for review, and hopefully a quick merge (i.e. in < 12 hrs preferentially ;-)

PR #78, opened 2020-01-23
Better lightcone

### LINKED ISSUE #39: [Feature Req.] Create more than just brightness_temp lightcones

Currently, the `run_lightcone` function only creates a lightcone out of the brightness temperature. 
While this is usually what people want, one may also want to see a lightcone of `xHI` or even the density field. 

The easiest solution is probably just to specify some reasonable outputs in the lightcone function, and add them to the `Lightcone`.

One could also try specifying exact a list of boxes to save in the lightcone, but this would be a bit brittle because it would be difficult to ascertain which structure they come from. Also, it would mean the `Lightcone` object would not be absolutely defined.

steven-murray:
Fixes #39 

Adds a keyword `lightcone_quantities` to the `run_lightcone` function which lets the user add any (string) quantity they want to turn into a lightcone. If that quantity can be found in an appropriate output struct, and it's an appropriate array, it will be interpolated the same was as the brightness_temp was before. These are exposed in the `Lightcone` object. 

It also adds a similar interface for saving global quantities. 

It should *not* introduce any backwards-incompatible changes (though it does deprecate some things).


BradGreig:
I just tried running this and it failed. Debugging some things fixed some issues, but still haven't managed to get things to work. Some issues I found:

1. The function `get_all_fieldnames` cannot be called as it hasn't been added to the `__init.py__` list.
2. The `outs` dictionary in `run_lightcone` has "PerturbField" rather than "PerturbedField".
3. I tried turning on all light_cone available fields to construct lightcones but it fails. The error I keep receiving is `TypeError: tuple indices must be integers or slices, not str`. It seems to pick up the first field (density), but cannot pick up the second (tried velocity and xH_box) as the second box and it failed both times with the above error.

steven-murray:
Wow haha all the tests passed -- clearly need some more coverage. I'll fix those issues.

qyx268:
I've tried this. I first run it in bash environment and cache the outputs last night. I then try to re-run it in Jupyter to do some plotting. However, the outputs are all zero while if I read the file directly using h5py, they are not. for instance:

image

This is with minihalo on since I wanted to test for the bug in #37

qyx268:
forgot to say, if I remove the line `lightcone_quantities=("dNrec_box", "density", 'Gamma12_box','J_21_LW_box','Ts_box','brightness_temp','dNrec_box','xH_box','z_re_box'),` `lightcone.brightness_temp` gives reasonable results. However, I don't have access to other lightcones

steven-murray:
Thanks @qyx268 . I'll look into it ASAP.

qyx268:
I also tried with minihalos turned off. The problem remains

steven-murray:
@qyx I fixed that bug and added a test to check that it is not zero. Can you review?

qyx268:
Hi @steven-murray, the better-lightcone feature looks good. There is a minor point. We need to initialize z_re_box to -1, otherwise, z_re_box will never be assigned a value 
`// keep track of the first time this cell is ionized (earliest time)                       `
`if (flag_options->INHOMO_RECO && (previous_ionize_box->z_re_box[HII_R_INDEX(x,y,z)] < 0)){ `
`    box->z_re_box[HII_R_INDEX(x,y,z)] = redshift;                                          `
`}                                                                                          `
    
After adding initialization of z_re_box=-1, we cannot use the mean to calculate the LC for this quantity. What I was doing is to simply take the maximum when doing the interpolation. I'm wondering if it's a good idea to allow users to specify the math for the LC calculation. 

However, It is not common that people actually need to use lightcone of z_re_box unless they are trying to estimate the gas temperature after reionization... 

PS. since I'm using z_re_box to calculation Mturn in minihalo, not sure if this is causing the boxy feature here #37. I'm still working on this...

qyx268:
I have an example result using 
`import numpy as np`
`import matplotlib`
`%matplotlib inline`
`import matplotlib.pyplot as plt`
`import py21cmfast as p21c`
``
`import logging`
`logger = logging.getLogger("21cmFAST")`
`logger.setLevel(logging.INFO)`
``
`EoR_colour = matplotlib.colors.LinearSegmentedColormap.from_list('mycmap',\`
`             [(0, 'white'),(0.33, 'yellow'),(0.5, 'orange'),(0.68, 'red'),\`
`              (0.83333, 'black'),(0.9, 'blue'),(1, 'cyan')])`
`plt.register_cmap(cmap=EoR_colour)`
``
``
`output_dir = '/home/yqin/aida/hybrid/no_mini-halos/'`
`initial_conditions = p21c.initial_conditions(user_params = {"HII_DIM":128, "BOX_LEN": 250},random_seed=1994, direc=output_dir)`
`lightcone_quantities = ('brightness_temp','Ts_box','xH_box',"dNrec_box",'z_re_box','Gamma12_box',"density")`
`lightcone = p21c.run_lightcone(`
`    redshift = 6.0,`
`    max_redshift = 30.0,`
`    flag_options = {"INHOMO_RECO":True, 'USE_MASS_DEPENDENT_ZETA':True, 'USE_TS_FLUCT':True, 'USE_MINI_HALOS': False},`
`    init_box = initial_conditions,`
`    lightcone_quantities=lightcone_quantities,`
`    random_seed = 1994,`
`    direc = output_dir`
`)`

`cmaps = [EoR_colour,'Reds','magma','magma','magma','cubehelix','viridis']`
`vmins = [-150,  20, 0, -1, 0, 0, -1]`
`vmaxs = [  30, 200, 1,  2, 5, 1,  1]`
`fig, axs = plt.subplots(len(lightcone_quantities),1,`
`            figsize=(getattr(lightcone, lightcone_quantities[0]).shape[2]*0.015,`
`                     getattr(lightcone, lightcone_quantities[0]).shape[1]*0.015*len(lightcone_quantities)))`
`for ii, lightcone_quantity in enumerate(lightcone_quantities):`
`    axs[ii].imshow(getattr(lightcone, lightcone_quantity)[0],`
`                   vmin=vmins[ii], vmax=vmaxs[ii],cmap=cmaps[ii])`
`    axs[ii].text(1, 0.05, lightcone_quantity,horizontalalignment='right',verticalalignment='bottom',`
`            transform=axs[ii].transAxes,color = 'red',backgroundcolor='white',fontsize = 15)`
`    axs[ii].xaxis.set_tick_params(labelsize=0)`
`    axs[ii].yaxis.set_tick_params(labelsize=0)`
`plt.tight_layout()`
`fig.subplots_adjust(hspace = 0.01)`


image

Everthing looks prefect. But you can see that for the z_re_box, there are pink strips because they are all <0 from the interpolation between an ionized cell with a neutral cell with the later being -1.

BradGreig:
Hi @qyx268, to me those pink strips in z_re_box are propagating through the xHI and brightness temperature lightcones as well. You can clearly see discoloured strips in the xHI that overlap with the strips in z_re_box and additionally, in those same locations a slightly different shade of blue for the brightness temperature.

steven-murray:
@qyx yup, I see what @BradGreig is talking about, at least in the xH. But why would it be propagated to xH if it's merely an artifact of interpolation? 

I can change `z_re_box` to use `max` rather than `mean`.

qyx268:
@BradGreig @steven-murray  I'm not really worried about finding those patterns in the xH_box and other related boxes. I think it somehow makes sense that some areas are just not ionized, and hence you see structures in e.g. xH LC and they have z_re=-1.

steven-murray:
OK. I've added the ability to do `max` interpolation instead of `mean` (and this could be extended further later). I've also added a plotting function that does something similar to what @qyx's example did. In fact, I get this plot from running his example (but with my plotting function):

image

Note that I'm running on this branch, not the mini-halos, hence they're not exactly the same, and `z_re_box` is all zeros (so you can't get a sense of what the `max` interpolation does yet).

qyx268:
> Hi @steven-murray, the better-lightcone feature looks good. There is a minor point. We need to initialize z_re_box to -1, otherwise, z_re_box will never be assigned a value
> `// keep track of the first time this cell is ionized (earliest time) `
> `if (flag_options->INHOMO_RECO && (previous_ionize_box->z_re_box[HII_R_INDEX(x,y,z)] < 0)){ `
> ` box->z_re_box[HII_R_INDEX(x,y,z)] = redshift; `
> `} `
> 
yeah, as I was mentioning, we need to initialize z_re_box to -1, which I did it in #37. The master doesn't have it yet...

qyx268:
#110  you can merge this to have the initialization

qyx268:
sorry to reopen this. @steven-murray  can you please update the calculation for global_z_re as well?

steven-murray:
@qyx what do you mean? You should already be able to pass `z_re_box` as a `global_quantity`?

qyx268:
I mean when we plot global_z_re, I think it's still calculated through mean rather than max.

steven-murray:
Ahhhh, of course.

Issue #69, opened 2020-01-06
Minimise memory usage for run_coeval and run_lightcone


BradGreig:
**Describe the bug:**
When using the `run_coeval` or `run_lightcone` classes, these are not the most memory efficient. In both, the `InitialConditions`class is kept around at all times. However, `InitialConditions` in most cases will be the most memory intensive component of the code. 

It might be nice to have the option to split how this is performed. If a relatively large amount of memory is required, then all perturb_boxes can be output to file first, before the rest of the calculation is performed. This way, we do not need to keep the initial conditions in memory, allowing larger boxes to be computed.

This would somewhat mimic the behaviour of the original 21cmFAST drivers. 

I prefer the existing structure, but I can see instances where being able to reduce memory will also be important.

steven-murray:
Hmm, yeah that does seem useful. Can we make this an update for v3.1? I'd
like to get a stable API for v3 out the door before Feb if possible.

BradGreig:
I think this should be fine. Doubt there will be too many people picking this new version up straight away. It shouldn't impact any MCMC or small box runs.

qyx268:
I think I'm hitting this bottle neck now. Perhaps, we can add another argument to `run_coeval` and `run_lightcone`, e.g. `RAM_heavy`. Then we can change `write=write` when calling `perturb_field` to `write=True if RAM_heavy else write`. We also need to skip the calculation of `init_boxes` of course to save RAM...

steven-murray:
I think we decided that this issue is high on the priority list for after
v3.0.0 is released. When you say you're hitting the bottle neck, do you
mean you _need_ this issue fixed right now? I think this issue in
particular is complex enough that we'll want to have a live discussion of
it to all get on the same page and implement the right solution.

qyx268:
sort of... I need to run an MCMC with large DIM, which requires large RAM. I can sort of avoid it by using N_THREAD>1. I took a look of the code in detail and understood it might not be easy to tackle this issue...

steven-murray:
Yup. If you can get around it for now, please try that. If you have ideas for how to make this memory usage better in the future, write them into this thread. v3.0.0 shouldn't be far away -- @BradGreig just has some finishing touches to put on.

BradGreig:
I think there are 3 main areas in which the memory usage can be minimised.

[...]

Issue #63, opened 2020-01-03
Minimise memory usage for InitialConditions class


BradGreig:
**Describe the bug:**
The InitialConditions class initialises all possible memory even if certain options are turned off (e.g. it declares memory for 2LPT and relative velocities even if these options are turned off). 

**Additional context**
The InitialConditions class is the heaviest memory user, thus it would be preferred if this can be minimised if possible by including flags in the class to only initialise memory that is required.

steven-murray:
I think this issue can be augmented in that when reading the cached initial conditions, not all components of the initial conditions are always needed -- sometimes we might just need one or two boxes in the structure. 

I think this can be solved by lazy-loading from the HDF5 files. 

Of course, a true full load is required before handing it off to C.

BradGreig:
Yeah, I was thinking of something similar. That when reading from cache, we are able to pre-define the relevant pieces of information to look for and load into memory. I think this structure will be necessary for quite a few classes.

JulianBMunoz:
As a small note, the relative velocity boxes, which are part of the initial conditions, are used within Ts and Ionizations. So it'd indeed be useful to not get rid of all the InitialConditions, or perhaps save and read the ones we care about.

JulianBMunoz:
Also, related to #69 , it seems to me that big-box runs are pretty unfeasible right now, given the memory requirements. It'd be good to think of a possible workaround/solution to run big boxes without big (memory) bucks.

qyx268:
@JulianBMunoz My impression is that it would take a while to properly implement functions that allow us to read part of the IC. If you wanted to do some big run at the moment, you could take a look at my `skip_init` branch (do git diff HEAD~1 to see the changes, I haven't updated this branch with master for a while). It is a quick hack that skip reading the init and allows us to not store IC in the RAM when running an MCMC

steven-murray:
In about two weeks I should have time to start working on the "proper" solution to this.

JulianBMunoz:
@qyx268 that's a good tip, thanks! For the relvels we only need the lowres box, so perhaps a good solution for now might be to create a new RelVel class that saves that box at the end of initialconditions. We could re-evaluate what to do when we have a solution

steven-murray:
[...]

Issue #27, opened 2019-09-30
[Feature Req.] Minimise memory usage for spin temperature for large memory runs


BradGreig:
**Is your feature request related to a problem? Please describe.**
The spin temperature calculation is optimised for small box realisations, which as a result uses a lot of extra memory. This is not ideal when large box realisations are required.

**Describe the solution you'd like**
Provide the ability to switch between low or high memory usage algorithms depending on if you are doing MCMC or single large realisations.

BradGreig:
Apparently I implemented this already in #224.

PR #87, opened 2020-01-31
Hires rho

catherinewatkinson:
This implements the option to perform the particle moving for the density field (be it with ZA or 2LPT) at high resolution. It was done on a backdated version of the code, so some work is needed to get it inline with the latest version of the code. 

I have added some basic tests to Tests/ and there are also some plots in Tests/DensityFieldTestData/ that show the difference in performing the density field particle movement at low and high resolution.


steven-murray:
Thanks @catherinewatkinson. I'm happy to help with getting the code working with the newest version, if others (eg. @BradGreig and @andreimesinger) can look at some of the plots/tests to make sure it's doing what they think it should be doing.

BradGreig:
@catherinewatkinson @steven-murray I've had a cursory glance of the code and the plots. The behaviour of the plots makes sense and is inline with discussions we've had in the past. I assume it is power rather than dimensionless power in the power spectrum plot?

As for the code (which I appreciate is out-dated and incomplete) I think there could be some further changes. I too would be happy to oversee the integration of this code into the updated version. Some changes are:
1. Renaming the flag (can't think of a good one yet though...)
2. Having both low-res and high-res smoothing options within the code (to enable backwards compatibility)
3. Refactoring the code. Currently there is a considerable amount of redundant code by having one if statement on the outermost level. I think it should be able to be re-written to minimise redundant code.

As I said, I am happy to do all these myself. Just let me know

steven-murray:
@BradGreig awesome, thanks for taking a look. I think this can wait til v3.1, as it shouldn't make any breaking API changes. I'm obviously happy for you to work on it before that though :-) Probably higher priority right now to get the full functionality of previous 21cmFAST though.

catherinewatkinson:
Hi Brad,

All sounds good, it currently has the functionality of being able to run in low res or high res mode, but I definitely agree it could be done in a more elegant fashion to reduce the redundancy of code. There may be ways to speed up the high res version as implemented here.

Cat

catherinewatkinson:
Regarding the plot question, looking at the test unit, I think this should be dimensionless power (i.e. P*k**3/(2*pi**2). However, it's possible I'm mistaken and there is an inconsistency; I did this all pretty quickly since I knew it would ultimately get altered drastically and I knew Andrei was keen to get it, I think for the minihalos stuff.

steven-murray:
Is the code to make the plots in that Test/ directory?

catherinewatkinson:
Yes. In the initial conditions .py test script (I know it's technically also perturb, but I was feeling lazy). I think I might've made the files in a separate python script, but I doubt i got very creative and would've just copied it across when I made the tests. I also think from the fact it's not failing the tests that it must be the same.

steven-murray:
I *might* get time to have a look through this today in a bit more detail.
However, I think at this point that this will be going into v3.1, and it
might be best if @BradGreig gets a chance to go through it first to iron
things out on the C side.

andreimesinger:
This should be pulled in before any Lyman alpha forest work, since the density and velocity fields are better resolved.

BradGreig:
Ok @andreimesinger, I'll try and get to this as soon as I can. I'll work on pulling in the multi-threading branch first though.

BradGreig:
Ok, question for @andreimesinger @catherinewatkinson. I have gone through and added in this high-res implementation of the Zel'dovich/2LPT for the perturbed fields (currently, it is only stored locally). I have a question regarding the down-sampling of this high-res perturbed density field to the low-res density field.

In the first figure, I am showing the density PDF for the original 21cmFAST implementation (perturbed on low-res grid) compared to @catherinewatkinson's implementation (blue) and an alternative (red). Here, the spikiness in the original implementation is gone in the updated implementation.

image

Now, the difference between the blue and red curves is in how the high-res grid is down-sampled to the low-res grid. In the past, when down-sampling, the field in question is filtered by a filter and then sampled in HII_DIM. Now, @catherinewatkinson, instead your implementation is to effectively  average the mass over the entire lower resolution cell volume. In the blue curve, I retain the original 21cmFAST method, down-sampling on the perturbed field (where perturbing was done at high-res).

There is no major difference in the density PDF from either method, both remove the spikiness, which is the desired result from doing the perturbing in the high-res box.

However, there is a considerable difference in the resultant density power spectrum, shown below:
image

In @catherinewatkinson's approach, in averaging the mass over the low-res voxel it smooths the density field, resulting in the turn-over of the density power spectrum. In only using the sampled field (i.e. original approach) the power spectrum does not turn-over (red curve).

My question is, which approach is more appropriate to use? I think the sampling approach (red curve) is more appropriate as having such a large drop-off in the power doesn't seem to be the desired behaviour. The density PS should follow something more closely mimicking the red curve if its goal is to include (quasi) non-linear structure.

catherinewatkinson:
I had a quick skim through the literature and I would say your sampling method looks more consistent. There is a turnover, which is why on my quick check I wasn't alarmed by it, but that is seen in P(k) not Delta(k) as plotted here. I didn't look at any of this stuff in great detail beyond simply ensuring the PDF looked okay. So it's very likely the implementation I did needs refining as suggested.

andreimesinger:
hi brad, if i am understanding correctly, the difference is in the downsampling, with one method doing it in k space with a spherical top hat and the other in real space with a box car...

in general, downsampling in k space is preferred over box car, though one is still prone to aliasing.  indeed the matter power spectrum should not be turning over like that, even though part of that rise might still be due tk aliasing, i think that is preferred

thanks
-a

BradGreig:
@andreimesinger both are smoothed with a real-space filter (as was always the case with 21cmFAST), the difference is the sampling of the filtered hi-res box.

I have,
```
for (i=0; i<user_params->HII_DIM; i++){
    for (j=0; j<user_params->HII_DIM; j++){
        for (k=0; k<user_params->HII_DIM; k++){
            *((float *)LOWRES_density_perturb + HII_R_FFT_INDEX(i,j,k) 
            =
            (*((float *)HIRES_density_perturb + R_FFT_INDEX(i*f_pixel_factor,j*f_pixel_factor,k*f_pixel_factor))/(float)TOT_NUM_PIXELS);
        }
    }
}
```
whereas @catherinewatkinson approach is to average all high resolution elements into a low-res voxel. Hence the power spectrum shape.

```
for (i=0; i<user_params->DIM; i++){
    for (j=0; j<user_params->DIM; j++){
        for (k=0; k<user_params->DIM; k++){
            *((float *)LOWRES_density_perturb + HII_R_FFT_INDEX((i + 0.)/f_pixel_factor),(j + 0.)/f_pixel_factor,(k + 0.)/f_pixel_factor) 
            +=
            (*((float *)HIRES_density_perturb + R_FFT_INDEX(i,j,k))/(float)TOT_NUM_PIXELS) - 1. )/mass_factor;
        }
    }
}
```
In any case, I'll go with what I have implemented as that seems to be the preferred.

BradGreig:
Hi @steven-murray, can you have a look at why this is now failing. One test fails, which I don't understand why given I updated the data for it. But it doesn't output anything useful because the tests never finish as there seems to be issues with travis-ci.

Turning off all integration tests still fails, same issue with travis-ci. Running things locally passes without any problems. I am running with pre-commit and everything.

steven-murray:
Huh. Very strange errors. I'm encountering errors on #127 as well (with the integration tests). I'll plug away at it today (I have it running through valgrind currently). I've found a way to reduce the test run time fairly significantly (if everything works), so that will be nice.

BradGreig:
@steven-murray @andreimesinger @catherinewatkinson, I have marked this PR for review. If you get the time can you have a quick look through. Its failing its test on travis, which @steven-murray is looking into. However, the cause should be due to anything on the C side of things.

src/py21cmfast/outputs.py
```diff
@@ -104,9 +109,21 @@ def _init_arrays(self):
         self.lowres_vz_2LPT = np.zeros(
             self.user_params.HII_tot_num_pixels, dtype=np.float32
         )
+
```
steven-murray, line 112: Should we be initializing these _only_ if doing the `PERTURB_ON_HIGH_RES`?

BradGreig: There are quite a few different arrays (in most classes) that should only be initialised under certain parameter combination. Issues #27, #63 and #69 discuss this to various degrees. At this point I am just aiming to get in all pieces of functionality, before going through and initialising only specific pieces.

steven-murray: Yup, fair enough. We can address this when addressing those issues.

BradGreig:
Hey @steven-murray, I have finally manage to get all the integration tests to pass. However, one test seems to keep failing, `test_readability` from `tests/test_output_structs.py`. Given you added this test, I figure you are the best person to resolve the issue.

This has actually been an issue in this branch for quite some time.

steven-murray:
Awesome @BradGreig ! I'll go and fix that. It's probably something that broke on rebase.

steven-murray:
OK I have gotten this to all work. The problem was actually the memory usage -- there were a couple of functions that were using the default user params and blowing the memory up.

PR #93, opened 2020-02-17
Use write/regen/direc in run_coeval and run_lightcone

### LINKED ISSUE #92: [BUG] Inconsistent usage of the write keyword argument
**Describe the bug:**
Within both the `run_coeval` and `run_lightcone` functions the `write` keyword argument does not get passed to the internal `brightness_temperature` function. This means `brightness_temperature` will default to the `config.yml` value for `write` and ignore whatever value is passed to `run_coeval` and `run_lightcone`.

**Expected behavior:**
Not sure what the best behaviour for this is. In almost all cases we definitely do not want all the intermediate data written out. However, there will be cases (like for what I was just trying to do) where we do not want anything written out at all.

steven-murray:
Fixes #92.


steven-murray:
@BradGreig thanks for the review -- yup missed that `perturb_field` as well.

PR #95, opened 2020-02-19
LF calculation

### LINKED ISSUE #91: [BUG]A minor bug on LF calculation
There is a kink on LFs when F_STAR(M_h) reaches unity. Most of the case they are ok, since F_STAR(M_h) exceeds unity at very bright-end (massive halo). However, some extreme cases the kink appears at the range of Muv we usually plot (e.g. -21 < Muv < -10).

I'm working on it and it will be fixed soon.

jaehongpark00:
Fixed #91 

There was a kink on LFs when F_STAR(M_h) crosses unity. 
To remove this kink, added the interpolation of the derivative (dMuvdMhalo) in a range where the kink appears.

Most of the cases the kinks are smoothed out but in some cases with ALPHA_STAR < -0.5 it doesn't work. However, because ALPHA_STAR < -0.5 is not physical (at least when we consider observations of LFs), this modification is ok.

Attached an example file.

[LF_test-1.pdf](https://github.com/21cmfast/21cmFAST/files/4224455/LF_test-1.pdf)

### DIFF
--- a/src/py21cmfast/src/ps.c
+++ b/src/py21cmfast/src/ps.c
@@ -1527,7 +1527,7 @@ int ComputeLF(int nbins, struct UserParams *user_params, struct CosmoParams *cos
     float Fstar, Fstar_temp;
 
     if (astro_params->ALPHA_STAR < -0.5)
-        LOG_WARNING("ALPHA_STAR is %f, which is unphysical value given the observational LFs.\nAlso, when ALPHA_STAR < -5, LFs may show a kink. It is recommended to set ALPHA_STAR > -0.5.\n",astro_params->ALPHA_STAR);
+        LOG_WARNING("ALPHA_STAR is %f, which is unphysical value given the observational LFs.\nAlso, when ALPHA_STAR < -.5, LFs may show a kink. It is recommended to set ALPHA_STAR > -0.5.",astro_params->ALPHA_STAR);
 
     mf = user_params_ps->HMF;
 
@@ -1638,7 +1638,6 @@ int ComputeLF(int nbins, struct UserParams *user_params, struct CosmoParams *cos
 
             deriv_spline_acc = gsl_interp_accel_alloc();
             deriv_spline = gsl_spline_alloc(gsl_interp_cspline, nbins_smth);
-            //deriv_spline = gsl_spline_alloc(gsl_interp_steffen, nbins_smth);
 
             // generate interpolation arrays to smooth discontinuity of the derivative causing a kink
             // Note that the number of array elements and the range of interpolation are made by arbitrary choices.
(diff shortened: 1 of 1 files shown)


src/py21cmfast/src/ps.c
```diff
@@ -1555,35 +1568,122 @@ int ComputeLF(int nbins, struct UserParams *user_params, struct CosmoParams *cos
...
+        	deriv[i_unity] = gsl_spline_eval(deriv_spline, lnMhalo_param[i_unity], deriv_spline_acc);
+        	deriv[i_unity + 1] = gsl_spline_eval(deriv_spline, lnMhalo_param[i_unity + 1], deriv_spline_acc);
+        	deriv[i_unity + 2] = gsl_spline_eval(deriv_spline, lnMhalo_param[i_unity + 2], deriv_spline_acc);
+        	deriv[i_unity + 3] = gsl_spline_eval(deriv_spline, lnMhalo_param[i_unity + 3], deriv_spline_acc);
+        	deriv[i_unity + 4] = gsl_spline_eval(deriv_spline, lnMhalo_param[i_unity + 4], deriv_spline_acc);
+        	deriv[i_unity + 5] = gsl_spline_eval(deriv_spline, lnMhalo_param[i_unity + 5], deriv_spline_acc);
+        	deriv[i_unity + 6] = gsl_spline_eval(deriv_spline, lnMhalo_param[i_unity + 6], deriv_spline_acc);
+        	deriv[i_unity + 7] = gsl_spline_eval(deriv_spline, lnMhalo_param[i_unity + 7], deriv_spline_acc);
+
+        	for (i=0; i<nbins; i++) {
```
steven-murray: I think we could even move this whole set of cases outside the if statement, and just divide by the deriv on the `if`.

src/py21cmfast/src/ps.c
```diff
@@ -1555,35 +1568,122 @@ int ComputeLF(int nbins, struct UserParams *user_params, struct CosmoParams *cos
...
-                return(2);
-            }
-            if (isinf(log10phi[i + i_z*nbins]) || isnan(log10phi[i + i_z*nbins]) || log10phi[i + i_z*nbins] < -30.) log10phi[i + i_z*nbins] = -30.;
-        }
+        // There is a kink on LFs at which Fstar crosses unity. This kink is a numerical artefact caused by the derivate of dMuvdMhalo. 
+		// Most of the cases the kink doesn't appear in magnitude ranges we are interested (e.g. -22 < Muv < -10). However, for some extreme 
+		// parameters, it appears. To avoid this kink, we use the interpolation of the derivate in the range where the kink appears. 
+		// 'i_unity' is the array number at which the kink appears. 'i_unity-3' and 'i_unity+12' are related to the range of interpolation, 
+		// which is an arbitrary choice. 
+		// NOTE: This method does NOT work in cases with ALPHA_STAR < -0.5. But, this parameter range is unphysical given that the
```
steven-murray: Since we know that `ALPHA_STAR<-0.5` is going to produce weird results, we should probably raise at least a warning here (you can do that with `LOG_WARNING`).

src/py21cmfast/src/ps.c
jaehongpark00: I think it is better to leave this part here. When it doesn't smooth LFs, the code doesn't generate 'deriv' array (using directly 'dMuvdMhalo'). So, if we move this part outside the if statement, we should generate 'deriv' array when the interpolation is not necessary.

src/py21cmfast/src/ps.c
```diff
@@ -1521,10 +1521,15 @@ int ComputeLF(int nbins, struct UserParams *user_params, struct CosmoParams *cos
...
     int i,i_z;
-	int i_unity, i_smth, nbins_smth=7; 
+    int i_unity, i_smth, mf, nbins_smth=7;
     double  dlnMhalo, lnMhalo_i, SFRparam, Muv_1, Muv_2, dMuvdMhalo;
     double Mhalo_i, lnMhalo_min, lnMhalo_max, lnMhalo_lo, lnMhalo_hi, dlnM, growthf;
-    float Fstar, Fstar_temp; 
+    float Fstar, Fstar_temp;
+
+    if (astro_params->ALPHA_STAR < -0.5)
+        LOG_WARNING("ALPHA_STAR is %f, which is unphysical value given the observational LFs.\nAlso, when ALPHA_STAR < -5, LFs may show a kink. It is recommended to set ALPHA_STAR > -0.5.\n",astro_params->ALPHA_STAR);
```
steven-murray: There's a typo here, you say ALPHA_STAR < -5 instead of -0.5.

PR #96, opened 2020-02-21
Add TsBox to Coeval and make a test

### LINKED ISSUE #90: The coeval class doesn't include the spin temperature box struct
The existing coeval class doesn't include boxes related to the spin temperature. It is useful to be able to access the mean spin temperature etc. so I think this should be added to the coeval class.

steven-murray:
Fixes #90 .


src/py21cmfast/wrapper.py
```diff
@@ -1459,6 +1459,7 @@ def __init__(
         perturb: PerturbedField,
         ib: IonizedBox,
         bt: BrightnessTemp,
+        st: [TsBox, None],
         photon_nonconservation_data=None,
     ):
         _check_compatible_inputs(init_box, perturb, ib, bt, ignore=[])
```
BradGreig: You don't check for compatible inputs for `st`. Is this because you are assuming that inputs will be the same?

src/py21cmfast/wrapper.py
```diff
@@ -1459,6 +1459,7 @@ def __init__(
         perturb: PerturbedField,
         ib: IonizedBox,
         bt: BrightnessTemp,
+        st: [TsBox, None],
```
BradGreig: Why is it in this format rather than just TsBox. Is it because of some convention? (just curious)

src/py21cmfast/wrapper.py
```diff
@@ -1734,6 +1741,9 @@ def run_coeval(
                     "PID={} doing brightness temp for z={}".format(os.getpid(), z)
                 )
                 ib_tracker[redshift.index(z)] = ib2
+                if flag_options.USE_TS_FLUCT:
```
BradGreig: Can't this just be a single line with `st_tracker[redshift.index(z)] = st2 if flag_options.USE_TS_FLUCT else None`

src/py21cmfast/wrapper.py
steven-murray: Nope, should check for that, thanks!

src/py21cmfast/wrapper.py
steven-murray: this indicates it can be a `TsBox` or None (since this is optional). Actually, I should set it to None by default...

src/py21cmfast/wrapper.py
steven-murray: Yes! Did that.

src/py21cmfast/wrapper.py
steven-murray: Made it None by default.

src/py21cmfast/wrapper.py
steven-murray: It's now checking for this.

PR #97, opened 2020-02-21
Add installation instructions from OSX issue

### LINKED ISSUE #77: Add note that gsl and fftw can be installed via conda
As @binnietom as suggested, both gsl and fftw can be installed with `conda`, which should make the installation process even easier for those using conda. 

This should be added to the installation instructions. We should also consider creating a conda package so that the entire installation can proceed in one line.

steven-murray:
Closes #77 .


INSTALLATION.rst
```diff
@@ -2,11 +2,62 @@
...
+MacOSX
+~~~~~~
+On MacOSX, obtaining `gsl` and `fftw` is typically more difficult, and in addition,
+the newer native `clang` does not offer `-fopenmp` support.
+
+For `conda` users (which we recommend using), the easiest way to get `gsl` and `fftw`
+is by doing `conda install -c conda-forge gsl fftw` in your environment.
+
+To get `gcc`, either use `homebrew`, or again, `conda`: `conda install -c anaconda gcc`.
+If you get the `conda` version, you still need to install the headers::
```
dprelogo: Maybe just to add a comment here (or in Compile Options) that for conda installed libs, the easiest command is just:
`CC=/path/to/gcc LIB=/path/to/conda_environment/lib INC=/path/to/conda_environment/include pip install ... `
as libs are all in the same folder in conda. Then all the problems one might have is to get a  "good" `gcc`

steven-murray: Ahhh, yes good point. I'll add that in.

steven-murray: Fixed in  75fbc29

steven-murray:
@dprelogo this is ready to review again :-)

PR #98, opened 2020-02-21
Change default cosmology

### LINKED ISSUE #82: Make default cosmology Planck15
It's probably a good idea to be up to date. We'll just have to be careful because this will be a breaking change. Either needs to be in v3.0.0 or delayed until v4.0.0. 

What say people? 

@JulianBMunoz I'm especially looking at you :-)

steven-murray:
Closes #82.

This makes the default cosmology Planck18.

@JulianBMunoz the issue referenced also has some discussion about how to get CLASS transfer functions on-the-fly. If you'd like to continue that discussion, I recommend making a new issue!


src/py21cmfast/inputs.py
```diff
@@ -25,6 +25,12 @@
 
 logger = logging.getLogger("21cmFAST")
 
+# Cosmology is from https://arxiv.org/pdf/1807.06209.pdf
```
JulianBMunoz: Specify it's TT,TE,EE+lowE+lensing+BAO, that is the last column of table 2.

PR #103, opened 2020-02-28
Write high level

### LINKED ISSUE #86: [Feature Req.] Have some way of saving LCs to its own file

It's kind of surprising to the user that Coeval and LightCone objects aren't directly written to file.
It would be good to be able to write those highest-level structures to a nice HDF5 format, and probably _not_ to cache  -- rather specifying the actual filename. 

Basically add a write() method to Coeval and LightCone to write them to a given filename (a default filename based on input parameters should be available too). We should also provide a `read_coeval` and `read_lightcone` function (we could also attach these to the relevant classes as classmethods). 

We could write them to the standard cache (unless otherwise specified). This may be OK as well, as a default. It will also be a lot easier if/when we have a better database management system in place.

We could also just keep our current system, but I think this feature makes sense.

**Additional context**

One issue for this, especially for Coeval cubes, is that we're increasing disk usage quite significantly -- each of the components should already be in the cache, and then the Coeval again contains all of them. We could circumvent this by having pointers in the Coeval file that just point to the other files. However, this will break if the other files are removed...

Looking for some suggestions! @andreimesinger @qyx268 @BradGreig

### LINKED ISSUE #102: [BUG] Output HDF5 files should have the version of 21cmFAST used saved into them
**Description**

Currently the output hdf5 files don't contain the version of 21cmFAST used to produce them. 
This is a bug, because that file could be read with a future of version of 21cmFAST and either not be in the same format (since the format could change over time), or contain errant results (which may be fixed in newer versions).

**Expected behavior**

I think the behaviour we want is that the version of 21cmFAST should be saved into the file (in fact, other metadata such as the time run, the OS, the python version etc. could also be saved into it). 

In particular, the 21cmFAST version should be checked when auto-reading from cache, to ensure that it is up to date. This is slightly more complex, because we also don't want to necessarily ignore files that were created with version 3.0.1 in version 3.0.2, if that update had nothing to do with either how the files are specified, or whether the results are accurate. I'm not entirely sure how to best go about this, but the safest thing is at _least_ to keep the version in the file, and so _some_ check!

steven-murray:
Adds the ability to write `Coeval` and `LightCone` classes to file, and read them back in.
Also adds a few round-trip tests to ensure that a box written then read is the same box.

Fixes #86 
Fixes #102


steven-murray:
@BradGreig @andreimesinger 

This is the solution to the writing of Coeval and Lightcone boxes to a specified file. There are a lot of changes in this PR, because apparently doing this is slightly complicated (to do robustly). I've also shown how to use the functionality in the tutorials. 

Let me know if this method makes sense to you (hopefully it does... basically it's just `Lightcone.save(filename)` and `Lightcone.read(filename)`).

Also, adds the ability to read in an `OutputStruct` from filename directly, i.e. `init = InitialConditions.from_file(filename)`.

src/py21cmfast/plotting.py
```diff
@@ -114,7 +114,7 @@ def coeval_sliceplot(
     Parameters
     ----------
     struct : :class:`~outputs._OutputStruct` or :class:`~wrapper.Coeval` instance
-        The output of a function such as `ionize_box` (a class containing several quantities), or
+        The output of a function such as `ionize_box` (a class containing several lightcones), or
```
BradGreig: Is this a typo? output shouldn't be containing several lightcones

src/py21cmfast/wrapper.py
```diff
@@ -34,7 +34,7 @@
 directly. This behaviour can be tuned in any of the low-level (or high-level) functions
 by setting the `write`, `direc`, `regenerate` and `match_seed` parameters (see docs for
 :func:`initial_conditions` for details). The function :func:`~query_cache` can be used
-to search the cache, and return empty datasets corresponding to each (these can the be
+to search the cache, and return empty datasets corresponding to each (and fthese can the be
```
BradGreig: typo

src/py21cmfast/plotting.py
steven-murray: Haha it's an IDE-typo. I did a change of a variable called "quantities" to "lightcones" and my IDE also looks for references in docstrings and comments. Unfortunately "quantities" is a word used for more than just a variable name.

src/py21cmfast/wrapper.py
steven-murray: Cheers.

steven-murray:
@BradGreig I've also added in a fix for #102, so that the current version of 21cmFAST is written to each box. 

At this point, while the whole version is written into the HDF5 file (eg. `3.0.0.dev2`), only the major and minor versions (eg `3.0`) are used to check compatibility. That is, when searching the cache, only files produced with the same `major.minor` version are matched. This is perhaps a bit extreme -- often minor updates will just be new features, which are entirely compatible, and sometimes patch versions will be bugfixes that are actual updates to the physics. Not exactly sure how to proceed here, but I thought this was a reasonable default place to start.

BradGreig:
This makes sense to me. Is it possible to have a non-standard option that the user can bypass this if they know what they are doing? Not sure of the value of it, but say they know that across versions they have used, the minor version difference made no alteration to their cached results. That way, passing an ignore_minor like quantity, it can ignore the minor differences. However, I guess this is fraught with danger if they have many files stored with slight differences in minor versioning...

steven-murray:
Yeah, you're right that sometimes you might just want to read a box
regardless of when it was written. I've changed the exception that was
raised when reading an old box into a warning. So, the cache will still
only be searched for matching `major.minor` versions, but if a box is
explicitly read, only a warning will be issued, rather than erroring out.
How does that sound?

PR #107, opened 2020-03-03
Photoncons bugfix

### LINKED ISSUE #106: [BUG] Seg-faults for corners of photon non-conservation parameter space.
**Describe the bug:**
For a few extreme corners of astrophysical parameter space the current photon non-conservation correction implementation can fail. 

Additionally, in extreme cases it is also possible to result in a neutral fraction history that increases with decreasing redshift. This is caused by a stagnation in the creation of ionising photons allowing recombinations to take over. Basically the analytic curve still goes to zero, but the 21cmFAST history plateau's resulting in a considerable divergence in the corrected redshift (delta-z). An extrapolation is applied, which somewhat works but in extreme cases cannot. It doesn't always fail, thus causing an exception error doesn't seem useful. Instead, I'll have it through up a warning if many time-steps use interpolation.

BradGreig:
A couple of minor fixes to the photon non-conservation section of the code. Fixes #106.

Throws up a warning and the current set of parameters when extrapolating the photon non-conservation correction for a few (5) consecutive snapshots. It doesn't always result in problems, but worthwhile to provide info to be able to double check results to see if things make sense.


BradGreig:
Hey @andreimesinger, for these extreme cases you cannot default to the median as they are a long way from the median relation. The delta z is ~ 2-3 when it goes into the extrapolation section. I know its not the ideal solution, but its an uninteresting region of parameter space anyway (xHI ~ 0.5 at z = 6). Theoretically it could be caught sooner, allowing the median relation to be used. However, if this was the case, you would still throw up warning's to the user explaining that its switching to the median relation.

You cannot store nearby points as if its the first point, it cannot use nearby points anyway. Additionally, if a user is just using 21cmFAST they wouldn't have the built up information from the MCMC.

PR #108, opened 2020-03-06
Plotting

### LINKED ISSUE #79: [Feature Req.] colorbar label in sliceplot

`plotting.coeval_sliceplot` is an awesome function, but should probably have labels on the colorbar. I'm sure there are other similar functions too (I only just installed 20min ago).

**Additional context**

steven-murray:
This PR adds a few new features to the plotting module. Mostly, these are for the paper, so that the example in there makes a nice plot by default. 

It swaps out the distance ticks on the redshift axis for the lightcone sliceplot for ticks in redshift.
It also adds the ability to add a colorbar label (and some default ones), which was requested in #79.
It also adds a new plotting function that simply plots global histories more easily.

Fixes #79.

This is based off the `write-high-level` branch, so #103 will have to be merged first.

Here's an example of a lightcone plot:

image

And a default global history plot:

image

I think possibly adding a twin x-axis on the lightcone plot to also show distance _might_ be nice. 
There's also possibly some easy/obvious features to put in for the global plot. 
Let me know!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,6 +42,5 @@ htmlcov/
 data/
 
 pip-wheel-metadata/py21cmfast\.dist-info/
-
 .nfs*
 docs/tutorials/*.h5
--- a/src/py21cmfast/__init__.py
+++ b/src/py21cmfast/__init__.py
@@ -31,6 +31,7 @@ from .wrapper import perturb_field
 from .wrapper import run_coeval
 from .wrapper import run_lightcone
 from .wrapper import spin_temperature
+from .wrapper import get_all_fieldnames
 
 configure_logging()
 
--- a/src/py21cmfast/__init__.py
+++ b/src/py21cmfast/__init__.py
@@ -6,6 +6,10 @@ from os import mkdir as _mkdir
 from os import path
 
 from . import cache_tools
+from . import inputs
+from . import outputs
+from . import plotting
+from . import wrapper
 from ._cfg import config
 from ._logging import configure_logging
 from .cache_tools import query_cache
--- a/src/py21cmfast/__init__.py
+++ b/src/py21cmfast/__init__.py
@@ -31,7 +31,6 @@ from .wrapper import perturb_field
 from .wrapper import run_coeval
 from .wrapper import run_lightcone
 from .wrapper import spin_temperature
-from .wrapper import get_all_fieldnames
 
 configure_logging()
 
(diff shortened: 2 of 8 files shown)


BradGreig:
Looks good @steven-murray. I have a couple of thoughts on functionality etc.

- Is this meant to act as the default plotting tool for exploring outputs? Or is it more just for generating quick figures etc.?
- I think the ability to be able to pass ranges for each quantity would be useful. Both min/max for plots and also a min/max for the colour bar for light-cone plots. Though, I don't know how trivial that will be.
- An alternative to the plan on having a twin axis with distance, one could also put frequency. 
- Brightness temperature is typically represented as \delta T_{\rm b}, as it is a contrast (thus differential)

steven-murray:
Thanks Brad.

I'm taking these plotting functions as mostly for producing quick plots for
inspection/validation that nothing went completely wrong. Of course, much
more complex plots can be produced manually. Nevertheless, I also want them
to produce nice default plots that will become "recognizable" in the
community, and to cover the range of obvious functionality that people
might want. Part of this is achieved by returning the figure/axis that
stuff was plotted on, so further modifications can be made directly by the
user.

Ranges are cool, I think that should be super simple (in fact, might
already be doable, I'll check).

The twin axis being frequency is a good idea. I'll try do that on Thursday
(my 21cmFAST day!). I'll also fix the missing Delta.

BradGreig:
Yeah, no problems. I just wanted a gauge of the intention, before providing way too many details/thoughts when it wasn't required. I think what I have above should be sufficient for a quick plotting package. Any other higher level stuff can be done manually as you say.

qyx268:
just a minor point. I've found it useful to allow plotting 'log' scale for some lightcones... e.g. radiation fields.

steven-murray:
OK. I've added the ability to change the redshift axis ticks/labels to either redshift, distance, frequency OR any other function of redshift that the astropy cosmology can compute. I've also tidied up some of the working so that the labels appear on the right axes.

I did *not* add the twin axis, because I just couldn't get it to work. Perhaps that's something to add at a later date.

I also added the ability to set `log=True` on both coeval and lightcone sliceplots,  though that *could* have been done manually already by passing `norm=colors.LogNorm()` to the function.

Ranges are already achievable by passing `clim` to any of the functions. 

I also added a `delta` for brightness temperature.

Here's a sample of lightcone plots:

Default: `p21c.plotting.lightcone_sliceplot(lc)`:
image

Change redshift axis to comoving distance: `p21c.plotting.lightcone_sliceplot(lc, zticks='comoving_distance')`:
image

Change lightcone to `xH` and axis to frequency, and plot in log-space: `p21c.plotting.lightcone_sliceplot(lc, zticks='frequency', kind='xH_box', log=True)`
image

PR #110, opened 2020-03-12
initialize z_res_box

qyx268:
we need to initialize z_res_box


qyx268:
just one line... @steven-murray  you can merge it directly I think...

PR #113, opened 2020-03-13
Fix failing to do photon_cons when reading init from cache

### LINKED ISSUE #109: [BUG]It fails to read cache files with the flag_options of 'PHOTON_CONS'
**Describe the bug:**

When I run the same script the second time, I expect the code reads in from cache instead of re-computing everything. But, with 'PHOTON_CONS', it fails with the following error message,
```
gsl: qag.c:261: ERROR: could not integrate function
Default GSL error handler invoked.
```

It happens if I exit from python (i.e. when I quit jupyter notebook and relaunch it). It happens regardless of specifying the cache folder.

Have you ever met this problem?
I attach my script.

```python
import py21cmfast as p21

p21.global_params.Z_HEAT_MAX = 30.0
p21.global_params.ZPRIME_STEP_FACTOR=1.05

flag_options = {'INHOMO_RECO': False, 'USE_TS_FLUCT': False, 'PHOTON_CONS': False,
                'SUBCELL_RSD': False, 'USE_MASS_DEPENDENT_ZETA': True}

user_params = {'DIM': 150, 'HII_DIM': 50, 'BOX_LEN': 100}
astro_params = {'F_ESC10': -1.09691, 'ALPHA_ESC': -0.4, 'M_TURN': 8.69897, 'L_X': 40.5, R_BUBBLE_MAX': 20.}

lightcone = p21.run_lightcone(
    redshift = 5.9,
    max_redshift = 15.0,
    flag_options = flag_options,
    user_params = user_params,
    astro_params = astro_params,
    direc='caches_simple',
    random_seed=999
)
```

p.s. Yuxiang tried to reproduce this problem with the above script. In his case the code recomputes everything.

steven-murray:
Fixes #109 

Simple fix that the `init_ps` function wasn't being called when reading from cache, since initial conditions weren't being run.

It was hardish to locate, but will get easier once #64 is merged. 

It does bring up the point that there are some of these high-level global initialisation routines which we have to think more carefully about. My fix here was to run `init_ps` for every time we initialise photon cons, but that re-runs things that sometimes don't need to be. We've also run in to this error a few times when code's added, because it's very easy to forget that a function is dependent on `init_ps` having already been run, and assume that it has been.

@BradGreig any ideas?

Let's pull this fix in ASAP.

### DIFF
--- a/src/py21cmfast/src/ps.c
+++ b/src/py21cmfast/src/ps.c
@@ -392,15 +392,10 @@ double sigma_z0(double M){
       kend = 350.0/R;
     }
 
-
-
-
     lower_limit = kstart;//log(kstart);
     upper_limit = kend;//log(kend);
 
     F.function = &dsigma_dk;
-    //  gsl_integration_qag (&F, lower_limit, upper_limit, 0, rel_tol,1000, GSL_INTEG_GAUSS61, w, &result, &error);
-    //    gsl_integration_qag (&F, lower_limit, upper_limit, 0, rel_tol,1000, GSL_INTEG_GAUSS41, w, &result, &error);
     gsl_integration_qag (&F, lower_limit, upper_limit, 0, rel_tol,1000, GSL_INTEG_GAUSS15, w, &result, &error);
     gsl_integration_workspace_free (w);
     return sigma_norm * sqrt(result);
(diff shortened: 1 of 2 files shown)


jaehongpark00:
@steven-murray I think this makes sense and it works.
but, it would also be good to wait for @BradGreig 's answer if he has any other idea.

steven-murray:
I'm going to merge this, but we should continue the conversation about the broader issue here: #114

PR #116, opened 2020-03-15
a bug in recombiation

qyx268:
there is a bug in recombination.c where we exceeded the boundary of the splined table...

I think this is causing the slight difference in #37 test where INHOMO_RECO is True. Could you please update the test_data after this PR?


BradGreig:
Hi @qyx268, yes, I just identified the same issue (don't think I pushed the change yet though). At the same time I think it is worth increasing `RR_lnGamma_NPTS` from 200 to 250. This is what I did, in addition to the change you made.

qyx268:
@BradGreig , me too. I only had it in #37. I've changed it to 250.

Issue #117, opened 2020-03-16
better interpolation for z_re_box


qyx268:
**Is your feature request related to a problem? Please describe.**

Previously in #78 , it was suggested to use the maximum rather than taking the mean to interpolate the z_re_box because the neutral cells have z_re_box=-1. While it is a good approximation for high redshift where only a small number of cells are reionized, it creates artificial stripes at the late stage of reionization when many cells have been reionized. e.g. below was run with {"INHOMO_RECO":True, 'USE_MASS_DEPENDENT_ZETA':True, 'USE_TS_FLUCT':True, 'USE_MINI_HALOS':False}

![image](https://user-images.githubusercontent.com/15994713/76713558-c8e11f80-6721-11ea-8597-53323c84b5e1.png)

**Describe the solution you'd like**

It will be more proper to do the following when interpolating 2 z_re_box values denoted as a and b wit the result being c:
`c = np.mean([a,b]) if (a>=0) and (b>=0) else np.max([a,b])`

**Describe alternatives you've considered**

**Additional context**

Also need to do the same for `global_z_re`

qyx268:
this is essentially considering z_re_box[z_re_box<0]=np.nan, and then doing np.nanmean

qyx268:
I just did it, but it's not helping... I guess these stripes are not due to interpolation but something physical or from the box configuration... 

![image](https://user-images.githubusercontent.com/15994713/76771878-49466580-67a0-11ea-9766-2125e01fa843.png)

andreimesinger:
it doesn’t make sense to plot a lightcone of z_re...  unlike all of the other fields, it doesn’t evolve with time, so each cell has one value, defined at the end of the simulation.

so just remove it from these plots

qyx268:
I know it's not common to plot LF of z_re, but you'll need to use them to calculate gas temperature after reionization in post-processing, e.g. for the forest in 21CMMC. A more proper calculation is, of course, to do it inside 21cmFAST, which might not be something everyone needs.

andreimesinger:
yeah, i am just saying it makes no sense to do a lightcone of it.  of course one should same the cube of z_re.  but it is meaningless do make it into a lightcone.  if you really want to plot something in this format, you could just plot the final box, glueing one next to the other, without interpolation.

but there is no need for a lightcone of z_re, or of any quantity that doesn’t have a redshift dependence

qyx268:
[...]

PR #119, opened 2020-03-16
Better z re

qyx268:
for #117


qyx268:
following #121, we've added back LC for z_re_box

Issue #120, opened 2020-03-17
Tk should be volume weighted temperature between HI and HII


qyx268:
**Is your feature request related to a problem? Please describe.**

Currently, Tk_box in spin_temp represents the kinetic temperature of the IGM for its neutral part only. In principle, it should be the volume-weighted temperature between HI and HII.

**Describe the solution you'd like**

Because we still need previous_spin_temp->Tk in SpinTemperatureBox.c, we cannot simply alter spin_temp->Tk_box inside IonisationBox.c, where we can have the required ionization history. Therefore, in branch `better_Tk`, I had to add a new box, TkIGM_box, in IonizedBox to store the real volume-weighted IGM Tk. 

This plot shows the difference, running with flag_options = {"INHOMO_RECO":True, 'USE_MASS_DEPENDENT_ZETA':False, 'USE_TS_FLUCT':True, 'USE_MINI_HALOS':False}, HII_DIM = 128 and BOX_LEN = 250.
LC (Ts, Tk and TkIGM have the same color scheme):
![image](https://user-images.githubusercontent.com/15994713/76811986-bd5e2900-67f3-11ea-8d52-4785044a7cd9.png)
mean:
![image](https://user-images.githubusercontent.com/15994713/76811989-c222dd00-67f3-11ea-8b00-62629e30faef.png)

I think this result, in general, makes sense. We see at high z, TkIGM basically follows Tk except for those ionized regions where TkIGM is either xHI * Tk + (1-xHI) * 1e4 in the case of partial ionization or equation 6 from McQuinn 2015 for fully ionized regions.
![image](https://user-images.githubusercontent.com/15994713/76811320-65262780-67f1-11ea-97dc-4e7c2a002400.png)
We also see the HII regions cool down after firstly getting ionized. I'm not sure if the temperature of those partially ionized regions is okay or not though. You see these weird jumps in the global evolution (although it's pretty smooth if plotting in linear scale). Note that at these redshifts, Tk is only 10K in HI region and thus HII filling factor of ~0.1% will raise the volume-weighted Tk to be twice the HI region... 

**Describe alternatives you've considered**

**Additional context**

At least, for the sake of saving storage, we might want to not output spin_temp->Tk.

Also, since I've added Fcoll and Fcoll_MINI, which are usually unnecessary to output and can be very storage-consuming if USE_MINI_HALOS. Can we also, by default, not output them? I'm not sure whether not outputting everything will affect caching or not...

[...]

PR #121, opened 2020-03-17
Better tk

qyx268:
This is for #120 

@andreimesinger  please take a look of the changed files and see if the calculation makes sense. There are some results shown in #120

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,10 +4,19 @@ Changelog
 v3.0.0.dev
 ----------
 Complete overhaul of 21cmFAST, including a robust python-wrapper and interface,
-caching mechanisms, and public repository with continuous integration.
+caching mechanisms, and public repository with continuous integration. All changes
+and equations for this version are found in https://arxiv.org/abs/2003.04442
 
 Added
 ~~~~~
+* Updated the radiation source model: (i) all radiation fields including X-rays, UV
+  ionizing, Lyman Werner and Lyman alpha are considered from two seperated population
+  namely atomic-cooling (ACGs) and minihalo-hosted molecular-cooling galaxies (MCGs);
+  (ii) the turn-over masses of ACGs and MCGs are estimated with cooling efficiency and
+  feedback from reionization and lyman werner suppression (Qin et al. 2020). This can
+  be switched on using new ``flag_options`` ``USE_MINI_HALOS``.
+  considered follwing McQuinn (2015), stored in IonizedBox as temp_kinetic_all_gas
+  Note that Tk in TsBox remains to be the kinetic temperature of HI.
 * Tests: many unit tests, and also some regression tests.
 * CLI: run 21cmFAST boxes from the command line, query the cache database, and produce
   plots for standard comparison runs.
--- a/src/py21cmfast/outputs.py
+++ b/src/py21cmfast/outputs.py
@@ -226,9 +226,6 @@ class IonizedBox(_AllParamsBox):
             self.user_params.HII_tot_num_pixels, dtype=np.float32
         )
         self.z_re_box = np.zeros(self.user_params.HII_tot_num_pixels, dtype=np.float32)
-        self.density_re = np.zeros(
-            self.user_params.HII_tot_num_pixels, dtype=np.float32
-        )
         self.dNrec_box = np.zeros(self.user_params.HII_tot_num_pixels, dtype=np.float32)
         self.TkIGM_box = np.zeros(self.user_params.HII_tot_num_pixels, dtype=np.float32)
         self.Fcoll = np.zeros(
@@ -249,7 +246,6 @@ class IonizedBox(_AllParamsBox):
         self.xH_box.shape = shape
         self.Gamma12_box.shape = shape
         self.z_re_box.shape = shape
-        self.density_re.shape = shape
         self.dNrec_box.shape = shape
         self.TkIGM_box.shape = shape
         self.Fcoll.shape = filter_shape
--- a/src/py21cmfast/outputs.py
+++ b/src/py21cmfast/outputs.py
@@ -227,6 +227,7 @@ class IonizedBox(_AllParamsBox):
         )
         self.z_re_box = np.zeros(self.user_params.HII_tot_num_pixels, dtype=np.float32)
         self.dNrec_box = np.zeros(self.user_params.HII_tot_num_pixels, dtype=np.float32)
+        self.TkIGM_box = np.zeros(self.user_params.HII_tot_num_pixels, dtype=np.float32)
         self.Fcoll = np.zeros(
             Nfiltering * self.user_params.HII_tot_num_pixels, dtype=np.float32
         )
@@ -246,6 +247,7 @@ class IonizedBox(_AllParamsBox):
         self.Gamma12_box.shape = shape
         self.z_re_box.shape = shape
         self.dNrec_box.shape = shape
+        self.TkIGM_box.shape = shape
         self.Fcoll.shape = filter_shape
 
         if self.flag_options.USE_MINI_HALOS:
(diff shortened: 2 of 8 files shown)


src/py21cmfast/outputs.py
```diff
@@ -227,6 +227,7 @@ def _init_arrays(self):
         )
         self.z_re_box = np.zeros(self.user_params.HII_tot_num_pixels, dtype=np.float32)
         self.dNrec_box = np.zeros(self.user_params.HII_tot_num_pixels, dtype=np.float32)
+        self.TkIGM_box = np.zeros(self.user_params.HII_tot_num_pixels, dtype=np.float32)
```
steven-murray: In Python, a more descriptive name is better (but you don't need to append the `_box`).

qyx268:
> I can't vouch for the physics, but the code looks good. Can you write a test that uses the new feature? Or at least make a plot comparing old vs. new?

yes, I made it in #120. We need to think about how to deal with partially ionized cells though...

steven-murray:
Yup, just saw that haha, sorry about that!

Yeah, let's wait for @andreimesinger to have a say.

qyx268:
@andreimesinger, if everything in #121 looks okay to you, can you approve this PR and merge it into master?

src/py21cmfast/outputs.py
andreimesinger: just to add on the naming convention: all the fields are "IGM" and "boxes"..  no need for this.  Tgas or Temperature or Tk should be fine...

steven-murray: Yeah, I originally felt that I didn't want to change too many of the names, so left most of these quantities with an appended `_box`, even though I feel like that's a bad name python-wise. Probably in v4 we'll try switch to better names for everything. But let's try to start with better names for new variables now :-)

steven-murray: I don't think `Tk` is particularly descriptive. `temp_gas`, `temp_igm`, or even `temp_gas_kinetic` or `temp_kinetic` (I presume that's what the `k` stands for?) would be better.

steven-murray:
One more thing to add to this PR -- an entry in the changelog!

src/py21cmfast/outputs.py
qyx268: I'm struggling with the naming. It's just both Tk and TkIGM are kinetic temperatures of the gas. It's just one is in SpinBox and one is in IonizedBox... With the former being the temperature of HI while the latter is all gas... 

But I'm a bit worried of renaming TkIGM IonizedBox to Tk and Tk in SpinBox to Tk_HI. I think people who are used to the previous naming convetion might get confused with what's going on with Tk...

steven-murray: Why don't you call `TkIGM` something like `temp_kinetic_all_gas` or `temp_kinetic_igm`?

qyx268: and keep Tk as the way it is?

steven-murray: Yup, for now. Since it already exists in the code, and we want to try to not break the API as much as possible from here on in until 3.0 is released. That will have to be changed on v4.0

andreimesinger:
Temp_kinetic_HI?

steven-murray:
@andreimesinger with a non-capital T, that sounds like a good candidate for v4.

src/py21cmfast/outputs.py
qyx268: okay, I renamed TkIGM to temp_kinetic_all_gas and wrote a note in the CHANGELOG about Tk and temp_kinetic_all_gas

qyx268:
Following #121, temp_kinetic_all_gas is the volume-weighted average between T_RE (the ionization front temperature,2e4K) and Tk (HI temperature) for partially ionized cells and follows McQuinn Eq for fully ionized cells with the density at reionization being estimated from linear extrapolation.
Here are the evolution of the mean(read from LC), mean (calculated from individual coeval box), and median (calculated from individual coeval box) with some randomly picked cells.


image


The large difference between mean and median is due to the bi-modal distribution of temp_kinetic_all_gas.  T-Delta relation at 3 typical ionization stages. (left panel fixed axis range, right zoom in)

TvD_22 1
TvD_09 9
TvD_05 8

CHANGELOG.rst
```diff
@@ -4,10 +4,19 @@ Changelog
 v3.0.0.dev
 ----------
 Complete overhaul of 21cmFAST, including a robust python-wrapper and interface,
-caching mechanisms, and public repository with continuous integration.
+caching mechanisms, and public repository with continuous integration. All changes
+and equations for this version are found in https://arxiv.org/abs/2003.04442
```
andreimesinger: not all of the changes are here..  for the physics, also the velocities from Julian's paper are included.  there there are photon conservation, etc...  i would include a few papers, and remove "all"

CHANGELOG.rst
```diff
@@ -4,10 +4,19 @@ Changelog
...
 
 Added
 ~~~~~
+* Updated the radiation source model: (i) all radiation fields including X-rays, UV
+  ionizing, Lyman Werner and Lyman alpha are considered from two seperated population
+  namely atomic-cooling (ACGs) and minihalo-hosted molecular-cooling galaxies (MCGs);
+  (ii) the turn-over masses of ACGs and MCGs are estimated with cooling efficiency and
+  feedback from reionization and lyman werner suppression (Qin et al. 2020). This can
+  be switched on using new ``flag_options`` ``USE_MINI_HALOS``.
+  considered follwing McQuinn (2015), stored in IonizedBox as temp_kinetic_all_gas
```
andreimesinger: is there missing text here?

src/py21cmfast/src/IonisationBox.c
```diff
@@ -1392,6 +1420,29 @@ LOG_DEBUG("prev_min_density=%f, prev_max_density=%f, prev_overdense_small_min=%f
                 counter += 1;
         }
 
+
+        for (x=0; x<user_params->HII_DIM; x++){
+            for (y=0; y<user_params->HII_DIM; y++){
+                for (z=0; z<user_params->HII_DIM; z++){
+                    if ((box->z_re_box[HII_R_INDEX(x,y,z)]>0) && (box->xH_box[HII_R_INDEX(x,y,z)] < TINY)){
```
andreimesinger: one of these is sufficient no?

qyx268: > one of these is sufficient no?

I think I was worried about cells that recombine a lot. In this case, z_re_box>0 but xH_box > TINY. I intend to keep their temperature as returned by ComputePartiallyIoinizedTemperature

CHANGELOG.rst
qyx268: I'll leave those to their PRs

BradGreig:
Why was this merged before the multithreading branch?

andreimesinger:
sorry, i didn’t know there was a queue, yuxiang had asked to merge it quickly since the changes were very minor…

qyx268:
It doesn't have to be merged at this point. It can wait. If this is causing
troubles, maybe Steven could reset the master to where it was? We can do
this PR after the multithreading one is done. Plus, i haven't fixed the
CHANGELOG yet...

BradGreig:
There isn't normally an order, but for the `mini-halos` and `Multithreading` branches there was (just did a big rebase of `mini-halos` into the `Multithreading`).

steven-murray:
@BradGreig can you try re-rebase onto master? I think reverting history on master once it's up on Github is pretty darn difficult. Hopefully there's not too many changes in this that affect things :/ You could probably try a merge instead of rebase.

andreimesinger:
sorry about that!!

but again the changes in this branch were minor, you can merge and overwrite and yuxiang can then do another PR

PR #124, opened 2020-03-23
a bug when deciding to use hisotry and do increment

qyx268:


steven-murray:
I'll let @BradGreig get #71 in before we merge this :-)

PR #127, opened 2020-03-30
Add timing info to test output

steven-murray:
Adds timing information to test output, so we can determine which tests are hogging all the time.


steven-murray:
@BradGreig: looking at the test timing output, it seems that your new tests are the ones that take the most time. I'm wondering if we can pare them down to bare necessities. We could set the redshift higher (z~15), and set the `zprime_step_factor` higher, as well as considering hard which tests are absolutely necessary. 

But I don't want to remove necessary parts of the tests (eg. perhaps you need to evolve down to a certain redshift to really test the threading?).

BradGreig:
So with these I want to make sure all the main features are working together, which means its going to take a while. With `PhotonCons` you want to make sure it runs far enough that it explores most of the different parts of the calibration. Then, you also want to check that things work with and without threading.

In principle, given you don't actually care about the output (i.e. physical accuracy), you could probably make do with running a large `zprime_step_factor` (maybe even 1.1) and a lower z_heat_max (maybe 25). That should shave off a fair chunk of time.

steven-murray:
@BradGreig: OK, so I've done a few more things with this branch. I thought I'd give you a rundown quickly before I merge (object quickly or forever hold your peace!):

1. I updated the integration tests so that within that module, they are cached (the cache doesn't persist from session to session, or even for other tests in other modules, so it's still well-contained). This split the time in about half.
2. I set the Z_HEAT_MAX=25 for a few of the tests, and the ZPRIME_STEP_FACTOR=1.1 as you suggested (which also brought down the time a bit).
3. I had to up the redshift of a few of the new tests from 8 to 8.5. They were failing for the higher-k parts of the power spectrum. I'd run the data here, upload and test and it would fail. It seems to just be numerical error. I could consider increasing the error tolerance in the integration tests in the future as well. When you think about it, we're doing sums over ~10^6 cells using floats with precision 10^-8. So absolute tolerance of 10^-5 might be too much to ask. Anyway, at z=8.5 they seem to be OK.

BradGreig:
Hey @steven-murray, thanks for the updates. Yeah, all these changes seem reasonable to me. Moving from 8 to 8.5 is fine, such a small change won't change any of the behaviour significantly to be worrisome. Great work on getting the test to run much quicker!

Yeah, I have started to see the same behaviour as you describe in point 3. I put it down to the exact same problem, numerical rounding. That seems to be the cause of the integration test fails I have been seeing. I was contemplating suggesting decreasing the tolerance as a result.

Issue #46, opened 2019-10-11
[Feature Req.] Default should evaluate integrals exactly


BradGreig:
**Is your feature request related to a problem? Please describe.**
The default 21cmFAST should directly perform all integrals, with the existing use of interpolation tables being an additional option. 21CMMC should then turn on interpolation tables automatically

**Describe the solution you'd like**
Add yet another flag option to be able to switch between full integral evaluation and fast, approximate evaluation with interpolation tables

**Describe alternatives you've considered**

**Additional context**

steven-murray:
This sounds like a good idea to me. It sounds like a v3.0.0 thing to me -- does that make sense to you?

BradGreig:
Yep, definitely

Issue #25, opened 2019-09-25
[Feature Req.] Return all functionality of original 21cmFAST 


BradGreig:
**Is your feature request related to a problem? Please describe.**

The current version is based on the 21CMMC version of the streamlined 21cmFAST. There is no support for explicitly using haloes, which is a feature of 21cmFAST.

**Describe the solution you'd like**

Must migrate over all the original functionality of 21cmFAST into this version of 21cmFAST.

steven-murray:
Can we try to break this up into individual issues? Then we can attack them in parallel. We can also prioritize those issues that will be API-breaking.

PR #131, opened 2020-04-08
Add flags for using the halo field and interpolation tables/performin…

BradGreig:
I've only just started this branch, but the intention of this branch is to put back the `USE_HALO_FIELD` from the original 21cmFAST (that is, using the halo field rather than the mean collapse fraction). This deals with #25.

At the same time, I'll add in support for switching between interpolation tables and performing the integrals explicitly. This is in response to #46.

While I am at it, I might try and re-factor some of the C code to make things more readable/manageable. This is mostly in reference to the ionisation and spin temperature fields.

This should be the last significant piece of code/functionality that needs to be added. @steven-murray, I intend to do all these in a single branch. I know that is probably not preferred, but it'll save me having to go through the same code over and over again (which I have done multiple times already). If you are vehemently opposed to this, let me know!


steven-murray:
@BradGreig I am usually opposed to this, but I think in this case it would be a good idea to get all of these remaining bits in in one go. We can start from v3 more carefully adding new things.

That said -- what kinds of C refactoring are you going to do? We can do refactoring after v3 as long as it doesn't change the public-facing API.

BradGreig:
In `ComputeIonizedBox` and `ComputeTsBox` there are large amounts of variables lying around due to attempts to make things efficient. These are then wrapped in if statements inside loops which just makes it all pretty incomprehensible. Ideally, I would move these to one or a couple of functions which are then callable, making it more user friendly. 

I agree it could wait, but it makes sense to do it now since I am messing around with all the code anyway as I will add in the option to flip between explicit integrals and interpolation tables.

Nothing should break the API, however, there may be extra structs that need to be computed and passed around other pieces for the halo fields. This will become clearer as I delve deeper into it.

steven-murray:
@BradGreig -- I just merged #64, so I'd try rebase this onto master now before you do too much :-)

BradGreig:
Excellent, thanks for the heads up!

src/py21cmfast/inputs.py
```diff
@@ -108,7 +108,26 @@ class GlobalParams(StructInstanceWrapper):
...
+        The delta_crit to be used for determining whether a halo exists in a cell
+            0: delta_crit is constant (i.e. 1.68)
+            1: delta_crit is the sheth tormen ellipsoidal collapse correction to delta_crit
+    HALO_FILTER : int
+        Filter for the density field used to generate the halo field with EPS
+            0: real space top hat filter
+            1: sharp k-space filter
+            2: gaussian filter
+    OPTIMIZE : int
+        Finding halos can be made more efficient if the filter size is sufficiently large that
```
steven-murray, line 127: It's not clear from this docstring whether OPTIMIZE is really a bool? I.e. does this just turn on optimization, or are there different levels?

steven-murray, line 127: I would tend towards including these in `UserParams`, so that they're easier to discover, and a bit more flexible. However, we could also opt for doing that later with all of the global params.

src/py21cmfast/src/Find_Haloes.c
```diff
@@ -0,0 +1,470 @@
+
+// Re-write of find_halos.c from the original 21cmFAST
```
steven-murray, line 2: Justa nit-pick, but can the name of this file please be `find_haloes.c` (no capitals)?

src/py21cmfast/src/Find_Haloes.c
```diff
@@ -0,0 +1,470 @@
...
+
+
+// Program ComputeHaloField takes in a k_space box of the linear overdensity field
```
steven-murray, line 5: Remove "Program"

src/py21cmfast/src/Find_Haloes.c
```diff
@@ -0,0 +1,470 @@
...
+// and filters it on decreasing scales in order to find virialized halos.
+// Virialized halos are defined according to the linear critical overdensity.
+// ComputeHaloField outputs a cube with non-zero elements containing the Mass of
+// the virialized halos
+
+int check_halo(char * in_halo, struct UserParams *user_params, float R, int x, int y, int z, int check_type);
+int pixel_in_halo(struct UserParams *user_params, int x, int x_index, int y, int y_index, int z, int z_index, float Rsq_curr_index );
+
+int ComputeHaloField(float redshift, struct UserParams *user_params, struct CosmoParams *cosmo_params, struct AstroParams *astro_params, struct FlagOptions *flag_options, struct InitialConditions *boxes, struct HaloField *halos) {
```
steven-murray, line 14: A lot of this function seems to be roughly the same as `ComputeInitialConditions` (without reading through and comparing explicitly). How much scope is there for refactoring out parts of the function to be used by both?

src/py21cmfast/src/GenerateICs.c
```diff
@@ -32,20 +33,28 @@
...
+// Re-write of init.c for being accessible within the MCMC
 
+int ComputeInitialConditions(
+    unsigned long long random_seed, struct UserParams *user_params,
+    struct CosmoParams *cosmo_params, struct InitialConditions *boxes
+){
     /*
      Generates the initial conditions: gaussian random density field (DIM^3) as well as the equal or lower resolution velocity fields, and smoothed density field (HII_DIM^3).
      See INIT_PARAMS.H and ANAL_PARAMS.H to set the appropriate parameters.
      Output is written to ../Boxes
```
steven-murray, line 50: Should probably remove these lines...

src/py21cmfast/src/GenerateICs.c
```diff
@@ -185,8 +193,11 @@ int ComputeInitialConditions(unsigned long long random_seed, struct UserParams *
 
                     // ok, now we can draw the values of the real and imaginary part
                     // of our k entry from a Gaussian distribution
-                    a = gsl_ran_ugaussian(r[omp_get_thread_num()]);
-                    b = gsl_ran_ugaussian(r[omp_get_thread_num()]);
+//                    a = gsl_ran_ugaussian(r[omp_get_thread_num()]);
+//                    b = gsl_ran_ugaussian(r[omp_get_thread_num()]);
+                    a = 1.;
+                    b = -1.;
```
steven-murray, line 199: is this temporary?

src/py21cmfast/src/PerturbField.c
```diff
@@ -1,15 +1,20 @@
...
-
+int ComputePerturbField(
+    float redshift, struct UserParams *user_params, struct CosmoParams *cosmo_params,
+    struct InitialConditions *boxes, struct PerturbedField *perturbed_field
+){
     /*
-     USAGE: perturb_field <REDSHIFT>
+     ComputePerturbField uses the first-order Langragian displacement field to move the
+     masses in the cells of the density field. The high-res density field is extrapolated
+     to some high-redshift (INITIAL_REDSHIFT in ANAL_PARAMS.H), then uses the zeldovich
```
steven-murray, line 11: Should change reference to `ANAL_PARAMS.h`

src/py21cmfast/src/PerturbHaloField.c
```diff
@@ -0,0 +1,327 @@
+
+// Re-write of update_halo_pos from the original 21cmFAST
+
+// Program UPDATE_HALO_POS reads in the linear velocity field, and uses
+// it to update halo locations with a corresponding displacement field
+// creating updated_halo_ in ../Output_files/Halo_lists/ directory
```
steven-murray, line 6: this text can be updated

src/py21cmfast/src/PerturbHaloField.c
```diff
@@ -0,0 +1,327 @@
...
+//                        boxes->lowres_vy[HII_R_INDEX(i,j,k)] /= growth_factor / user_params->BOX_LEN;
+//                        boxes->lowres_vz[HII_R_INDEX(i,j,k)] /= growth_factor / user_params->BOX_LEN;
+//                        if(global_params.SECOND_ORDER_LPT_CORRECTIONS){
+//                            boxes->lowres_vx_2LPT[HII_R_INDEX(i,j,k)] /= displacement_factor_2LPT / user_params->BOX_LEN;
+//                            boxes->lowres_vy_2LPT[HII_R_INDEX(i,j,k)] /= displacement_factor_2LPT / user_params->BOX_LEN;
+//                            boxes->lowres_vz_2LPT[HII_R_INDEX(i,j,k)] /= displacement_factor_2LPT / user_params->BOX_LEN;
+//
+//                        }
+//                    }
+                    boxes->lowres_vx[HII_R_INDEX(i,j,k)] /= growth_factor / user_params->BOX_LEN;
```
steven-murray, line 297: could pre-calculate `growth_factor/BOX_LEN` here to save some cycles. Also, calculate `HII_R_INDEX` once each iter.

src/py21cmfast/src/ps.c
```diff
@@ -1529,6 +1537,51 @@ void free_lvector(unsigned long *v, long nl, long nh)
...
+/* dnbiasdM */
+double dnbiasdM(double M, float z, double M_o, float del_o){
+    double sigsq, del, sig_one, sig_o;
+
+//    if ((M_o-M) < TINY){
+//        fprintf(stderr, "WARNING:  In function dnbiasdM: M must be less than M_o!\nAborting...\n");
+//        return -1;
+//    }
+    if ((M_o-M) < TINY){
+        LOG_ERROR(" WARNING:  In function dnbiasdM: M must be less than M_o!\nAborting...\n");
```
steven-murray, line 1550: Should just be `LOG_ERROR("M must be less than M_o! Got M=%g, M_o=%g", M, M_o")` then Throw a ValueError.

src/py21cmfast/src/ps.c
```diff
@@ -1529,6 +1537,51 @@ void free_lvector(unsigned long *v, long nl, long nh)
...
+        return(-1);
+    }
+    del = Deltac/dicke(z) - del_o;
+    if (del < 0){
+        LOG_ERROR(" In function dnbiasdM: del_o must be less than del_1 = del_crit/dicke(z)!\nAborting...\n");
```
steven-murray, line 1555: Similar to above comment.

src/py21cmfast/wrapper.py
```diff
@@ -923,6 +1016,282 @@ def perturb_field(
...
+                user_params=user_params,
+                cosmo_params=cosmo_params,
+                regenerate=regenerate,
+                write=write,
+                direc=direc,
+                random_seed=random_seed,
+            )
+
+            # Need to update fields to have the same seed as init_boxes
+            fields._random_seed = init_boxes.random_seed
```
steven-murray, line 1131: Should this be un-indented?

src/py21cmfast/wrapper.py
```diff
@@ -2610,7 +2820,7 @@ def calibrate_photon_cons(
         neutral_fraction_photon_cons = np.array(neutral_fraction_photon_cons[::-1])
 
         # Construct the spline for the calibration curve
-        logger.debug("Calibrating photon conservation correction")
+        logger.info("Calibrating photon conservation correction")
         _calibrate_photon_conservation_correction(
             redshifts_estimate=z_for_photon_cons,
             nf_estimate=neutral_fraction_photon_cons,
```
steven-murray, line 2826: It seems that `run_coeval` and `run_lightcone` don't use the new haloes code at all? Is this intentional?

BradGreig:
Hi @steven-murray, thanks for all the comments (unexpected!). Most of these are because they are still a work in progress, so some things still need to be refined etc. I've commented on a couple seeking clarification or adding context, which I can treat as I go.

src/py21cmfast/inputs.py
BradGreig: This is an optimisation flag purely for finding halos. It is off by default in the old 21cmFAST. Hence, I lumped it into global_params as its probably something that won't be altered too much. But as you say below, this change can happen later.

src/py21cmfast/src/Find_Haloes.c
BradGreig: Do you mean re-naming my existing file "Find_Haloes.c" to "find_haloes.c". If this is true nit-picking, wouldn't that imply you would want all .c files to be renamed as lower case?

src/py21cmfast/src/Find_Haloes.c
BradGreig: It is actually quite different to `ComputeInitialConditions`, there are no redundant calculations between `ComputeHaloField` and `ComputeInitialConditions`. What might look similar is the generating wisdoms if they don't already exist.

That being said, I am toying with the idea of scrapping the halo classes altogether and attaching the halo lists directly to the `InitialConditions` and `PerturbField` classes. This would minimise the number of objects that need to be passed around to functions, requiring less modification of the existing framework.

I think I will make this change for this reason.

src/py21cmfast/src/GenerateICs.c
BradGreig: Yes, this is just me temporarily turning off the RNG so I can easily compare outputs generated from old 21cmFAST as I import new functions etc.

This is the change I referred to potentially making to alleviate the tests failing owing to rounding errors. In principle, we can add a flag (i.e. `TURN_OFF_RNG`) which turns off all RNG usage in 21cmFAST. Then, all tests can take this, and all tests are run without RNG on. It would also be useful for debugging and developing new code.

I think this would be a useful feature, and should be implemented. I can implement this in a separate branch to push this change through faster if you agree.

src/py21cmfast/wrapper.py
BradGreig: Yeah, I haven't finalised the halo behaviour yet. Working on individual pieces before putting them into these functions.

src/py21cmfast/inputs.py
steven-murray: Cool. I reckon it's type should be changed to bool.

src/py21cmfast/src/Find_Haloes.c
steven-murray: The python style is to name modules in `snake_case`, and classes in `CamelCase`. This is C, and I don't know the typical style there. But a mixture of `Snake_Camel` is definitely odd to me. I think at the moment the file-naming convention we seem to have arrived at is that non-public C files (tools and such) are `snake_case` and API-level files seem to be `CamelCase`. I'm happy if we stick with that. Also happy to make all filenames `snake_case`.

src/py21cmfast/src/Find_Haloes.c
steven-murray: OK cool, yeah that might be simpler overall.

src/py21cmfast/src/GenerateICs.c
steven-murray: Yeah, let's add that flag. It can probably just stay as part of this branch since so many changes are going into this. You should try to keep a list of all the additions/changes you make in this branch as you go!

src/py21cmfast/wrapper.py
steven-murray: Cool, that's what I thought might be the case.

steven-murray:
Thanks @BradGreig -- this looks like it's coming along very nicely!

src/py21cmfast/src/GenerateICs.c
BradGreig: @steven-murray, would it make sense to add that flag to #87 instead? Essentially that PR is finished (failing because of the rounding issue). It could be implemented there, update the test data and be pushed to master.

Those changes can then be brought into this branch.

steven-murray:
Hmmm, yeah it could be done there -- might help it pass!

src/py21cmfast/wrapper.py
```diff
@@ -1823,6 +2080,22 @@ def run_coeval(
...
+        # Ensure perturbed halo field is a list of boxes, not just one.
+        if flag_options is not None and pt_halos is not None:
+            if (
+                flag_options["USE_HALO_FIELD"]
+                if type(flag_options) is dict
+                else flag_options.USE_HALO_FIELD
+            ):
+                if not hasattr(pt_halos, "__len__"):
+                    pt_halos = [pt_halos]
+                else:
```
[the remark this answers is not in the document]
steven-murray: Nope, can't figure out a way to do this better.

PR #140, opened 2020-05-05
Easier dev install

### LINKED ISSUE #139: [BUG] Minimum dependency on pytest
**Describe the bug:**
Following updates to the testing, it now appears there are minimum version requirements for `pytest` and `pytest-remotedata`. Without correct versioning, `pytest` will not work locally.

**Details:**
Unsurprisingly, I had an old version of `pytest` (~3.4). Updating to the latest `pytest` (5.4.1) did not immediately resolve the issue. I updated `pytest-cov` from 2.6->2.8, which also did not resolve the issue. Finally, updating `pytest-remotedata` to the latest version (0.3.2) resolved issues 

I think minimum requirements of `pytest` need to be > 5.0.0, most importantly `pytest-remotedata` > 0.3.1 (maybe 0.3.0, but not sure).

steven-murray:
Fixes #139 

Updates the minimum requirement for `pytest-remotedata`, and makes it easier for the user to install dev dependencies.


setup.py
```diff
@@ -67,6 +67,17 @@ def _safe_copy_tree(src, dst, safe=None):
 if "TOXENV" in os.environ and "SETUPPY_CFLAGS" in os.environ:
     os.environ["CFLAGS"] = os.environ["SETUPPY_CFLAGS"]
 
+test_req = [
+    "pre-commit",
+    "pytest",
+    "pytest-cov",
+    "tox",
+    "pytest-remotedata>=0.3.2",
```
BradGreig, line 75: I had to update pytest as well, so there might be dependency on that too. Not sure how easy that is to test though. Perhaps minimum of > 5.0.0. You could also just move it to the latest version.

PR #142, opened 2020-05-14
Compute lf bug

qyx268:


qyx268:
There were some minor bugs when calculating the LF

PR #143, opened 2020-05-14
free HIRES_density_perturb if PERTURB_ON_HIGH_RES

qyx268:
we need to free these arrays otherwise ram cumulates too much during mcmc


qyx268:
It's weird. I cannot reproduce these flake8 errors, which seem also happen in #142

steven-murray:
The flake8 error is a bit strange (since it's only happening on Travis, not locally), but the error is quite clear. I've fixed it in #141, so once that's merged, you can rebase here and it should go away.

Issue #144, opened 2020-05-17
PhotonCon crashed


qyx268:
**Describe the bug:**

I think this line suggests that the program crashed due to a bug related to photon conservation.
`2020-05-17 19:51:09 | DEBUG   | IonisationBox.c | ComputeIonizedBox:156 [pid=234219] | original redshift=5.261127, updated redshift=nan delta-z = nan       `              

At first, I thought it was due to the limited redshift range in InitialisePhotonCons. It was set to z=5 while the lightcone was requested with redshift=4.9. However, after increasing a_end to z=4.9 and rerunning without any cached files, the program crashed again at the same redshift. I suspected "delta-z" could be larger than 0.3 for photon conservation calibration and therefore with an original redshift = 5.26, the updated one could be smaller than 4.9. However, after reducing from z=4.9 to z=4.4, the program (with cached files) still crashed...
                                                          
**To Reproduce:**
The following set of parameters is causing this problem.
astro_params = py21cmfast.AstroParams(ALPHA_ESC=-0.5782171938644294,
                                      ALPHA_STAR=0.2931801482826073,
                                      F_ESC10=np.log10(0.06528048324333377),
                                      F_STAR10=np.log10(0.04046237545524282),
                                      M_TURN=np.log10(1222992760.4016855),
                                      t_STAR=0.5344442329701387)
flag_options = py21cmfast.FlagOptions(INHOMO_RECO=True, PHOTON_CONS=True, USE_MASS_DEPENDENT_ZETA=True)
user_params = py21cmfast.UserParams(USE_FFTW_WISDOM=True)
py21cmfast.global_params.Z_HEAT_MAX=20.0
py21cmfast.global_params.ZPRIME_STEP_FACTOR= 1.02

py21cmfast.run_lightcone(redshift=4.9, user_params=user_params, astro_params=astro_params, flag_options=flag_options,random_seed=1993)

**Expected behavior:**

**Details:**
 - OS: linux (nefertem)
 - Python Version: 3.6.5
 - Relevant compiler versions:
 - Relevant package versions:

**Additional context**

qyx268:
I found the array NeutralFractions has a minimum of 0.029 which is larger than PhotonConsAsymptoteTo (0.01). This led 
`*absolute_delta_z = gsl_spline_eval(deltaz_spline_for_photoncons, global_params.PhotonConsAsymptoteTo, deltaz_spline_for_photoncons_acc);`
to return NaN.

[...]

PR #145, opened 2020-05-18
Expand photoncons range

BradGreig:
This is in response to #144. I have added in the ability for the user to modify the end-point of the photon non-conservation correction calibration. Additionally, added in errors to `run_coeval` and `run_lightcone` to catch this condition as soon as possible.

As discussed in #144, I have added this functionality to `global_params` because it will be very rare for the user to ever need this behaviour.

@steven-murray, I have tagged you to make sure the errors to catch this condition I added make sense (or if there is a better way).

### DIFF
--- a/src/py21cmfast/wrapper.py
+++ b/src/py21cmfast/wrapper.py
@@ -1858,18 +1858,6 @@ def run_coeval(
             else:
                 redshift = [p.redshift for p in perturb]
 
-        if flag_options.PHOTON_CONS:
-            calibrate_photon_cons(
-                user_params,
-                cosmo_params,
-                astro_params,
-                flag_options,
-                init_box,
-                regenerate,
-                write,
-                direc,
-            )
-
         if not hasattr(redshift, "__len__"):
             singleton = True
             redshift = [redshift]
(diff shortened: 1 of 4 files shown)


src/py21cmfast/wrapper.py
```diff
@@ -1902,6 +1890,35 @@ def run_coeval(
         redshifts += redshift
         redshifts = sorted(set(redshifts), reverse=True)
 
+        if (
+            flag_options.PHOTON_CONS
+            and np.amin(redshifts) < global_params.PhotonConsEndCalibz
+        ):
+            raise ValueError(
+                """
+                The final (lowest) redshift (z = %g) is lower than the endpoint of the photon
```
steven-murray, line 1899: I think I would change this particular error message to start with "You have passed a redshift (z = %g) that is lower than...".

Actually for both this and the other error message, you could take advantage of f-strings to put the parameters in, so:

```python
f"""You have passed a redshift (z = {np.amin(redshifts)}) that is lower than..."""
```

qyx268:
Testing with the same parameter set as in #144 shows this branch solved the issue.

BradGreig:
@steven-murray, I have updated the Python error message, plus added some C-side errors to catch this behaviour. Unfortunately the C-side errors will only catch once it occurs (which is right at the end), so hopefully the Python side always does its job in catching it!

Issue #153, opened 2020-06-26
[BUG] In first couple of steps for mini halos


BradGreig:
**Describe the bug:**
Potentially two separate issues: 
1) if selecting a user defined Z_HEAT_MAX and running at high redshift it will crash.
2) Ionisation box doesn't seem to use the previous density field at the second redshift in the scrolling loop.

**To Reproduce:**
For 1): run_coeval(
        redshift=18,
        regenerate=True,
        write=False,
        z_heat_max = 25,
        zprime_step_factor=1.2,
        ) 

It'll start at z = 26.36, then next iteration will be z = 21.8 (prev_z = 26.36). But, ionize_box won't find a box at 26.36 and thus fail (it'll fail because ionise box will populate it with an empty box as prev_z > Z_HEAT_MAX.

For 2): run_coeval(
        redshift=18,
        regenerate=True,
        write=False,
        zprime_step_factor=1.2,
        )

It'll start at z = 38.398, then next iteration will be z = 31.8 (prev_z = 38.398). But in ionize_box, the variables (output in ULTRA-DEBUG) prev_min_density and prev_max_density will be 0.00. Yet, this is the second iteration, and these should be non-zero, as the ionisation box at z = 38.398 is the first redshift (and min_density and max_density at z = 38.398 are non-zero). Therefore, on second iteration it doesn't seem to use the prev_density_field (the one from z=38.398). Not sure if this behaviour is related to the first (it should be as in both instances prev_z > Z_HEAT_MAX, however, it results in two different behaviours which is weird to me).

BradGreig:
By the way, this behaviour occurs on the master branch

BradGreig:
One possible solution I came up with was to replace this code in `ionize_box` in the `run_coeval` fuction:

```previous_perturbed_field=(
                    None
                    if iz == 0 or not flag_options.USE_MINI_HALOS
                    else (
                        perturb[redshift.index(redshifts[iz - 1])]
                        if redshifts[iz - 1] in redshift
                        else None
                    )
                ),
```

with:

```
previous_perturbed_field=(
                    None
                    if iz == 0 or not flag_options.USE_MINI_HALOS
                    else (
                        perturb[redshift.index(redshifts[iz - 1])]
                        if redshifts[iz - 1] in redshift
                        else perturb_field(
                                       redshift=z,
                                       regenerate=regenerate,
                                       write=write,
                                       direc=direc,
                                   )
                    )
                ),
```

[...]

PR #154, opened 2020-06-27
add more perturb fields when using MINI_HALOS

qyx268:
also need to add exemptions of C901 in flake


qyx268:
this fixes #153

src/py21cmfast/wrapper.py
```diff
@@ -1884,11 +1884,31 @@ def run_coeval(
         else:
             redshifts = [min(redshift)]
 
+        # USE_MINI_HALOS needs all perturb_field in redshifts!
+        if flag_options.USE_MINI_HALOS:
+            perturb_all_scroll = []
+            for ii in range(len(redshift)):
+                perturb_all_scroll += [perturb[ii]]
+            for z in redshifts:
+                perturb_all_scroll += [
```
BradGreig, line 1893: I don't think that generating another list of perturb_field boxes is the way to go as it can require a large amount of memory. Given that its something that is only required with `USE_MINI_HALOS`, perhaps my suggestion is better as it only requires one box at a time to be kept around. Theoretically, you could add another box (like st, ib) which can be populated with the previous perturb field as it goes through the loop (i.e. it starts at None, then after each iteration it is populated with the perturb field used at the previous redshift snapshot).

qyx268:
You are right. I didn't pay attention to the RAM. I now redo the changes and store perturb at only one extra redshift.

.flake8
```diff
@@ -11,6 +11,7 @@ ignore =
     N806  # Naming upper/lowercase -- too hard right now.
     N802  # Naming upper/lowercase -- too hard right now.
     D401  # Docstring in imperative mood. This should *not* be the case for @property's, but can't ignore them atm.
+	C901
```
steven-murray, line 14: Instead of ignoring C901, increase max-complexity slightly below it.

BradGreig:
@steven-murray I think this test would have failed had the `USE_MINI_HALOS` tests not have been skipped. Unfortunately this only arises with the `USE_MINI_HALOS`option, hence it not being picked up. However, given the recent modifications you made to the testing suite perhaps the `USE_MINI_HALOS` option isn't as slow? If so, we could pass a lower value for `NUM_FILTER_STEPS_FOR_Ts`, say 8 or 10 which should make things faster.

qyx268:
I have enabled the test of mini halos. But I have to also add `z_heat_max` = 25 to speed it up. As @BradGreig said, this test should have captured the error if it were not passed. I can confirm that the integration feature test does fail (more precisely `pytest` reported `Segmentation fault (core dumped)` and crashed) on master with this configuration.

steven-murray:
@qyx268 great! The test is still taking ~5.5 min on its own, but I can't really see a way around this right now. Perhaps in the future if we can break the c-code up a bit more, we can test the individual parts more simply, and get this time down. Or we can have different sets of tests -- some longer ones need only to be run when making a new release, say.

PR #156, opened 2020-07-08
Skip init

qyx268:
I tested. this allows us to skip reading initial conditions

### DIFF
--- a/src/py21cmfast/wrapper.py
+++ b/src/py21cmfast/wrapper.py
@@ -1567,7 +1567,7 @@ def spin_temperature(
             )
 
         # Dynamically produce the initial conditions.
-        if init_boxes is None or not init_boxes.filled:
+        if (init_boxes is None or not init_boxes.filled) and not perturbed_field.filled:
             init_boxes = initial_conditions(
                 user_params=user_params,
                 cosmo_params=cosmo_params,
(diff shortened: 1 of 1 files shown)


qyx268:
we still need to assign `skip_init=True` to `run_lightcone` and `run_coeval` in `build_model_data`

steven-murray:
Thinking about this a bit more, I *think* this can be done even more easily -- if we just try to load the perturb field boxes first, since that's all we really need, then if they are cached they will be read without any reading of init boxes. Otherwise, the init boxes will automatically be created. Could you try that as a solution?

steven-murray:
Thinking more about it, I think you're right @qyx268. Generally we don't want to do the idea I had above, because it would typically require reading in the init_box lots of times (every redshift), instead of keeping it in memory. 

I would like to have a unit test of this though. Playing around with caching has a way of introducing subtle bugs.

BradGreig:
To me it seems unsatisfactory to have to have more flags/keywords to solve this. I suspect the way to go is down the path of whats discussed in #69 (picking up on @steven-murray 's suggestion).

steven-murray:
I agree with @BradGreig here. I think it could be implemented quickly for
`run_lightcone`, and we can discuss details for implementing in
`run_coeval` and `ionize_box` etc. in a separate PR.

qyx268:
no problem. I was looking into a quick solution for my current MCMC with high resolution density. We could of course come back to this and see how to implement the solution mentioned in #69.

I will close this PR now.

PR #160, opened 2020-07-17
v3

steven-murray:
This PR simply updates the code to the official v3. 
It bumps the version and updates the changelog.

Please comment if you think anything is not ready!

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,12 +1,16 @@
 Changelog
 =========
 
-v3.0.0.dev
-----------
+v3.0.0
+------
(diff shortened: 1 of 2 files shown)


BradGreig:
The most pressing things now are improving memory management and adding a function to estimate how much memory is needed before actually performing a calculation. But we agreed that these were to come after anyway. I can't think of much else, but I could be wrong. Where does the relative velocities stuff sit? Was that completed?

steven-murray:
@BradGreig I believe all the relevative velocities stuff was pulled in. @JulianBMunoz can comment further?

PR #162, opened 2020-07-23
Refactor fftw wisdoms

BradGreig:
I have refactored the code for creating the FFTW Wisdoms. Originally the C code would first check to see if the wisdom exists, if it didn't it would create it. Now, there is a separate function for creating FFTW Wisdoms, which will do exactly the same.

Basically, this minimises redundant code.

### DIFF
--- a/src/py21cmfast/wrapper.py
+++ b/src/py21cmfast/wrapper.py
@@ -308,7 +308,7 @@ def get_all_fieldnames(arrays_only=True, lightcone_only=False, as_dict=False):
 # WRAPPING FUNCTIONS
 # ======================================================================================
 def construct_fftw_wisdoms(*, user_params=None, cosmo_params=None):
-    """Construct all necessary FFTW wisdomes.
+    """Construct all necessary FFTW wisdoms.
 
     Parameters
     ----------
@@ -320,7 +320,7 @@ def construct_fftw_wisdoms(*, user_params=None, cosmo_params=None):
     cosmo_params = CosmoParams(cosmo_params)
 
     # Run the C code
-    return lib.CreateFFTW_Wisdoms(user_params(), cosmo_params())
+    return lib.CreateFFTWWisdoms(user_params(), cosmo_params())
 
 
 def compute_tau(*, redshifts, global_xHI, user_params=None, cosmo_params=None):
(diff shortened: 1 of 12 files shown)


src/py21cmfast/wrapper.py
```diff
@@ -307,6 +307,22 @@ def get_all_subclasses(cls):
 # ======================================================================================
 # WRAPPING FUNCTIONS
 # ======================================================================================
+def construct_fftw_wisdoms(*, user_params=None, cosmo_params=None):
+    """Construct all necessary FFTW wisdomes.
```
steven-murray, line 311: "wisdomes"

src/py21cmfast/wrapper.py
```diff
@@ -944,6 +963,9 @@ def perturb_field(
             except IOError:
                 pass
 
+        # Construct FFTW wisdoms. Only if required
```
steven-murray, line 966: How does the "only if required" part work? there's nothing in the Python function that stops it from executing?

BradGreig: Its all encapsulated in the C function. The C function only does something **if** the ``USE_FFTW_WISDOM`` flag is passed to it, and will only create new wisdoms if it cannot locate existing ones.

I deliberately avoided logic on the Python side to minimise function complexity.

BradGreig:
I agree that there can be more refactoring done. I'll create a more generalised PR for this at a later date. This was just a simple and obvious one that popped up while I was prepared to look into estimating memory usage etc.

src/py21cmfast/wrapper.py
steven-murray: Ah, OK cool.

steven-murray:
Ah, one more thing to do and I'll approve: add the change to the changelog, under 3.0.1, and bump the version.

src/py21cmfast/wrapper.py
BradGreig: Interesting... Can't even consistently spell

BradGreig:
I think I have made the necessary changes. I thought about breaking up the FFTW creation into individual functions, however, I do not think this is necessary. Having the one function do them all should be sufficient.

PR #171, opened 2020-09-26
Lightcone callback

dprelogo:
Add optional `coeval_callback` flag to `wrapper.run_lightcone`, which represents arbitrary function calculated on `p21c.Coeval` instance for each `node_redshift`. Passed function receives only one parameter, precisely `p21c.Coeval` instance. 
### Changes
- implementation in `src/py21cmfast/wrapper.py`
- simple test in ` tests/test_wrapper.py`
- CHANGELOG.rst


src/py21cmfast/wrapper.py
```diff
@@ -2673,6 +2681,21 @@ def run_lightcone(
...
+                    perturbed_field=pf2,
+                    ionized_box=ib2,
+                    brightness_temp=bt2,
+                    ts_box=st2 if flag_options.USE_TS_FLUCT else None,
+                    photon_nonconservation_data=_get_photon_nonconservation_data()
+                    if flag_options.PHOTON_CONS
+                    else None,
+                    _globals=None,
+                )
+                coeval_callback_output.append(coeval_callback(coeval))
```
qyx268, line 2697: I suggest adding a check here to see whether the given function will work

src/py21cmfast/wrapper.py
```diff
@@ -2446,6 +2447,9 @@ def run_lightcone(
         If given, must be compatible with init_box. It will merely negate the necessity of
         re-calculating the
         perturb fields. It will also be used to set the redshift if given.
+    coeval_callback : callable, optional
+        User-defined arbitrary function computed on :class:`~Coeval`, at each node redshift.
```
qyx268, line 2451: Do we really want to do all the node redshifts which can be pretty large? Perhaps, it's better to use an user-defined list with the results being interpolated between the closest two redshifts or (even better) adding the list into `scrollz`. In particular, `scrollz` doesn't always contain the redshifts that are needed for coeval outputs, for instance, we usually need integer redshifts like 6,8,10,15...

qyx268 (approved):
The PR looks nice to me. I have two suggestions.

src/py21cmfast/wrapper.py
dprelogo: Do you think it is okay just to run a function once? What if it does some irreversible change?

src/py21cmfast/wrapper.py
dprelogo: Yep, that's true I agree, I don't know if it can be easily done, will try

andreimesinger:
for now, it is best to avoid interpolation, as it adds overhead.  giving a list of redshifts or “all” and outputting the closest in redshift among those already computed, seems easiest and quickest

src/py21cmfast/wrapper.py
qyx268: I'm not sure what you mean. I would just put an `try, except`...

src/py21cmfast/wrapper.py
qyx268: following andrei's suggestion, you could also give a new argument like `redshifts_coeval`, and e.g. identify the closest redshift before going into the loop and then only call coeval_callback when you are at those pre-identified points. If you don't like the idea of adding more arguments, maybe the users can define a tolerant z range in the callback function and skip those unnecessary redshifts.

steven-murray:
A couple of things:

1. Currently, we always compute the `scrollz` using a roughly logarithmic series with user-input log-spacing and end-points. We could consider an input option of `node_redshifts` where the user can explicitly set the redshifts they want (by default, using the thing we already do).
2. I feel like *another* option of `callback_redshifts` could be slightly confusing, but is probably OK. By default, it should be the same as `scrollz` (as it already is in this PR).
3. If we do include such an option, the last thing to consider is how to treat redshifts not in the `scrollz` (or the `node_redshifts`). I think having *another* option `z_tolerance` may be useful. If a `callback_redshift` is within `z_tolerance` of a `node_redshift`, then we just use the coeval at that redshift directly. If it's not within tolerance, we have two options: (i) add that redshift to the `node_redshifts` before doing the loop, or (ii) interpolating. While @andreimesinger argues that interpolating introduces overhead, computing a whole new coeval is *more* overhead. If going this route, I'd be inclined to by default set `z_tolerance` quite large (perhaps even infinite), so that by default, it always just snaps to the closest `node_redshift`, but the user can set it lower so that it will do some interpolation. 

The downside is that this will introduce two new input parameters to the function (`z_tolerance` and `callback_redshifts`). However these will both have sensible defaults, so it's probably not so bad.

andreimesinger:
thanks steven, but imo, the user shouldn’t have too much control over the redshifts used in the calculation.  there is implicit trapezoidal integration in the code and we don’t want them to be able to set long step sizes.  the logarithmic scrolling was empirically tested to give convergence for a fiducial model.

the question here was just what to output *given* we already have done the calculation at various redshifts (as far as i understand the coevals are all already created at zscroll redshifts, so it is a matter of writing out) to avoid the overheads, to just output at the nearest zscroll output to a given value, or just output all or none

steven-murray:
To be fair, the user already has control over the cadence of the zscroll
redshifts -- through the `Z_PRIME_STEP_FACTOR` global parameter. But its
default is reasonable. But I take your point -- we don't want to give too
much chance for the user to screw things up. I'm happy to not let the user
pass the redshifts. But then, we just need to decide on how to manage the
outputs at coevals that *are* calculated. I believe the scheme I outlined
is the most general.

dprelogo:
Considering the discussion, if I could suggest then the following:
```
coeval_callback = {
    "function": function_to_calculate,
    "redshifts": list or int > 1
    }
```
For redshifts: if list, return computed function output for the closest redshifts with respect to `scrollz`; if int, computes every n-th redshift in `scrollz`. Imo this should give more than enough control to the user.

steven-murray:
I think it's easier and less error-prone to just provide two arguments,
rather than a dictionary of arguments. Otherwise the user could potentially
pass random stuff into the dictionary, and we have to check that the
correct keys are in there etc. Simpler just to offer those two arguments in
the `run_lightcone` function itself, and document them there. If we think
that the `scrollz` are going to be close enough together that interpolation
isn't an issue, then these arguments will be enough. We could also consider
adding the interpolation functionality at a later date.

dprelogo:
Thank you all, I've added then `coeval_callback_redshifts` argument to the `run_lightcone`. It either takes closest values to specified redshifts or computes every n-th redshift in `scrollz`.

src/py21cmfast/wrapper.py
dprelogo: Ah okay. So you would just skip in the case the function doesn't work? Just thinking if maybe it's a good thing that the code fails in the case the function fails...

src/py21cmfast/wrapper.py
```diff
@@ -2540,6 +2553,32 @@ def run_lightcone(
...
+                if sum(compute_coeval_callback) != len(coeval_callback_redshifts):
+                    logger.warning(
+                        "some of the coeval_callback_redshifts refer to the same node_redshift"
+                    )
+            elif (
+                isinstance(coeval_callback_redshifts, int)
+                and coeval_callback_redshifts > 0
+            ):
+                compute_coeval_callback = [
+                    True if i % coeval_callback_redshifts == 0 else False
```
steven-murray, line 2574: this would be simpler as `[not i % coeval_callback_redshifts for i in range(len(scrollz))]`

src/py21cmfast/wrapper.py
```diff
@@ -2673,6 +2712,22 @@ def run_lightcone(
                 regenerate=regenerate,
             )
 
+            if coeval_callback is not None:
+                if compute_coeval_callback[iz]:
```
steven-murray, line 2716: just make this `if coeval_callback is not None and compute_coeval_callback[iz]`

src/py21cmfast/wrapper.py
```diff
@@ -2673,6 +2712,22 @@ def run_lightcone(
...
+                        perturbed_field=pf2,
+                        ionized_box=ib2,
+                        brightness_temp=bt2,
+                        ts_box=st2 if flag_options.USE_TS_FLUCT else None,
+                        photon_nonconservation_data=_get_photon_nonconservation_data()
+                        if flag_options.PHOTON_CONS
+                        else None,
+                        _globals=None,
+                    )
+                    coeval_callback_output.append(coeval_callback(coeval))
```
steven-murray, line 2729: You can wrap this in a Try: Except:, but in the Except, still raise an error -- it just means the error can maybe make a little more sense (you can attach which redshift it broke at, for example). Perhaps even better -- if it's the first time in the function that the callback is being called, and it fails, raise an error. Otherwise, raise a warning and keep going.

steven-murray (commented):
I think this is looking good. Just a few comments to tighten some of the code. Thanks @dprelogo!

dprelogo:
Thanks @steven-murray, please see if I'm using logger properly.

src/py21cmfast/wrapper.py
```diff
@@ -2712,22 +2711,31 @@ def run_lightcone(
...
+                except Exception as e:
+                    if sum(compute_coeval_callback[: iz + 1]) == 1:
+                        raise RuntimeError(
+                            f"coeval_callback computation failed on first trial, z={z}."
+                        )
+                    else:
+                        logger.warning(
+                            f"coeval_callback computation failed on z={z}, skipping."
+                        )
+                        logger.warning(e)
```
steven-murray, line 2738: I *think* this won't work -- you might need `e.message` or something like that. You should be able to test this to make sure (as a pytest test, use the `pytest.warns` function to check that a suitable warning is raised).

dprelogo: Would instead of the two the following work: `logger.exception(f"coeval_callback bla bla")` ?

steven-murray: I'm not sure exactly what you mean? Even to the `logger.exception` I think you'll have to give it a string, not an actual `Exception` instance. I can't remember if calling `logger.exception` actually raises an exception or not...

dprelogo: Okay I checked this, exception class has default `__str__` method, so both `print(e)` or `logger.warning(e)` or similar will work. However this doesn't print exact exception class name but only final error message, so I think something like `f"{type(e).__name__}, {e}"` should be sufficient. `logger.exception("message")` prints out "message" and full traceback, which is maybe redundant?

steven-murray (approved):
Looks good to me!

PR #172, opened 2020-09-29
Add basic description and install instructions to readme

steven-murray:
Adds basic installation instructions to the readme, as well as a small description,
and also Yuxiang's awesome plot to make it prettier. Essentially responding to
JOSS paper review.


steven-murray:
@qyx268  -- I love this plot that I added. However, perhaps it would be better to have a smaller/thinner one horizontally oriented, with fewer panels?

qyx268:
Of course, this plot was made to fix in an A4 page. I can easily redo it with few panel horizontally.

qyx268 (approved):
looks nice, i will send you another lightcone asap.

qyx268:
hmm... I just recalled that these LCs were reproduced using 21cmFAST v2 (as the two minihalo papers were not using v3). For instance, we cannot get J or Mcrit lightcones currently without hacking the v3 code...  After publishing the paper, I've also deleted the files needed to reproduce these lightcones. But I can easily redo them since I have kept the parameters. However, before I start doing this, I'm wondering if we want to update these lightcones (possibly in the JOSS paper as well) with something the current code can produce. 

PS. I have something for the forest paper which was a direct output of v3 and also includes new features of temperature and lyman alpha transmission. I think this will be the first published LCs using v3. I'm wondering if we should use this one instead... However, I don't have the 21-cm brightness temperature for this though and this project focus on z<8... 
reference

In summary, 3 options:
1. redo the minihalo LCs using v2;
2. use the forest LCs shown above; and
3. redo the minihalo LCs (or Park+18 LCs) using v3.
(Each candidate will have 2 uninterpreted minutes to explain why the 21cmFAST people should vote for them rather than their opponent)

andreimesinger:
i would suggest to not bother with this extra work, and just keep the current figure…  indeed the ability to output any of the computed fields is one of the biggest development priorities for v3..

steven-murray:
I think we can leave it as-is for now. We can update the plot at a later time. I think having 2-3 horizontal (thin) lightcones would be nice. Probably simple stuff like density, xH and Tb.

PR #174, opened 2020-10-02
Address Sultan's Comments on Paper

### LINKED ISSUE #81: [Feature Req.] Functionality for obtaining the full attributes of a Coeval object

For the user, getting what all the attributes of a Coeval instance are is kind of annoying -- they have to actually create one and then check, or go check each OutputStruct. It would be nice to have access to a clean, up-to-date compiled list of these attributes. This relates to comments made in  https://github.com/21cmfast/21CMMC/pull/26, where these attributes can be set to be saved in MCMC.

It would be nice to be able to *document* the attributes in the `Coeval` documentation. This may be difficult, I'm not sure. Failing this, at least a function that can get the names without creating an instance would be nice.

steven-murray:
This PR attempts to address most/all of the comments from @sultanier
at https://github.com/openjournals/joss-reviews/issues/2582.

It adds some documentation on simple usage, especially the CLI, to the readme,
fixes a couple of issues with the CLI, and adds the ability to output the full
Coeval/Lightcone objects created.

Partially resolves #81 by defining a `Coeval.get_fields` method which can be run as a classmethod to obtain the different "kinds" that are plot-able.

### DIFF
--- a/INSTALLATION.rst
+++ b/INSTALLATION.rst
@@ -23,6 +23,14 @@ The C libraries required are:
 As it turns out, though these are fairly common libraries, getting them installed in a
 way that `21cmFAST` understands on various operating systems can be slightly non-trivial.
 
+HPC
+~~~
+These libraries will *usually* be available on a HPC environment by using the
+``module load gsl`` and similar commands. Note that just because they are loaded
+doesn't mean that ``21cmFAST`` will be able to find them. You may have to point to the
+relevant ``lib/`` and ``include/`` folders for both ``gsl`` and ``fftw`` (these should
+be available using ``module show gsl`` etc.)
+
 Linux
 ~~~~~
 Most linux distros come with packages for the requirements, and also `gcc` by default,
--- a/INSTALLATION.rst
+++ b/INSTALLATION.rst
@@ -14,22 +14,26 @@ However, since `21cmFAST` relies on some C libraries, this is not always possibl
 
 The C libraries required are:
 
-* `gsl`
-* `fftw` (compiled with floating-point enabled, and `--enable-shared`)
-* `openmp`
-* A C-compiler with compatibility with the `-fopenmp` flag. **Note:** it seems that on
-  OSX, if using `gcc`, you will need `v4.9.4+`.
+* ``gsl``
+* ``fftw`` (compiled with floating-point enabled, and `--enable-shared`)
+* ``openmp``
+* A C-compiler with compatibility with the ``-fopenmp`` flag. **Note:** it seems that on
+  OSX, if using ``gcc``, you will need ``v4.9.4+``.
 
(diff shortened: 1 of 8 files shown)


README.rst
```diff
@@ -57,15 +57,82 @@ Installation
...
+    >>> )
+    >>> p21c.plotting.coeval_sliceplot(coeval, kind='brightness_temp')
+
+The coeval object here has much more than just the ``brightness_temp`` field in it. You
+can plot the ``density`` field, ``velocity`` field or a number of other fields.
+To simulate a full lightcone::
+
+    >>> lc = p21c.run_lightcone(
+    >>>     redshift=8.0,
+    >>>     max_redshift=15.0,
```
andreimesinger, line 92: is the default to not include spin temperature fluctuations?  if not, then z=15 is too low.

README.rst
```diff
@@ -57,15 +57,82 @@ Installation
...
+
+The simulation box produced is automatically saved into the cache (by default, at ``~/21cmFAST-cache``).
+You can list all the files in your cache (and the parameters used in each of the simulations)
+with::
+
+    $ 21cmfast query
+
+To run an entire coeval cube, use the following as an example::
+
+    $ 21cmfast coeval 8.0 --out=output/coeval.h5 --HII_DIM=35
```
andreimesinger, line 121: resolution is really low here..  does the result look ok?

andreimesinger (commented):
double check defaults here, but otherwise looks ok.  though i didn't look though cli.py carefully

README.rst
steven-murray: Yes, default is to not include spin temp.

README.rst
steven-murray: I've upped the resolution to `HII_DIM=100` because that'll be a bit more satisfying for a first-time run. Obviously, this is intended to be a super fast simple example.

INSTALLATION.rst
```diff
@@ -23,6 +23,14 @@ The C libraries required are:
 As it turns out, though these are fairly common libraries, getting them installed in a
 way that `21cmFAST` understands on various operating systems can be slightly non-trivial.
 
+HPC
+~~~
+These libraries will *usually* be available on a HPC environment by using the
+``module load gsl`` and similar commands. Note that just because they are loaded
```
BradGreig, line 29: In my experience this may not necessarily be true. FFTW used by 21cmFAST is the float, multiprocessing version which may not always be installed.

BradGreig (approved):

INSTALLATION.rst
steven-murray: Yeah, true. I've updated it to say "often" instead of usually, and point out that FFTW has to have the right options.

Issue #176, opened 2020-10-10
No error raised or warning when a decreasing redshifts list is used in .wrapper.compute_tau


YihaoZhou27:
**Describe the bug:**
The function `.wrapper.compute_tau` assumes the input `redshifts` is an increasing list. When the user inputs a decreasing redshift list, it doesn’t raise an error but gets a tau value which doesn’t depend on the input parameters.

**To Reproduce:**
1. Call `.wrapper.compute_tau` several times with the same `user_params`, `cosmo_params`, and a `redshifts` which is decreasing, but different gloabl_xHI lists each time.
2. You will get the same output.

**Expected behavior:**
When the user inputs a decreasing redshift list, the function works as well or raise an error.

**Additional context**
In `UsefulFunctions.c` codes, when it calls `ComputeTau` , it sets the `zstart` to be 0 instead of redshift_list[0], making the raise-error codes based on whether` zstart >= zend` useless.

BradGreig:
Hi @YihaoZhou27 thank you for pointing this out. I have submitted a PR to add in an error message indicating these quantities must be ascending.

PR #177, opened 2020-10-12
Add in error that redshifts/xHI must be ascending for computing tau

BradGreig:
This is a minor bug-fix associated with #176. Note, warning has been added if redshifts are not ascending, but not for neutral fractions. Neutral fractions need to be ascending, however a check is made more complicated by the fact that multiple entries of zero or one can be added in the list of neutral fractions. Thus the check is only performed on redshifts, but the warning mentions both must be ascending.


src/py21cmfast/wrapper.py
```diff
@@ -339,14 +339,20 @@ def compute_tau(*, redshifts, global_xHI, user_params=None, cosmo_params=None):
...
     cosmo_params = CosmoParams(cosmo_params)
 
     if len(redshifts) != len(global_xHI):
         raise ValueError("redshifts and global_xHI must have same length")
 
+    if sum(
+        (redshifts[i + 1] - redshifts[i]) > 0 for i in range(len(redshifts) - 1)
+    ) != (len(redshifts) - 1):
+        raise ValueError("redshifts and global_xHI must be in ascending order")
+
```
steven-murray, line 355: I think this would be cleaner as `if not np.all(np.diff(redshifts) > 0`

steven-murray (commented):
HI Brad, just one suggested change to make the code more pythonic

steven-murray:
Also, couldn't you check for neutral fraction just by doing `if not np.all(np.diff(xH)>=0)`?

BradGreig:
Hey @steven-murray, I've added the recommended fix.

It doesn't work so cleanly/easily for the neutral fraction. At high-z, prior to ionising sources, the neutral fraction doesn't always decrease. It can sometimes fluctuate at 10^-3 - 10^-4 precision. One could put in a tolerance for this, but couldn't guarantee the adopted tolerance would always be successful. I've also seen instances (within PhotonCons and recombinations on) where reionisation can completely stagnate and the neutral fraction can increase again. Thus, a simple check that the neutral fraction data is also always increasing doesn't quite work.

steven-murray (approved):
Thanks @BradGreig. Yeah, I get your rationale on `xHI`. I wonder if we shouldn't put a `warning` in there instead? So, eg. put another line like

```python
if not np.median(np.diff(global_xHI)>-1e-3):
    warnings.warn("Non-increasing global_xHI detected. This could indicate a problem. Please check!")
```

I'll leave it up to you whether you think that's a good idea.

BradGreig:
Hey @steven-murray, yeah, a warning would be nice. I think this would work in most instances, however, I can think of some instances where it might fail (as it is dependent on the number of True/False entries). For example, where this occurs is at high-z, if reionisation is sufficiently delayed (or doesn't happen at all) then you could end up with a large number of entries that will cause this warning to be thrown. Though I guess that in any case, warning the user of this might not be a bad thing anyway.

Ideally I would like to minimise this warning appearing too many times. I tried out masking the data to be sure you are sampling reionisation (e.g. `0.1 < xHI < 0.9`) and can then at least through an Error instead, however, it will equally fail when reionisation never commences.

I'm not sure if you want to go down the route of having multiple conditions. Say, if you at least get to `xHI < 0.9' actually through an Error using the mask method. However, if this cannot occur (i.e. `xHI > 0.9`), through a warning using your method.

qyx268:
I'm wondering if an ascending neutral fraction is indeed required. There were models of minihalo where the early population ionizes the universe very quickly (to xHI~10%) and disappears. In this case, xHI rises until atomic cooling galaxies become important. 

In my view, I think it is better to add lines to sort the inputted redshift list to our required ascending list. e.g. if this is already in the correct order, then use the inputs directly, otherwise, flip the redshift and xH before using them. I think users, in general, won't pass a xH list that doesn't match the redshift. So we shoulnd't need to worry about the mismatching and I would argue that it is the user's freedom to pass any arbitrary EoR history. 

In fact, I got people here using this function asking me why the resulting tau was already 0.3something (which is the value you get when inputting in the wrong order).

steven-murray:
I think raising an error on redshift order is best -- it's also possible
(though not likely) that the user could pass an unsorted list (so not just
reverse order), in which case flipping will also be wrong. It's easy to
sort your own input, so we might as well just raise the error and let the
user do the right thing.

If there are reasonable cases in which xHI may increase with decreasing
redshift, then I guess we better not use a warning OR error on that, but
just assume that the user has input xHI and redshift consistently. OTOH, if
the algorithm actually breaks when xHI doesn't always decrease (or close to
it), we should emit a warning/error so that we can report a more useful
message.

BradGreig:
@steven-murray are you content with how things are, or are you advocating for adding in further warnings? You last response was a little ambiguous to that.

steven-murray:
I'm content, on the condition that  `ComputeTau` itself doesn't break if `xHI` isn't monotonic. If it could break, then we should add a warning (at least). If not, and it really does just depend on z increasing, then perfectly happy as-is.

BradGreig:
No, `xHI` doesn't need to be monotonic. Only `z` needs to be monotonically increasing, as it is a requirement of the integration routine. Theoretically `xHI` can do whatever it wants!

BradGreig:
@steven-murray, by the way, I noticed a minor bug in the `_get_photon_nonconservation_data` function. Two labels were flipped. I have gone and added a fix into this bug fix. I know it is not ideal to do this, but it was a simple fix, and this bug fix itself should be good to go now.

src/py21cmfast/wrapper.py
```diff
@@ -765,8 +769,8 @@ def _get_photon_nonconservation_data():
         "Q_analytic",
         "z_calibration",
         "nf_calibration",
-        "delta_z_photon_cons",
         "nf_photoncons",
+        "delta_z_photon_consnf_photoncons",
```
steven-murray, line 773: This looks like it needs a comma!!

BradGreig: Not a comma, but delete some text. Nice catch, trying to do things too quickly...

steven-murray (approved):
Great! Thanks @BradGreig

steven-murray:
@qyx268 if you could just add a note to the changelog, I'll merge this one in!

Issue #179, opened 2020-10-13
Can anyone reproduce this FFTW bug?


qyx268:
**Describe what you would like to do**
minimum steps to reproduce this bug:
`python -c "import py21cmfast as p21c; p21c.construct_fftw_wisdoms(user_params={'USE_FFTW_WISDOM': True, 'HII_DIM':25})"`

You will find no wisdom outputs in the current directory. However, this only happens on `thanatos` and I cannot reproduce this bug on other machines. Can somebody else try on their machine? @dprelogo, it will be great if you can confirm this is also happening to you on `thanatos`. So we will know if this is just because of my environment or it's the machine...

Python 3.6.5
FFTW 3.3.8

**What I've tried**
I managed to narrow down the error to `fftwf_export_wisdom_to_filename` in `GenerateICs.c`, which return a status of 0 (failed). However, `fftwf-wisdom -o ~/wisdomf rof25 rob25` does show an output.

**Other context**

This might be related to #134 as I didn't get any information from `run_lightcone` apart from ERROR saying no wisdom files can be found...

steven-murray:
When running that command both on my own machine, and on a cluster here, I
do not get any output files in my current directory. That is on Arch Linux,
and Ubuntu.

dprelogo:
I'm confirming that there's no output on `thanatos`.
Python 3.8.1
FFTW 3.3.8

BradGreig:
I cannot reproduce this on any machine I have access too. This will make if difficult for me to try and debug...

Is someone else willing to try and figure this one out?

qyx268:
@BradGreig  Just wanted to double check. Do you mean you do see wisdom outputs in the current directory when you say you cannot reproduce this? 

I will look into this in the next couple of days.

BradGreig:
@qyx268, yes, when following your instructions, I am left with several output files in the directory (clean to begin with).

Unfortunately, I do not have access to thanatos to test this myself. 

One quick thing worth checking. Does it create the wisdoms if you have a clean directory and just call `initial_conditions`? That is bypass your direct call to `construct_fftw_wisdoms` and leave it to be done internally by `initial_conditions`. I can't see why, but a good sanity check.

qyx268:
I still cannot figure this out. I'm postponing this issue to later times. I just wanted to summarize what I did. I tried the followings:
1. manually create the file with fopen and write wisdom using `fftw_export_wisdom_to_file`. Nothing appears in the file
[...]

PR #181, opened 2020-10-19
fix some major memory leak (+ some small leak...)

### LINKED ISSUE #161: RAM gradually builds up when running 21cmmc (memory leak?)
**Describe the bug:**
There seems to be still memory leak in the code. When running 21cmmc, the RAM gradually builds up after hundreds of simulations. I did a quick check, it doesn't seem to be related to fftw arrays... 

**To Reproduce:**
Just run a 21cmmc simulation with USE_MASS_DEPENDENT_ZETA and INHOMO_RECO (haven't checked if these are necessary) and monitor the RAM usage.

**Expected behavior:**

**Details:**
 - Python Version: [e.g. 22]
 - Relevant compiler versions:
 - Relevant package versions:

**Additional context**

qyx268:
fix #161 

Before merging into the master, can @jaehongpark00 pull the changes from this `free_mem` branch and test with the current photonCon run on `nefertem` to see if the RAM still rises?


qyx268:
hmm... currently, some of the test failed at test_integration_features when using FFTW. This is probably related to #179. I'm wondering why this didn't show up before...

I'll come back to this later.

qyx268:
er... the problem was solved after fixing a bug which i dont think it's responsble... anyhow. can someone please do a review?  @steven-murray @BradGreig

src/py21cmfast/src/IonisationBox.c
```diff
@@ -1688,12 +1688,30 @@ LOG_SUPER_DEBUG("freed fftw boxes");
 
     }
 
+    if (prev_redshift < 1){
+        free(previous_ionize_box->z_re_box);
+        if (flag_options->USE_MASS_DEPENDENT_ZETA) {
```
BradGreig, line 1693: Maybe (flag_options->USE_MASS_DEPENDENT_ZETA && flag_options->USE_MINI_HALOS)

src/py21cmfast/src/IonisationBox.c
```diff
@@ -1688,12 +1688,30 @@ LOG_SUPER_DEBUG("freed fftw boxes");
...
+        }
+    }
+
     if(!flag_options->USE_TS_FLUCT) {
         if(user_params->USE_INTERPOLATION_TABLES) {
             freeSigmaMInterpTable();
         }
     }
 
+    if(INIT_ERFC_INTERPOLATION) {
```
BradGreig, line 1709: No, this shouldn't be done like this (well, at least not currently).

These arrays are generated once per session, thus they'll need to be freed differently. At the top of the file (line 194), where these are first allocated, instead do:
if(!flag_options->USE_MASS_DEPENDENT_ZETA && INIT_ERFC_INTERPOLATION) {

These are only used for the old, T_VIR dependence

src/py21cmfast/wrapper.py
```diff
@@ -2366,6 +2366,9 @@ def run_coeval(
         else:
             photon_nonconservation_data = None
 
+        if flag_options.USE_TS_FLUCT and user_params.USE_INTERPOLATION_TABLES:
+            lib.FreeTsInterpolationTables(flag_options())
```
BradGreig, line 2370: This is only for the `run_coeval` instance. You need to ensure this is also done for the `run_lightcone` as well.

BradGreig (changes requested):

src/py21cmfast/src/IonisationBox.c
qyx268: Can I confirm that you are saying to add `!flag_options->USE_MASS_DEPENDENT_ZETA` to both this line and 194 rather than saying freeing these two arrays here is wrong? I agree this is only used when USE_MASS_DEPENDENT_ZETA==False

qyx268:
The recent commits should also fix #182

src/py21cmfast/src/IonisationBox.c
BradGreig: Ah, yeah, sorry for not being clear. I was saying to add `!flag_options->USE_MASS_DEPENDENT_ZETA` to line 194 (as you have done). However, you cannot free `ERFC_VALS` or `ERFC_VALS_DIFF`. Thus, I would remove the freeing of these. These are populated once per session, thus freeing them will cause segmentation faults.

We'll need to come up with a better way to free these. However, adding `!flag_options->USE_MASS_DEPENDENT_ZETA` to line 194 will at least ensure this isn't populated in most cases.

qyx268: Ah, I see what you mean. I didn't realize `INIT_ERFC_INTERPOLATION` is a global parameter rather than an input parameter. I will remove the freeing then

qyx268:
for the record, sometime this line in `SpinTemperatureBox.c`
`(this_spin_temp->first_box || (fabs(initialised_redshift - perturbed_field_redshift) > 0.0001` is True at multiple redshifts which results in executing `initialiseSigmaMInterpTable`, `initialise_Nion_Ts_spline` and `initialise_SFRD_spline` and allocating memory for interpolation tables multiple times. Therefore, I added a check to see if the pointer is NULL before allocating. 

However, this is not easy for local static pointers. Therefore, I moved init_heat() out of the aforementioned check and do destruct_heat() for each SpinTemperatureBox call. This is not a heavy overhead since the tables are very small.

qyx268:
@jaehongpark00's MCMC run shows much better improvement for RAM. There are only some small leakages due to repeated interpolation table allocation, which is now fixed.

@BradGreig  could you please see the change you requested and let me know if the new implementations make sense to you.

BradGreig (approved):

PR #185, opened 2020-10-20
doc: update ref to Yuxiang's paper

steven-murray:


qyx268 (approved):
Sorry, I think you are making qin2020 the same as qin2020a. I've updated it.

qyx268:
for some reason, docs build has failed....

PR #192, opened 2020-11-17
Fixed FFTW missing with both 2LPT and relvels are on

### LINKED ISSUE #191: Turning USE_RELATIVE_VELOCITIES to True screws up density lightcones [BUG]
**Describe the bug:**

Including relvels makes the lightcones appear wrong, both on T21 but also on densities (attached both with and without USE_RELATIVE_VELOCITIES: True; densities should *not* depend on the relative velocities at all). Densities and velocities look fine out of initial_conditions.

Screen Shot 2020-11-17 at 2 52 41 PM
screen_shot_2020-11-16_at_11 11 03_pm

**To Reproduce:**
1. Run initial conditions with:
user_params = {"HII_DIM":HII_DIM, "BOX_LEN": BOX_LEN, "USE_FFTW_WISDOM": True, 'USE_INTERPOLATION_TABLES': True,
              "USE_RELATIVE_VELOCITIES": True, "POWER_SPECTRUM": 5}
2. Run lightcone with:
lightcone_fid_vcb = p21c.run_lightcone(
        redshift = 20.0,
        init_box = initial_conditions,
        flag_options = flag_options_fid,
        astro_params = astro_params_fid,
        lightcone_quantities=lightcone_quantities,
        global_quantities=lightcone_quantities,
        random_seed = random_seed,
        direc = output_dir,
        write=True
)

3. Plot T21 or densities in the lightcone. 

**Expected behavior:**

I have checked that using “USE_RELATIVE_VELOCITIES”: False, “POWER_SPECTRUM”: 5 I get sensible results, but If I set “USE_RELATIVE_VELOCITIES”: True

I’ve checked that doing plotting.coeval_sliceplot(initial_conditions, “hires/lowres_density”) is normal (the same as with no relvels), so something must go wrong when processing the ICs into the lightcone or into the coeval boxes (which also look noisy and insane, also attached).

screen_shot_2020-11-16_at_11 12 09_pm

I'm checking if setting EVOLVE_DENSITY_LINEARLY=True fixes it, to see if the problem is in perturb.c. It looks like memory corruption to me.

JulianBMunoz:
Fixes #191


src/py21cmfast/src/GenerateICs.c
```diff
@@ -754,12 +754,12 @@ int ComputeInitialConditions(
                 if (fftwf_import_wisdom_from_filename(wisdom_filename) != 0) {
                     plan = fftwf_plan_dft_r2c_3d(user_params->DIM, user_params->DIM, user_params->DIM,
                                                  (float *) HIRES_box, (fftwf_complex *) HIRES_box, FFTW_WISDOM_ONLY);
-                    fftwf_execute(plan);
                 } else {
                     LOG_ERROR("Cannot locate FFTW Wisdom: %s file not found",wisdom_filename);
                     Throw(FileError);
                 }
             }
+            fftwf_execute(plan);
```
steven-murray, line 762: you could actually just move it out of here as well, to line 769

steven-murray (approved):
Thanks @JulianBMunoz!

PR #193, opened 2020-12-01
Less redundant coeval

### LINKED ISSUE #164: [Feature Req.] Option to save coeval T21 boxes at all intermediate z, instead of just the endpoint
Right now each coeval box has a structure associated with it, including initial condition (IC) boxes, etc. So if a user needs the 21cm coeval box at many z there will be many unnecessary copies of the ICs, taking a lot of disk space.

Ideally, we could add an option to save all intermediate coeval boxes between Z_HEAT_MAX and the z indicated by the user. Since those are calculated anyway it's just a matter of saving to disk if requested (maybe warning the user it'd take more space than standard).

Another possibility, mentioned by @steven-murray , is to save all coeval boxes independently, but linking to the same IC cached file. It might be a problem if the IC file gets deleted.

steven-murray:
This is mostly @JulianBMunoz's work. It adds the ability to access Coeval cached data at higher redshifts that contributed to a given data product (either Coeval or Lightcone). It also adds the ability to gather all that data into a single file, if you really want.

Fixes #164


.flake8
```diff
@@ -12,7 +12,7 @@ ignore =
     N802  # Naming upper/lowercase -- too hard right now.
     D401  # Docstring in imperative mood. This should *not* be the case for @property's, but can't ignore them atm.
 max-line-length = 88
-max-complexity = 30
+max-complexity = 35
```
JulianBMunoz, line 15: Mo' complexity mo' problems. But what con you do.

JulianBMunoz (approved):

PR #199, opened 2021-01-05
Fix wisdoms

steven-murray:
This partly addresses #179 with a quick workaround where wisdoms are used if found, and otherwise don't produce an error, just a warning. 

More importantly, it removes a lot of boilerplate code from several C files, replacing it with a two simple functions in `dft.c`.

### DIFF
--- a/src/py21cmfast/src/BrightnessTemperatureBox.c
+++ b/src/py21cmfast/src/BrightnessTemperatureBox.c
@@ -119,25 +119,7 @@ int ComputeBrightnessTemp(float redshift, struct UserParams *user_params, struct
 
         memcpy(vel_gradient, v, sizeof(fftwf_complex)*HII_KSPACE_NUM_PIXELS);
 
-        if(user_params->USE_FFTW_WISDOM) {
-            // Check to see if the wisdom exists, create it if it doesn't
-            sprintf(wisdom_filename,"real_to_complex_DIM%d_NTHREADS%d.fftwf_wisdom",user_params->HII_DIM,user_params->N_THREADS);
-            if(fftwf_import_wisdom_from_filename(wisdom_filename)!=0) {
-                plan = fftwf_plan_dft_r2c_3d(user_params->HII_DIM, user_params->HII_DIM, user_params->HII_DIM,
-                                             (float *)vel_gradient, (fftwf_complex *)vel_gradient, FFTW_WISDOM_ONLY);
-                fftwf_execute(plan);
-            }
-            else {
-                LOG_ERROR("Cannot locate FFTW Wisdom: %s file not found",wisdom_filename);
-                Throw(FileError);
-            }
-        }
-        else {
-            plan = fftwf_plan_dft_r2c_3d(user_params->HII_DIM, user_params->HII_DIM, user_params->HII_DIM,
-                                         (float *)vel_gradient, (fftwf_complex *)vel_gradient, FFTW_ESTIMATE);
-            fftwf_execute(plan);
-        }
-        fftwf_destroy_plan(plan);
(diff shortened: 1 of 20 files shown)


src/py21cmfast/src/dft.c
```diff
@@ -0,0 +1,169 @@
...
+
+    } // End of Try{}
+
+    Catch(status){
+        return(status);
+    }
+    return(0);
+}
+
+//int CreateFFTWWisdoms(struct UserParams *user_params, struct CosmoParams *cosmo_params) {
```
qyx268, line 156: below can be removed.

src/py21cmfast/src/dft.c
```diff
@@ -0,0 +1,169 @@
...
+        omp_set_num_threads(user_params->N_THREADS);
+        fftwf_init_threads();
+        fftwf_plan_with_nthreads(user_params->N_THREADS);
+        fftwf_cleanup_threads();
+
+        // allocate array for the k-space and real-space boxes
+        fftwf_complex *HIRES_box = (fftwf_complex *) fftwf_malloc(sizeof(fftwf_complex)*KSPACE_NUM_PIXELS);
+        fftwf_complex *LOWRES_box = (fftwf_complex *) fftwf_malloc(sizeof(fftwf_complex)*HII_KSPACE_NUM_PIXELS);
+//
+//#pragma omp parallel shared(HIRES_box) private(i,j,k) num_threads(user_params->N_THREADS)
```
qyx268, line 81: if we dont need to initialize them, perhaps just delete the commented lines

src/py21cmfast/src/BrightnessTemperatureBox.c
```diff
@@ -119,25 +119,7 @@ int ComputeBrightnessTemp(float redshift, struct UserParams *user_params, struct
...
-                LOG_ERROR("Cannot locate FFTW Wisdom: %s file not found",wisdom_filename);
-                Throw(FileError);
-            }
-        }
-        else {
-            plan = fftwf_plan_dft_r2c_3d(user_params->HII_DIM, user_params->HII_DIM, user_params->HII_DIM,
-                                         (float *)vel_gradient, (fftwf_complex *)vel_gradient, FFTW_ESTIMATE);
-            fftwf_execute(plan);
-        }
-        fftwf_destroy_plan(plan);
```
qyx268, line 140: we can remove line 29 where plan is defined

qyx268: also similar lines in other .c files

qyx268 (approved):
Some comments on the INDEX macros In UsefulFunctions.c (see more ref in http://www.fftw.org/fftw3_doc/Real_002ddata-DFT-Array-Format.html#Real_002ddata-DFT-Array-Format).

As explained in this page, I dont think having DIM being an odd number is a problem. For the real array used by in-place fft, it needs to pad 2 more values if DIM is even and 1 more if DIM is odd. This is because, the complex array used by fft needs only half of itself due to symmetries (http://www.fftw.org/fftw3_doc/The-1d-Real_002ddata-DFT.html#The-1d-Real_002ddata-DFT). Therefore, we only need DIM/2 elements if DIM is even, or floor(DIM/2)+1 if DIM is odd.

Therefore, the extra 1 or 2 *DIM*DIM elements are added for in-place calculation, but are never used.

src/py21cmfast/src/BrightnessTemperatureBox.c
steven-murray: Good point. I've removed them now.

src/py21cmfast/src/dft.c
steven-murray: Done.

steven-murray:
Thanks @qyx268, I've made those changes. I moved the indexing macros to their own little file `indexing.c`, and have some nicer documentation at the top trying to explain it.

PR #200, opened 2021-01-21
Fast Fcoll Tables using analytic approximation

JulianBMunoz:
The Fcoll tables require an integral over the Press-Schechter HMF for many densities and variances. This is computationally very expensive. For the case of no power-law in mass within the integral (no feedback, alpha=0) this integral is just an erfc, but there is no exact solution for alpha>0. However, I have found a really good approximation for alpha!=0 (assuming a sharp mass cutoff instead of exponential), which returns the same 21-cm signal to better than 10% precision, and it speeds up the generation of Fcoll tables by a factor of ~20. This is especially important for minihaloes, where the tables have to be calculated for many Mturns. 

The new flag is flag_options->FAST_FCOLL_TABLES, and it's set to automatically true with minihaloes. The science reference will likely be the VAO paper I'm working on (probably an appendix), we can add it later.

As an example, here are the lightcones, global signal, and power spectra for the case with minihaloes turned on and alpha=0.5 for both MCGs and ACGs. For this small box the FAST_FCOLL_TABLES = True case took 900 s, and the old one took 16000 s, a factor of 20 faster. For ACGs only (no minihaloes) the improvement is a factor of 2. There is barely any difference in the results.

image

image

image


src/py21cmfast/inputs.py
```diff
@@ -659,6 +663,17 @@ def INHOMO_RECO(self):
         else:
             return self._INHOMO_RECO
 
+    @property
+    def FAST_FCOLL_TABLES(self):
+        """Automatically setting FAST_FCOLL_TABLES to True if USE_MINI_HALOS. For speed."""
+        if self.USE_MINI_HALOS and not self._FAST_FCOLL_TABLES:
+            logger.warning(
+                "You have set USE_MINI_HALOS to True but FAST_FCOLL_TABLES to False, so the code will be very slow. Setting to true"
+            )
```
BradGreig, line 672: I don't think this should automatically be set if USE_MINI_HALOS is set. It just caused my run to crash when I set FAST_FCOLL_TABLES=False. Which should be a perfectly legitimate option.

I think a default should be selected (I think FAST_FCOLL_TABLES = FALSE) should be the default, with the option of speeding it up with setting this to true. Setting it to always return True when USE_MIN

src/py21cmfast/src/ps.c
```diff
@@ -2217,27 +2252,125 @@ float GaussLegendreQuad_Nion_MINI(int Type, int n, float growthf, float M2, floa
...
+
+      if(MassTurnover_upper <= MassTurnover){
+        return 1e-40; //in sharp cut it's zero
+      }
+
+      double LogMass=log(MassTurnover);
+      int MassBin = (int)floor( (LogMass - MinMass )*inv_mass_bin_width );
+      double MassBinLow = MinMass + mass_bin_width*(double)MassBin;
+      double sigmaM1 = Sigma_InterpTable[MassBin] + ( LogMass - MassBinLow )*( Sigma_InterpTable[MassBin+1] - Sigma_InterpTable[MassBin] )*inv_mass_bin_width;
+      nu_lower_limit = (delta1-delta2)*(delta1-delta2)/(sigmaM1*sigmaM1-sigma2*sigma2)/growthf/growthf;
```
BradGreig, line 2275: I'd re-write this due to redundant calculations throughout.

Consider;

double delta_arg = pow( (delta1 - delta2)/growthf , 2.);

then have; 

nu_lower_limit = delta_arg * (sigmaM1*sigmaM1 - sigma2*sigma2)

You can then repeat this for nu_higher_limit, nupivot1 and nupivot2.

src/py21cmfast/src/ps.c
```diff
@@ -2217,27 +2252,125 @@ float GaussLegendreQuad_Nion_MINI(int Type, int n, float growthf, float M2, floa
...
+
+      LogMass = log(MassTurnover_upper);
+      MassBin = (int)floor( (LogMass - MinMass )*inv_mass_bin_width );
+      MassBinLow = MinMass + mass_bin_width*(double)MassBin;
+      double sigmaM2 = Sigma_InterpTable[MassBin] + ( LogMass - MassBinLow )*( Sigma_InterpTable[MassBin+1] - Sigma_InterpTable[MassBin] )*inv_mass_bin_width;
+      nu_higher_limit = (delta1-delta2)*(delta1-delta2)/(sigmaM2*sigmaM2-sigma2*sigma2)/growthf/growthf;
+
+
+      //note we keep nupivot1 just in case very negative delta makes it reach that nu
+      LogMass = log(1.5e9); //jbm could be done outside and it'd be even faster
```
BradGreig, line 2285: This (along with the other one below) should definitely be defined in Globals.h (or Constants.h). It is something that is unlikely to ever change, but maybe it might under certain conditions so should probably be in Globals.h

src/py21cmfast/src/ps.c
```diff
@@ -2256,8 +2389,69 @@ float GaussLegendreQuad_Nion(int Type, int n, float growthf, float M2, float sig
         .LimitMass_Fesc = Mlim_Fesc
     };
 
-    if(delta2 > delta1){
-        return 1.;
+    if (FAST_FCOLL_TABLES){ //JBM: Fast tables. Assume sharp Mturn, not exponential cutoff.
+
+      double LogMass=log(MassTurnover);
+      int MassBin = (int)floor( (LogMass - MinMass )*inv_mass_bin_width );
+      double MassBinLow = MinMass + mass_bin_width*(double)MassBin;
```
BradGreig, line 2396: Similar comments as above

src/py21cmfast/src/ps.c
```diff
@@ -2256,8 +2389,69 @@ float GaussLegendreQuad_Nion(int Type, int n, float growthf, float M2, float sig
...
+
+      double beta1 = (Alpha_star+Alpha_esc) * 9.0 * (0.5); //exponent for Fcollapprox for nu>nupivot1 (large M)
+      double beta2 = (Alpha_star+Alpha_esc) * 13.6 * (0.5); //exponent for Fcollapprox for nupivot2<nu<nupivot1 (small M)
+      double beta3 = (Alpha_star+Alpha_esc) * 21.0 * (0.5); //exponent for Fcollapprox for nu<nupivot2 (smallest M)
+    //beta2 fixed by continuity.
+
+
+      double corrfact=sigma2/(delta1-delta2)*growthf+1e-10; //small number added to avoid infinities
+      double nucrit_sigma2 = 1.0/corrfact/corrfact; //above this nu sigma2>sigma1, so HMF=0
+
```
BradGreig, line 2422: Could re-write as;

corrfact = sigma2 * pow( (delta1 - delta2)*growth , -1.);
nucrit_sigma2 = pow( corrfact, -2.);

Generally speaking multiplication is computationally faster than division, so often I err on making things multiplications. I think from memory, calling pow for inverse powers is also faster than normal division (although the difference is likely to be extremely minimal).

src/py21cmfast/src/ps.c
```diff
@@ -2274,7 +2468,20 @@ float GaussLegendreQuad_Nion(int Type, int n, float growthf, float M2, float sig
...
 
-void initialise_Nion_General_spline(float z, float min_density, float max_density, float Mmax, float MassTurnover, float Alpha_star, float Alpha_esc, float Fstar10, float Fesc10, float Mlim_Fstar, float Mlim_Fesc){
+
+#include <gsl/gsl_sf_gamma.h>
+//JBM: Integral of a power-law times exponential for EPS: \int dnu nu^beta * exp(-nu/2)/sqrt(nu) from numin to infty.
+double Fcollapprox (double numin, double beta){
+//nu is deltacrit^2/sigma^2, corrected by delta(R) and sigma(R)
+  double gg = gsl_sf_gamma_inc(1.0/2.0+beta,numin/2.0);
+  return gg*pow(2.0,1.0/2.0+beta)/sqrt(2.0*PI);
+}
```
BradGreig, line 2478: re-write as;

double gg = gsl_sf_gamma_inc(0.5 + beta, 0.5*numin);
return gg * pow(2,0.5+beta)*pow(2.*PI,-0.5)

BradGreig (changes requested):
Hey @JulianBMunoz great work on this! I haven't looked at the equations themselves, I've assumed it is all correct. I'll leave that for @qyx268 to verify.

Most of my suggested changes are purely for computational efficiency. Adding in my suggestions made the initialisations of all the interpolation tables another factor of 2 faster, so decent in terms of improvements. However, for increasing HII_DIM, the evaluation of the tables for the main calculation becomes the dominant term. Nevertheless, always good to minimise things as much as possible.

I did not see a factor of 20 improvement on my machine (more like 6-8), though likely didn't do the same size/scale test (mine was HII_DIM = 64, L=100Mpc, down to z = 25.0).

With the 21-cm PS plots you show, am I interpreting things correct that the power is negative? I've never seen negative power from the 21-cm PS from 21cmFAST. Further, in the top 21-cm PS plot, these variations appear to be > 10%. Could you maybe plot the ratio to better highlight the variation. Also, is the variation minimised by considering larger volumes? Curious how much of it might be statistical from low ionisation fluctuations.

src/py21cmfast/src/ps.c
```diff
@@ -2217,27 +2252,125 @@ float GaussLegendreQuad_Nion_MINI(int Type, int n, float growthf, float M2, floa
...
+        fcollres=(Fcollapprox(nupivot2,beta2))*pow(nupivot1,-beta2);
+        fcollres+=(Fcollapprox(nu_lower_limit,beta3)-Fcollapprox(nupivot2,beta3) )*pow(nupivot2,-beta3);
+      }
+
+      }
+
+      if (fcollres < fcollres_high){
+      return 1e-40;
+      }
+      return (fcollres-fcollres_high);
```
qyx268, line 2329: hmm... so effectively, we also have a hard cut for the upper limit...

qyx268 (approved):
The code looks all good. I think the new change has been implemented at all correct spots and has avoided the calculation of ST tables. I only have a few suggestion for speeding up. 

P.S. I can't check the exact form of those equations either. I can't find the fitting functions. Maybe it's worth just posting the PDF in this PR? However, since the test showed it works, I doubt there is any issue with the functions.

Can you maybe do a test with stronger minihalo's contribution? I assumed you used the default one since the global signal's trough is at z=15. Also, what are the redshifts of those PS?

src/py21cmfast/inputs.py
JulianBMunoz: Great comments, thanks! I have moved USE_INTERPOLATION_TABLES to UserParams, and left the default as FALSE. If TRUE it checks that USE_INTERPOLATION_TABLES is TRUE too.

src/py21cmfast/src/ps.c
JulianBMunoz: Very true, I've changed it in both the regular and the MINI versions of the calculation. Thanks!

src/py21cmfast/src/ps.c
JulianBMunoz: I'm not sure it should be a global variable, since it's not something likely to change (since the sigma(M) fit is good to ~10%, which is worse than the errors in the cosmological parameters). I have moved it to constants.h, which is already better than having numbers in the code.

src/py21cmfast/src/ps.c
JulianBMunoz: Good point, I've fixed this too to do fewer operations, and also to be just one line since we don't need corrfact anywhere else.

src/py21cmfast/src/ps.c
JulianBMunoz: Thanks!

src/py21cmfast/inputs.py
JulianBMunoz: Question, are the initial conditions re-ran automatically if anything is changed in UserParams? Or only for DIM and HII_DIM, etc? If it's the former we should not have FAST_FCOLL_TABLES (or USE_INTERPOLATION_TABLES) in there

steven-murray: @JulianBMunoz they are re-run if any of the parameters in `UserParams` are modified, so we're all good.

src/py21cmfast/src/ps.c
JulianBMunoz: Yes, we need hard cut on both ends

src/py21cmfast/src/ps.c
```diff
@@ -2217,27 +2252,125 @@ float GaussLegendreQuad_Nion_MINI(int Type, int n, float growthf, float M2, floa
...
-                integrand += wi_SFR_Xray[i]*Nion_ConditionallnM_GL_MINI(x,parameters_gsl_SFR_con);
-            }
-            if(Type==0) {
-                x = xi_SFR[i];
-                integrand += wi_SFR[i]*Nion_ConditionallnM_GL_MINI(x,parameters_gsl_SFR_con);
-            }
-        }
-        return integrand;
+
+      if(delta2 > delta1){
```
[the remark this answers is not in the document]
JulianBMunoz: Good point, it is duplicated, I just kept it from the old code but I have now removed it

src/py21cmfast/inputs.py
JulianBMunoz: But this is a problem, isn't it? We don't necessarily want to re-run the ICs if FAST_FCOLL_TABLES or USE_INTERPOLATION_TABLES change, right?

Also, I just re-ran the same case with all the new changes and it still takes ~900 s for a tiny box (so its almost entirely table cost), so I did not see any improvement in speed.

steven-murray: Ah, I see what you mean. Yeah, this is a more fundamental problem with the setup, and why I have previously advocated for having an `options` and `params` structure for each major computational component, so that the dependencies can be tracked with more granularity. I think there are ways of tuning this at the moment, but it's also probably not a massive problem.

JulianBMunoz:
Alright, I made the changes, and made plots of the ratio of power spectra old and new. The errors are larger in the first chunk than the last ones, not sure why (this is based on @qyx268's minihalo code, I think the first bin is at low z?). Even then, difference is ~25% at its worst, and around or below 10% for the rest. I'm not sure I would trust extended PS better than that anyway.


ratio_powsp

src/py21cmfast/src/IonisationBox.c
```diff
@@ -468,14 +468,21 @@ LOG_SUPER_DEBUG("average turnover masses are %.2f and %.2f for ACGs and MCGs", b
 LOG_SUPER_DEBUG("minimum source mass has been set: %f", M_MIN);
 
     if(user_params->USE_INTERPOLATION_TABLES) {
+      if(flag_options->FAST_FCOLL_TABLES){
+        initialiseSigmaMInterpTable(fmin(2e4,M_MIN),1e20);
```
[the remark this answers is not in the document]
JulianBMunoz: True, I changed it to a new constants.h variable, MMIN_FAST=1e5 in Msun

src/py21cmfast/inputs.py
JulianBMunoz: Ok this seems like a separate issue we can address at a different point if we so wish. For now it makes sense that FAST_FCOLL_TABLES sticks in the same category as USE_INTERPOLATION_TABLES. Thanks y'all!

BradGreig:
@JulianBMunoz what are the redshifts (and neutral fractions) for the boxes with the ~25% variation. Seems quite a large difference, and could be significant depending on the investigation that you are running. Guess @qyx268 and @andreimesinger would have a better handle on how comfortable they are with such a variation.

I wonder if this could be mitigated somewhat by modifying M_turn in this approximation somewhat. Here, rather than the nominal exponential turn-over you replace it with a hard (step function) cut. Is it possible to calculate (analytically or approximately) a modified M_turn which better mimic's the exponential turn-over. That is find a new M_turn which produces roughly the same result as if an exponential cut was used. Pure speculation here, but I wonder if that would reduce the difference somewhat. Does that make sense?

JulianBMunoz:
Great question, thanks @BradGreig! Those are the lowest z, and the difference is caused due to the regular ACGs rather than the minihaloes (and probably ionization rather than spin temp is my intuition). My guess is that the \sigma^2(R) term becomes large enough at low z that the approximation I'm taking starts to break down (remember this breaks the power-law dependence).

There's a silly workaround that I have found, which is to only speed up the minihalo part of the Fcoll tables and keep the ACGs the same as before. This makes the table generation slower (by x3), but overall it's still ~x5 times faster than the original since the minihaloes take most of the Fcoll computations. Times for a tiny box where (15643 s for no fast, 953 s for all fast, and 2763 s for only MCGs fast). Attached are the ratios of power, they agree at the ~% level now.

I'm looking now to improve this by doing ACGs normally at high z and only turning off the approximation at low z and/or large variances. Will keep you all posted.

I also tried playing with the cutoff btw, since in principle I can plug a log(2) factor out of principle (exp(-log2) =0.5 is where the Heaviside could be said to begin). I found that it makes very little difference, and it does not preferentially fix lower z.

image

JulianBMunoz:
Ok here's one more update. I think this may be a good compromise. I turn off the FAST option for z<10 and sigma2>1, which are the regions where the approximation fails the most. By doing so I only make the FAST tables marginally slower (30%, increasing the time to generate them from 900 s to 1200 s), but it agrees with the standard/slow version to better than 10% (plots attached).

What do you think? @BradGreig @andreimesinger @qyx268, should we keep this "compromise" option? I think the further x2 speed up (from not doing FAST for ACGs at all) is worth it, since I don't know if I trust EPS to better than 10% anyway. 

image
image

I have tested other permutations, like not doing FAST for z<10 for all sigma2 (1696 s), or for sigma2>1 at all z (1803 s). They both result in slightly better agreement (not dramatically better tho), but the cost is almost x2 of the fastest version.

andreimesinger:
i agree that the speed up is worth it, but is your compromise solution here contiguous?  smoothness across these boundaries needs to be strictly enforced.  the 21cm PS isn’t the only metric we necessarily care about, maybe check directly the HMF integrals...

JulianBMunoz:
Hmm, that's a good question @andreimesinger... You mean discontinuity in the z and sigma2 variables, right? Spatially everything is continuous, since we are not cutting in \delta which is the variable that is x-dependent.

I believe that the discontinuity in z is ok, since the tables are recalculated at each z anyway, so it's not a problem to jump from one method to the next at some particular z. The discontinuity in sigma2 may be trickier, since the tables are computed for many sigma2 at fixed z. I do not have a simple solution to this, since the procedure is discontinuous by nature (either we do the FAST approximation or not). If we limit ourselves to z<10 (and all sigma2) the timing is ~1700 s instead of ~1200 s, which is another ~40% cut in speed. It's not clear to me that this matters at all, since the spacing between sigma2s is pretty broad (like 10% step or so if I recall), so it seems to me that it's OK if we switch method from one to the other, but you three probably know the code better than me, what do you think?

BradGreig:
Hey @JulianBMunoz, I'm definitely a lot more comfortable with the sub 10 per cent differences. How sensitive is the approximation to z and sigma2? I worry that for your testing, you are considering only one set of astro parameters. Changing these parameters (and thus the timing of reionisation) could impact your transitions for switching between FAST and full (especially in z I can see this being a possible issue, sigma2 may not be such a large problem).

An important question to consider here is what is your HII_DIM for your timing tests? Very quickly, the actual computation of the tables becomes irrelevant and the longest part is actually their evaluation. This is particularly true for the ACGs. Thus, depending on your answer to your choice of HII_DIM, perhaps applying this approximation to only MCGs would suffice.

andreimesinger:
discontinuity in either could be a problem, potentially resulting in “kinks”, for some parameter combinations at least.

imo, having your approximation by default only for the MCGs might be a good compromise. since we are modeling it as a distinct population, we should still get stable, smooth evolutions.

JulianBMunoz:
Those are both great points! 

A reasonably sized run of HII_DIM=100 takes 3158s total in my laptop with the compromise approximation from above. Given that the tables alone take 1263s that means that even for regular-sized runs, the tables are half of the cost now (remember, they used to be ~15,000 s before!) which isn't too bad then.

I agree that using the approximation just for MCGs is very fair. There is a physical reason for the tables to work better for those too, as explained in the note PDF since the large-\nu part of the integral is the tricky one, and it cancels out when you have a large-mass turnover. 

I will keep the code for ACGs too, but I have written a FLAG_FAST_ATOMIC defined as 0 in Constants.h, but advanced users may want to play with it. 

How does it look now?

BradGreig (commented):
Hey @JulianBMunoz, I agree for now that perhaps its best to split these up (as there is no reason to think the ACGs part wouldn't be used in future).

However, I think your `FLAG_FAST_ATOMIC` should appear in Globals.h instead. This way it can easily be turned on without having to recompile the code for those users wanting to use the feature

steven-murray:
@JulianBMunoz, I'd also change the name of the variable to something like `USE_FAST_ATOMIC` or `DO_FAST_ATOMIC`

JulianBMunoz:
Great suggestions, thank you! I just implemented both.

steven-murray (approved):

steven-murray:
@JulianBMunoz it looks like linting is failing -- you should do 

```
git pull
pre-commit install
pre-commit run -a
```

JulianBMunoz:
Hmm, I run those commands but flake8 and black failed:

flake8...................................................................Failed
- hook id: flake8
- exit code: 1

black....................................................................Failed
- hook id: black
- files were modified by this hook

reformatted /Users/julian/Dropbox/Research/21/fastfcolltabs/github_fast/21cmFAST/src/py21cmfast/inputs.py
All done! ✨ 🍰 ✨
1 file reformatted, 32 files left unchanged.

steven-murray:
@JulianBMunoz that is why the Linting is failing :-). If you run it again, it may have fixed the problem automatically, otherwise, just read the error and you should be able to fix it.

JulianBMunoz:
Aha, I did not realize running it again would fix some stuff on its own, thanks. Let me know if it works now!

PR #203, opened 2021-01-30
remove mkl to fix fft bug in mkl_fft

qyx268:
a quick fix for #179


steven-murray (approved):
Thanks @qyx268. I still have to make sure this fix is applied to the actual conda package as well.

PR #205, opened 2021-02-03
Fix user data

### LINKED ISSUE #190: Installing on Ubuntu 18.04 LTS
I faced some difficulties in getting to run 21cmFAST on Ubuntu 18.04 LTS, but it is running smoothly now. I just wanted to keep the steps I took noted here for future reference and possibly also for others.

After installing anaconda3 on Ubuntu 18.04 LTS, create a new conda environment: `conda create -n myenv python=3.6`

Activate the environment: `conda activate myenv`

Install gsl and fftw: `conda install -c conda-forge gsl fftw`

Other dependencies are automatically installed while installing 21cmFAST using conda-forge as (do not forget to specify the LIB and INC paths before the conda command)

```
LIB=~/anaconda3/envs/myenv/lib INC=~/anaconda3/envs/myenv/include conda install -c conda-forge 21cmFAST
```

21cmFAST is now installed, but I could not run the example scripts because I did not have a "user data" directory in my home. So download or clone the 21cmFAST github package `https://github.com/21cmfast/21cmFAST`, unzip and then move only the `user_data` directory to your home folder. Finally rename this directory to `.21cmfast` (a hidden directory):

```
mv ~/Downloads/21cmFAST/user_data ~/
cd
mv user_data .21cmfast
```

Then the following code should run smoothly:

```
import py21cmfast as p21c
coeval8, coeval9, coeval10 = p21c.run_coeval(
    redshift = [8.0, 9.0, 10.0],
    user_params = {"HII_DIM": 100, "BOX_LEN": 100, "USE_INTERPOLATION_TABLES": True},
    cosmo_params = p21c.CosmoParams(SIGMA_8=0.8),
    astro_params = p21c.AstroParams({"HII_EFF_FACTOR":20.0}),
    random_seed=12345
)
```

steven-murray:
Fixes #190.

Since `conda` doesn't seem to be installing the `user_data/` directory, I figured its easier just to install the `External_tables` as data within the package itself. The config file is still located at the same place, but its default values are hard-coded into the `Config` object, and no file needs to be present (a file will be written to that location if the user changes the defaults and calls `write()`).


BradGreig:
Hey @steven-murray, this doesn't appear to be working. It appears it cannot locate the data as it appears to be looking in the wrong location.

Also, I cannot install this locally (download source code and pip install it). Likely due to some weirdness on my end though. I'll have to track that down at some point...

BradGreig:
@steven-murray, a quick update, the issue appears to be with setuptools-scm (```LookupError: setuptools-scm was unable to detect version for '/Users/bradleygreig/Documents/py21cmFAST/fix-user-data/'```). Which is the location I am attempting to install from.

After some rudimentary googling, I tried a few different versions of `setuptools` and `setuptools-scm` (older and upgrading to newest versions). No luck yet.

BradGreig:
Hey @steven-murray. Ok, I fixed my installation issue (no .git, sigh...).

However, I can confirm that the user data is not written, thus fails on my local machine.

steven-murray:
Hi @BradGreig, I have this working now. There were a couple of issues: 

1. It seems that I needed to put a `__init__.py` in the `_data/` folder to get python to install it properly. 
2. The `input_filename` variable in `initialize_interp_arrays` was being allocated 100 entries, but since the filenames changed, they are sometimes more than 100 in length, and it was breaking (this was difficult to locate...). I upped it to 256.

Concerning `setuptools-scm`:

1. I know that the install will work either with a cloned repo, or by installing from PyPI or conda (these latter ones are NOT git repos, and it works fine). 
2. I would also like it to work when someone just downloads the zip file from github, via a `pip install .`, though I should say that I consider this case much less important: typically if someone wants access to the actual source code, they'll clone. If they just want to install the code, they'll do it from pypi/conda. But it's good to support this regardless. I had thought I looked into this at some stage and it worked. I'll look into it again now.

steven-murray:
I created an issue to remind us to look into the setuptools-scm thing, so it doesn't have to be fixed in this PR.

qyx268:
why is the doc check still pending?

steven-murray:
@qyx268 it was a problem with readthedocs, but seems to have fixed itself now.

qyx268:
sorry, how do I test it using conda install without it being merged into master?

qyx268:
the code looks fine after a quick skim

steven-murray:
@qyx268 I don't think you can conda install it without it being on master, sorry. It is _possible_, but I haven't looked into it. Are you good to approve?

qyx268 (approved):

qyx268:
just approve it. we can do a conda install to see how it goes.

PR #210, opened 2021-02-09
Store all v

### LINKED ISSUE #194: ring sturcture shows up in large-volume simulations
**Describe the bug:**

When we have large-volume simulations, we get ring structures on scales of Gpc, in particularly in the photoionization rate box. I'm thinking if this is due to the 

image

**To Reproduce:**
You just need a large-volume simulation... You can reproduce with this script. 
[run.txt](https://github.com/21cmfast/21cmFAST/files/5638255/run.txt)

**Expected behavior:**

**Details:**
 - OS: Linux nefertem.sns.it 2.6.32-504.hz100.el6.x86_64 #1 SMP Fri Oct 31 20:08:27 CDT 2014 x86_64 x86_64 x86_64 GNU/Linux
 - Python Version: 3.6.5

**Additional context**

Has anyone had large-volume simulations performed with the old 21cmfast or 21CMMC driver? It will be helpful to see if this is a new thing in v3 or it has always been there but missed.

qyx268:
fix #194 
also add output of MFP_box
also introduced 1e15 factor in dzdt and recombination rate for precision


src/py21cmfast/src/IonisationBox.c
```diff
@@ -150,7 +154,7 @@ LOG_SUPER_DEBUG("defined parameters");
         }
     }
 
-    fabs_dtdz = fabs(dtdz(redshift));
+    fabs_dtdz = fabs(dtdz(redshift))/1e15; //reduce to have good precision
```
steven-murray, line 157: I don't quite understand what this does, but it's a MASSIVE change... you sure it makes sense?

steven-murray (commented):
Thanks @qyx268 and well done for finding this incredibly strange error. Please update the CHANGELOG with the bugfix, and answer my one question, and then I'll approve :-)

BradGreig:
Hey @qyx268, to be super pedantic, rather than fixing ZSTEP = 0.2 on the first redshift, you can actually calculate it directly from `ZPRIME_STEP_FACTOR`.

Given that the cadence for the sampling will almost always be set by this, I think it makes sense to do it this way rather than having it fixed. It is highly likely that it doesn't matter, but it would be more consistent with the nominal calculations.

Thus I'd suggest replacing:

```
 if (prev_redshift < 1) //first redshift, just assign ZSTEP = 2
         ZSTEP = 0.2;
```
with
```
 if (prev_redshift < 1) //first redshift, assign ZSTEP using global_params.ZPRIME_STEP_FACTOR
         ZSTEP = ((1. + redshift)* global_params.ZPRIME_STEP_FACTOR - 1.);
```

Which under your setup would result in ZSTEP = 0.3. So not really much of a difference, but looks more like v2.

qyx268:
yep, make more sense

src/py21cmfast/src/IonisationBox.c
qyx268: this is not related to this bug. But when I debug, I found fabs_dtdz is in order of 1e15 while recombination rate is in order of 1e-15, which is also why we have to use double for variables in recombination. Therefore I just added this factor

qyx268:
These failed tests are likely because the existing data were also affected by this bug 

FAILED tests/test_integration_features.py::test_power_spectra_coeval[18-kwargs12]
FAILED tests/test_integration_features.py::test_power_spectra_coeval[12.0-kwargs23]
FAILED tests/test_integration_features.py::test_power_spectra_lightcone[18-kwargs12]
FAILED tests/test_integration_features.py::test_power_spectra_lightcone[12.0-kwargs23]

e.g.
```
tests/test_integration_features.py:61: in test_power_spectra_coeval
    assert np.allclose(
E   assert False
E    +  where False = <function allclose at 0x7fd2f87f7950>(array([1.17942870e+08, 2.94883253e+04, 1.24103858e+04, 5.41815152e+03,\n       1.32032102e+03, 1.28101612e+03, 9.63208391e+02, 5.62934068e+02,\n       4.63296436e+02, 3.82961915e+02, 3.07256894e+02]), array([1.18065178e+08, 2.88373690e+04, 1.13437935e+04, 5.43312344e+03,\n       1.32410393e+03, 1.19082910e+03, 9.23079919e+02, 5.65427179e+02,\n       4.52957544e+02, 3.65026264e+02, 3.13104618e+02]), atol=0, rtol=0.01)
E    +    where <function allclose at 0x7fd2f87f7950> = np.allclose
```

I suggest replacing all or just these 2 test data.

steven-murray:
Yes, I agree, replacing those two is a good idea. Do you know how to run just those two to replace? You should just run `python produce_integration_test_data.py --nums 12,23`

qyx268:
i've updated the data. i'd like to wait for the big simulation to finish so I can confirm the bug is fixed before marking this PR ready for review.

qyx268:
Some one merge the PR please @BradGreig @steven-murray

steven-murray (approved):
Thanks @qyx268 !

PR #211, opened 2021-02-09
Updated generation of relvels to not require extra memory

JulianBMunoz:
I re-used the HIRES_box variable to loop over the 3 v_cb directions and add them in quadrature directly to boxes->lowres_vcb. The high-res vcb box is never used, so there isn't much point in generating it.

This is fast and does not add any memory overload over a no-vcb run. The modification follows what was already in generateICs.c for the perturbation velocities, though with different k weightings (to track the power in vcb).

Plots attached are the old vcb both in lowres and hires (which look very similar as vcb has not power at small scales), and the new vcb. Which is identical to the old lowres.

Screen Shot 2021-02-08 at 9 11 12 PM

Screen Shot 2021-02-09 at 12 15 03 PM

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,6 +9,7 @@ Added
   class, or the Lightcone.
 * Ability to gather all evolutionary antecedents from a Coeval/Lightcone into the one
   file.
+* Fast and low-memory generation of relative-velocity (vcb) initial conditions. Eliminated hi-res vcb boxes, as they are never needed.
 
 Fixed
 ~~~~~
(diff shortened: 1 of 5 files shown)


steven-murray (commented):
This looks great, thanks @JulianBMunoz. Can you add the update to the changelog?

JulianBMunoz:
It's complaining about less coverage because I removed the hires_vcb box tests, but that box doesn't exist anymore. I'm confused...

steven-murray (approved):

steven-murray:
It doesn't make sense to me why the coverage went down. It also seems to be having trouble reporting the test status (even though the tests ran and passed). I'm going to merge with my Admin abilities because everything looks fine.

PR #213, opened 2021-03-03
Add relative velocities to the code, find VAOs, and new molecular-cooling Lyman-Werner feedback

JulianBMunoz:
Alright. This is it. I added the relative velocities (vcb, relvels) to the code. I also gave more flexibility to the parametrization of minihaloes.

In terms of MCGs. The main parameters that one can modify are:
 A_LW; BETA_LW; which set the strength of Lyman-Werner feedback, and are a new fit from simulations 
 A_VCB; BETA_VCB; which set the strength of the feedback from relative velocities.
The vcb parameters seem pretty well known, whereas the LW ones less so. In particular A_LW (set fiducially to 2) can vary. I also made ALPHA_STAR_MINI different from ALPHA_STAR, for extra freedom.

For the relative velocities I have a flag FIX_VCB_AVG, which sets vcb to its average in LCDM, which provides almost the same background evolution as the real thing but no velocity-induced acoustic oscillations (VAOs).

The vcb-dependent part of the feedback described above is not all of the story. The matter power spectrum is also modified, which reduces the amount of haloes (even if they all formed the same amount of stars). This is, as of today, very unwieldy to add, since all our power spectra would have to be position-dependent. Instead, I have added the average-vcb suppression to the power spectrum in an approximate way in ps.c, which gives a ~20% suppression at k=100 h/Mpc (but is smaller at smaller k). It makes some difference but not huge.

Because we have to read the relative velocity boxes, and those only evolve trivially (redshift as (1+z)), they are stored in the initial_conditions, so those have to be read for most calculations. It's no overhead, since those are not deleted as of yet.

I made a python notebook to play with the relvels. Still not fully finished, though.


src/py21cmfast/src/SpinTemperatureBox.c
```diff
@@ -890,16 +893,38 @@ LOG_SUPER_DEBUG("got density gridpoints");
...
+                                else{ //set vcb to a constant, either zero or vavg.
+                                  curr_vcb = 0.0;
+                                }
+                              }
+
+                              *((float *)log10_Mcrit_LW_unfiltered + HII_R_FFT_INDEX(i,j,k)) = \
+                                              log10(lyman_werner_threshold(zp, previous_spin_temp->J_21_LW_box[HII_R_INDEX(i,j,k)],
+                                              curr_vcb, astro_params) );
+
+//JBM: There might be a better way than just reading ALL the ICs? Is the struct populated by defualt?
```
qyx268, line 921: it is on the TODO list (Issue #69 )

src/py21cmfast/src/UsefulFunctions.c
```diff
@@ -758,11 +758,19 @@ double molecular_cooling_threshold(float z){
...
-double lyman_werner_threshold(float z, float J_21_LW){
-    // this follows Visbal+15, which is taken as the optimal fit from Fialkov+12 which
-    // was calibrated with the simulations of Stacy+11 and Greif+11;
-    double mcrit_noLW = 3.314e7 * pow( 1.+z, -1.5);
-    return  mcrit_noLW * (1. + 22.8685 * pow(J_21_LW, 0.47));
+double lyman_werner_threshold(float z, float J_21_LW, float vcb, struct AstroParams *astro_params){
+    // correction follows Schauer+20, fit jointly to LW feedback and relative velocities. They find weaker effect of LW feedback than before (Stacy+11, Greif+11, etc.) due to HII self shielding.
+    double mcrit_noLW = 3.314e7 * pow( 1.+z, -1.5);// this follows Visbal+15, which is taken as the optimal fit from Fialkov+12 which was calibrated with the simulations of Stacy+11 and Greif+11;
+
+    double f_LW = 1.0 + astro_params->A_LW * pow(J_21_LW, astro_params->BETA_LW);
```
qyx268, line 765: The reason those tests with MINI_HALOs on failed is that feedback strength is now different from previous. By default when vcb=0, f_LW is still !=(1. + 22.8685 * pow(J_21_LW, 0.47)). 

I dont have a strong opition here, but either we support the previous strong feedback prescription (e.g. have a flag or sth) or we update the test files. Supporting previous version allow users to easily reprocude previous results, but it's also kind of redundant to do so considering it's just different ways of treating LW feedback...

I'll let other @21cmfast/admins decide

qyx268 (approved):
looks good, let's fix the test and merge it!

qyx268:
the notebook also looks good, just need a bit clean up :)

qyx268:
don't forget to update the change log

src/py21cmfast/src/UsefulFunctions.c
JulianBMunoz: Good point. As you say, the old fit can be obtained by setting astro_params->A_LW=22.8685 and astro_params->BETA_LW=0.47, that is the old Macachek 2001 result. Maybe we can set it to that just for the test?

qyx268: Ah, no idea why I didn't realize that... Yeah, in this case, resetting the default will do

JulianBMunoz: It may make sense to update the "golden" results to the new MCG feedback prescription, which is much lower bc of proper H_2 self shielding. I have no strong feelings about this, since it's a test. @21cmfast/admins?

src/py21cmfast/src/SpinTemperatureBox.c
JulianBMunoz: Good, we'll circle back to this at some point then, thanks!

qyx268:
sorry, how do I test it using conda install without it being merged into master?

qyx268 (approved):
This should work now. I've just changed the default A_LW and BETA_LW back to the Machacek01 value. Then the test passed on my machine. 

I also had to checkout the MINIHALO test files from the master branch. You might have altered this as well?

Let me know if you want me to update the test files and use A_LW=2.0 and BETA_LW=0.6 instead.

src/py21cmfast/src/IonisationBox.c
```diff
@@ -413,15 +417,30 @@ LOG_SUPER_DEBUG("previous density field calculated");
...
                             Mcrit_RE = reionization_feedback(redshift, previous_ionize_box->Gamma12_box[HII_R_INDEX(x, y, z)], previous_ionize_box->z_re_box[HII_R_INDEX(x, y, z)]);
-                            Mcrit_LW = lyman_werner_threshold(redshift, spin_temp->J_21_LW_box[HII_R_INDEX(x, y, z)]);
+                            if (flag_options->FIX_VCB_AVG){ //with this flag we ignore reading vcb box
+                              curr_vcb = VAVG;
+                            }
+                            else{
+                              if(user_params->USE_RELATIVE_VELOCITIES ){
+                                curr_vcb = ini_boxes->lowres_vcb[HII_R_INDEX(x,y,z)];
+                              }
+                              else{ //set vcb to a constant, either zero or vavg.
```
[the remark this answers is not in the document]
JulianBMunoz: Good point, let me make that change!

src/py21cmfast/inputs.py
```diff
@@ -775,14 +798,18 @@ class AstroParams(StructWithDefaults):
         units. Default is `ION_Tvir_MIN`.
     F_H2_SHIELD: float, optional
         Self-shielding factor of molecular hydrogen when experiencing LW suppression.
-        Cf. Eq. 12 of Qin+2020
+        Cf. Eq. 12 of Qin+2020. Obsolete in new version.
```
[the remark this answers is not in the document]
JulianBMunoz: I see. Still, the new prescription includes shielding self-consistently from the sims, so the user should not play with that parameter unless they go back to Macachek+01, right?

> I also had to checkout the MINIHALO test files from the master branch. You might have altered this as well?

Yeah I tried to regenerate those with the new defaults, but my output must've been broken. Thanks for fixing those!

> Let me know if you want me to update the test files and use A_LW=2.0 and BETA_LW=0.6 instead.

I think we should change the defaults to be A_LW=2 and BETA_LW=0.6, since the Macachek values produce nearly a factor of 10 more feedback, so it's not very realistic.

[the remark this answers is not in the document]
qyx268: You mean they consider self-shielding of LW? I didn't realize that. I this case, we can probably remove F_H2_SHIELD, or simply emphasize here that users should only use this parameter for extra shielding... You could decide then

src/py21cmfast/src/IonisationBox.c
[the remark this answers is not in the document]
qyx268: okay. I'll regenerate the test files after you finish it.

src/py21cmfast/inputs.py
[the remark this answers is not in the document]
JulianBMunoz: I kept it with extra comments to the user. No reason to remove it entirely, I think, it may be useful for someone some day.

src/py21cmfast/src/IonisationBox.c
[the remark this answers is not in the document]
JulianBMunoz: I created one more test (last of OPTIONS) with relvels on, since we did not have any with that option, the data for it will have to be generated too

qyx268:
should be fine now, just need others to approve before launching some simulations :)

qyx268:
no idea what's going one... macos passed all but not ubuntu. perhaps we shouldn't set atol=0?

qyx268:
actually no, i generated them on a linux machine... 

Linux ananke.sns.it 3.10.0-327.el7.x86_64 #1 SMP Thu Nov 19 22:10:57 UTC 2015 x86_64 x86_64 x86_64 GNU/Linux

steven-murray:
Perhaps it has something to do with whatever breaks when optimization is off?

JulianBMunoz:
I tried re-running test 12 in my MacOS machine (with -Ofast so it works) to see if it’d break on MacOS or Ubuntu when testing, andddd it breaks on both. So Yuxiang-ran tests somehow passed MacOS but not Ubuntu, and mine passed neither.

steven-murray:
@JulianBMunoz that's odd. The idea is that each test should be using the same seed, so results should be *very* close. But perhaps different systems/compilers make the same seed produce different random numbers?

Perhaps we should make these tests a little more relaxed so that the random seed doesn't play such a large role. OR figure out how to make sure the same thing is run every time...

Perhaps we can _provide_ the ICs, and then just do the computations on top?

JulianBMunoz:
Hmm, is that what is happening? If so, I don’t understand why it’s breaking only for tests with MINIHALOES.

steven-murray:
Those might be breaking because those are the ones you re-produced, because that's the stuff you changed here. It might be that the others would fail if you re-produced them.

JulianBMunoz:
Hmm, when I generate the same test with NTHREAD=1 and NTHREAD=4, I get the same power spectrum that varies by 0.3%. This is the power spectra for test12:

NTHREAD=1:
[2.32603886e+08 1.86465967e+06 2.03414023e+06 1.98311154e+06
 1.93381779e+06 1.88005826e+06 1.83310486e+06 1.80510628e+06
 1.78711569e+06 1.78826954e+06 1.79904943e+06 1.81606692e+06
 1.83490765e+06 1.82641995e+06 1.83666270e+06 1.86456612e+06
 1.89859656e+06 1.91900251e+06 1.94510113e+06 2.00988394e+06
 2.05483402e+06 2.10997434e+06]

NTHREAD=4:
[2.32624242e+08 1.87111203e+06 2.04092560e+06 1.98983398e+06
 1.94044783e+06 1.88660057e+06 1.83956649e+06 1.81151918e+06
 1.79350052e+06 1.79465493e+06 1.80545610e+06 1.82249941e+06
 1.84137330e+06 1.83287139e+06 1.84313575e+06 1.87108590e+06
 1.90516624e+06 1.92561937e+06 1.95176243e+06 2.01663086e+06
 2.06171327e+06 2.11681528e+06]

So can it be the ICs? At least within MacOS it is consistent...

steven-murray:
So, you get power spectra within tolerance for different numbers of threads on one machine, but on difference machines (with same number of threads) it goes out of tolerance.

I have to wonder whether the random numbers are different. I'm pretty sure the random numbers are printed out with SUPER_DEBUG or something -- you could turn that on locally and on Github Actions to see?

BradGreig:
The whole test suite should have `NO_RNG = True`, meaning there should be no random number generation. This was the point of this flag, to minimise this kind of issue.

Variations in summed numbers across one or multiple threads can be due to rounding errors in summations. Depends on if the variable is a float/double. With multiple threads the order of summation will differ from that of a single thread, and thus you can have differences from cumulative rounding errors.

Thus a variation of ~0.3% isn't really that concerning to me.

steven-murray:
@BradGreig, I think the problem here is that 0.3% is *not* all that's going wrong. The tests on the CI are yielding much greater differences. It's hard to put this down to cumulative rounding errors. It might be completely different numbers though?

Will anything break with `NO_RNG` set? I mean, won't the ICs look super weird? Will that break things physically?

JulianBMunoz:
As a quick note, I was going thru the errors in my commit and I believe the Ubuntu 3.6 one was agreeing well with my MacOS one in test 12 (and passing), whereas the MacOS one didn't. The thing that broke in Ubuntu was test 25 (which I did not touch)

BradGreig:
@steven-murray I agree that something else is likely the issue, my point was to highlight that it could not be the ICs, as these should be the same across any format (as these are generated without RNGs).

Without question the ICs will look weird and certainly things won't make any sense physically, however, the point is to have reproduceable results across multiple formats for stable tests. As otherwise we were running into differences everywhere. Given the tests are on small boxes etc., all the physics would still be broken none the less.

steven-murray:
@BradGreig ahhhh I see what you mean -- it's already got `NO_RNG` on! I had forgotten we did that :-)

OK, so it's not RNG stuff that's breaking...

src/py21cmfast/wrapper.py
```diff
@@ -1637,6 +1637,7 @@ def ionize_box(
             previous_ionize_box,
             spin_temp,
             pt_halos,
+            init_boxes,  # JBM: added this
```
steven-murray, line 1640: can take out the comments that you added it... we can `git blame` it to get this info :-)

steven-murray (commented):
@JulianBMunoz there's really just one thing I don't think is quite right. When checking the `UserParams._defaults_["USE_RELATIVE_VELOCITIES"]` it does not check if relvels are being used -- it just checks whether they're on by default. To actually check if they're used would require passing the value from an *instance* of user_params every time we instantiate a `FlagOptions`. Something similar is done to `AstroParams` with `INHOMO_RECO` (see eg. line 1469 of `wrapper.py`). 

In the long run this is clearly not very scalable, and we should think of something more generic we can do.

JulianBMunoz:
Thanks @steven-murray! What can we do to check the actual instance passed rather than the default? I just want the user to know that minihaloes with no relvels will have the wrong evolution, but the two parameters live in different structures...

steven-murray:
Have a look at that line I mentioned. The problem is you should do that
*every* time the flag options is defined (in all of the functions...
spin_temp, ionize_box, etc), which is a bit tedious.

steven-murray (approved):
Beautiful.

PR #215, opened 2021-03-15
Fixed relvel generation with hires PT

JulianBMunoz:
Code was crashing with "`PERTURB_ON_HIGH_RES`": True, because the `boxes->lowres_vcb` loop indices were set to `dimension` rather than `HII_DIM`


JulianBMunoz:
I put this on the slack, but with this fix the lightcone works well. Plots are the perturb `PERTURB_ON_HIGH_RES = False` and `True` for comparison.

image
image

andreimesinger (approved):

Issue #189, opened 2020-10-30
More informative error 


qyx268:
**Is your feature request related to a problem? Please describe.**

It came to us (@JulianBMunoz ) that when setting BOX_LEN to a very small value (e.g. 25 for testing codes), the program crashes and reports 
`ParameterError: 21cmFAST does not support this combination of parameters.`

**Describe the solution you'd like**

It should be more informative, e.g. providing information of which part of the parameter set is not supported (in this case, BOX_LEN cannot be this small)

**Describe alternatives you've considered**

**Additional context**

@JulianBMunoz  was also trapped by setting `USE_MINI_HALOS=True` (which automatically set `USE_MASS_DEPENDENT_ZETA=True`) without setting `USE_INTERPOLATION_TABLES=True`, supporting the idea proposed in #186 that we should provide information of this...

steven-murray:
Agreed. That is a poor error message!

dprelogo:
I'm creating a small database, pulling samples from the following [prior.pdf](https://github.com/21cmfast/21cmFAST/files/5824360/prior.pdf) and getting the same error in ~5-10% of cases: `py21cmfast._utils.ParameterError: 21cmFAST does not support this combination of parameters.`
In the following file are parameters used for runs that failed. Not specified are default ones. [failed.txt](https://github.com/21cmfast/21cmFAST/files/5824358/failed.txt)

Do you see something weird about those runs?

BradGreig:
Hi @dprelogo, there are no obvious issues with the parameter values themselves. I tried running the second parameter combination from the failed list and it worked without issue on my machine (i.e. laptop). Which version are you running?

It is possible that there is an issue with the parameter combination and/or the evaluation of one of the interpolation tables. I would recommend running one of the failed models with full logging (compiling with LOG_DEBUG etc.) to try and see if that helps identify where it occurs (ionisation/spin temperature etc.).

steven-murray:
Also related to #134

dprelogo:
Hi @BradGreig, I followed your advice and made a run after compiling with `LOG_LEVEL=DEBUG`, for one set of parameters, using different machine and setup. I'm attaching the (truncated) [output](https://github.com/21cmfast/21cmFAST/files/6038013/simple_run.txt) of a simple [run](https://github.com/21cmfast/21cmFAST/files/6038048/simple_run_script.txt) which fails on both machines.

[...]

Issue #134, opened 2020-04-10
[Feature Req.] Better Exception Handling


steven-murray:
Currently we have exception handling based on a *convention*: any C API-level function should return an int (the error status), and should have the entire body wrapped in a `Try{} Catch{}`. This is fine, and works. However, given that it's a pattern, we could shift it to be enforced via some kind of wrapping C or Python function. 

I'm not extremely well-versed in the limitations of C functions, but I feel like some kind of variadic function could be defined that takes a pointer to a function as input, and wraps the call in the Try/Catch, and adds the status etc. In Python, we'd just write a wrapper which calls the basic function _through_ the wrapper.

BradGreig:
Hey @steven-murray, have you had any more thoughts on the implementation of this?

I have been thinking it might be worth my time sitting down and going through the code updating error messages etc. At the moment, an error message is output through the logging, then every parameter is dumped and then finally the uninformative error message (e.g. `21cmFAST does not support this combination of parameters.`).

I suspect in most instances people will not scroll up beyond all the dumped numbers to see the more informative error that is output through `LOG_ERROR`.

Is there a way for the `LOG_ERROR` message to be passed to python so that we get something more informative at the end of error message.

The other way I was thinking of this is to make the current exception coding considerably more granular to try and encompass a broader range of potential issues that might make the output python exception more informative.

If you think increased granularity in the exceptions is useful regardless, I can go ahead and do it in the meantime.

steven-murray:
Hi @BradGreig. No, I have not had time to think about this more. 

It would be awesome to pass back the logged error message to python, but I think it will be very tricky. We can think about making the return type of the `Throw{}` statement to be a `char*`, in which case we can pass back arbitrary strings as the message. But I have a feeling this will be complicated.

I think more granularity is always good. When thinking about this, we should think about `Exception` class inheritance in Python. The reason I have it set up the way it is now is that `ParameterError` stuff should *not* cause a full exit when running MCMC -- you just give -inf for that likelihood and move on. Whereas fatal errors should cause the whole thing to crash. If you want to make the parameter errors more specific, they should inherit from `ParameterError`, so that we can still catch on `ParameterError` in 21cmMC.

BradGreig:
Hi @steven-murray, ok I'll take a look at how this is currently implemented for `ParameterError` and keep it in mind if I choose to expand the error reporting. Might get to this in the next couple of days.

BradGreig:
[...]

PR #219, opened 2021-03-19
More informative errors

BradGreig:
Hey @steven-murray, I have taken a first stab at updating the granularity of the error coding and modifying the error messaging to try and make them a bit more informative. As I mentioned I did away with the `ParameterError` thing entirely, turning everything into a `FatalCError`.

Obviously the long `elif` code is ugly, but just temporary for now. Let me know what you think.

In some places I added to the output error through LOG, in particular including information known from some recent failings with the PhotonCons option (under certain parameter combinations).

Also, I have added some error checks for some AstroParams, causing it to exit with an informative error as to why.

I'm contemplating adding a warning for the other parameters when they are outside the usual parameter ranges we have used in publications. The idea would be that if outside that range, we haven't tested if they'll pass/fail. So warn the user that this might be the case (that their parameter choice is outside of the ranges it has been tested on, but could still be fine).

### DIFF
--- a/src/py21cmfast/inputs.py
+++ b/src/py21cmfast/inputs.py
@@ -875,3 +875,32 @@ class AstroParams(StructWithDefaults):
     def X_RAY_Tvir_MIN(self):
         """Minimum virial temperature of X-ray emitting sources (unlogged and set dynamically)."""
         return self._X_RAY_Tvir_MIN if self._X_RAY_Tvir_MIN else self.ION_Tvir_MIN
+
+    @property
+    def NU_X_THRESH(self):
+        """Check if the choice of NU_X_THRESH is sensible."""
+        if self._NU_X_THRESH < 100.0:
+            raise ValueError(
+                "Chosen NU_X_THRESH is < 100 eV. NU_X_THRESH must be above 100 eV as it describes X-ray photons"
+            )
+        elif self._NU_X_THRESH >= global_params.NU_X_BAND_MAX:
+            raise ValueError(
+                """
+                Chosen NU_X_THRESH > {}, which is the upper limit of the adopted X-ray band
+                (fiducially the soft band 0.5 - 2.0 keV). If you know what you are doing with this
+                choice, please modify the global parameter: NU_X_BAND_MAX""".format(
+                    global_params.NU_X_BAND_MAX
+                )
+            )
+        else:
+            if global_params.NU_X_BAND_MAX > global_params.NU_X_MAX:
+                raise ValueError(
+                    """
+                    Chosen NU_X_BAND_MAX > {}, which is the upper limit of X-ray integrals (fiducially 10 keV)
+                    If you know what you are doing, please modify the global parameter:
+                    NU_X_MAX""".format(
+                        global_params.NU_X_MAX
+                    )
+                )
+            else:
+                return self._NU_X_THRESH
--- a/src/py21cmfast/inputs.py
+++ b/src/py21cmfast/inputs.py
@@ -904,3 +904,11 @@ class AstroParams(StructWithDefaults):
                 )
             else:
                 return self._NU_X_THRESH
+
+    @property
+    def t_STAR(self):
+        """Check if the choice of NU_X_THRESH is sensible."""
+        if self._t_STAR <= 0.0 or self._t_STAR > 1.0:
+            raise ValueError("t_STAR must be above zero and less than or equal to one")
+        else:
+            return self._t_STAR
--- a/src/py21cmfast/_utils.py
+++ b/src/py21cmfast/_utils.py
@@ -39,7 +39,7 @@ class FatalCError(Exception):
 class ErrorIO(FatalCError):
     """An exception when an error occurs with file I/O."""
 
-    default_message = "Expected file could not be found! (check the LOG for more info"
+    default_message = "Expected file could not be found! (check the LOG for more info)"
 
 
 class GSLError(ParameterError):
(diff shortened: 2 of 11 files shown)


BradGreig:
Obviously at this point I don't really care if it passes/fails the tests as I probably haven't checked everything in detail regarding the `ParameterError` usage.

BradGreig:
It's ok, I know why these are failing. I wanted the whole suite to run in case I messed something else up.

steven-murray:
Hey @BradGreig, yeah I'll have a look, hopefully today (I'm pretty swamped though).

src/py21cmfast/inputs.py
```diff
@@ -875,3 +875,40 @@ def R_BUBBLE_MAX(self):
...
+            raise ValueError(
+                """
+                Chosen NU_X_THRESH > {}, which is the upper limit of the adopted X-ray band
+                (fiducially the soft band 0.5 - 2.0 keV). If you know what you are doing with this
+                choice, please modify the global parameter: NU_X_BAND_MAX""".format(
+                    global_params.NU_X_BAND_MAX
+                )
+            )
+        else:
+            if global_params.NU_X_BAND_MAX > global_params.NU_X_MAX:
```
steven-murray, line 896: change this to `elif`

steven-murray (commented):
Thanks @BradGreig, this adds so much good stuff.

In general, I definitely agree with the increased granularity, and the extra information in the LOG_ERROR messages. 

I also really like sending specific error messages in the python, given which error occurred. That is incredibly helpful (and was always my intention).

I do think it still makes sense to keep two major heirarchies of errors in python -- one that should definitely exit, and one that is softer (and the user can decide if it exits). We don't have to _enforce_ this, just let the user decide in 21cmMC. Also, given that you're now sending different messages per error, it might just make sense to define Python exceptions for each case, each with its default message? They can be subclassed from FatalCError and NonfatalCError :-)

BradGreig:
Hey @steven-murray thanks for the feedback. Yep, I knew that certainly wasn't the best way to do it, and knew you'd point me in the right direction (re. subclasses).

I agree with the planned hierarchy for errors. It's just that my preference is to have these exit by default. Certainly all for the option of suppressing/bypassing these. At some level I feel like the majority of these errors would be weird edge/corner cases of parameter combinations, which should be acceptable, it's just that they haven't been robustly checked. Thus, I feel the more that can be caught, reported and dealt with the more robust the code itself would become to these errors.

src/py21cmfast/inputs.py
BradGreig: But this `if/else` statement is not `NU_X_THRESH` based, so it doesn't make sense to be an `elif`. It's to catch something the user might incorrectly set with `global_params`. It may be that this is not the correct place to have this check, but couldn't think where else to put it.

BradGreig:
Hey @steven-murray, I tried implementing the subclass suggestion. I'm not sure this is exactly what you meant, as I still ended up with a large `if/elif` logic branch. I think I must be missing something...

steven-murray (commented):
@BradGreig this is almost there, I think. 

To make it a little more compact, we can do two things. First, define the `FatalCError` like such:

```python
class FatalCError(Exception):
    default_message = "21cmFAST is exiting"
    def __init__(self, msg=None):
        super().__init__(msg or self.default_message)
```

Then each subclass can just be, eg.:

```python
class ErrorIO(FatalCError):
    default_message = 'Expected file could not be found...'
```

Then in the `_process_exitcode` function, you can simplify by just doing something like:

```python
if exitcode:
   try:
      raise {IOERROR: ErrorIO, VALUEERROR: ErrorValue, ...}[exitcode]
   except KeyError:
      raise FatalCError("Unknown error in C. Please report this error!")
```

I also still recommend making several of these errors derive from something like `ParameterError`. Then we should open a PR on 21cmMC to change the default behavior to exit on these errors (but supply a flag in which those are just output as warnings).

BradGreig:
Hey @steven-murray thanks for the further feedback. I have implement these, and switched some of these errors to `ParameterError`. Let me know if there is anything further.

With these current errors I'm not entirely convinced that these should be able to be switched to warnings to continue an MCMC. Mostly because it is hard to know whether ignoring the problematic errors is impacting the posteriors. However, I guess that's up to the user whether that is a smart thing to do.

steven-murray (approved):
Hey @BradGreig. I'm very happy with this. I made a couple more very small comments. You can go ahead and un-draft the PR, and merge once those two things are updated.

I created https://github.com/21cmfast/21CMMC/issues/52 to track this on the 21cmMC side.

BradGreig:
Hey @steven-murray, thanks for the comments. I have updated those accordingly. I was going to create the 21CMMC PR once this was finalised, but it seems you did it already.

Just a quick question. Is this just to be merged into the dev version? Basically, should it be lumped into that under a minor version release? Or should it just be kept separate. To me it makes sense that it should all go into the one minor version update, just wasn't sure of your plans.

steven-murray:
Thanks Brad.

Yeah, this should be merged into master, and we'll cut a full 3.1.0 release
ASAP. The next thing we release has to be a minor version (i.e. 3.1.0)
because we have feature updates already in master. The update I did to the
versioning a while back can really kick in after that -- then we can easily
do smaller bugfixes *after* more serious feature updates by branching
correctly.

Now that we have some really cool stuff in master, we should really push on
getting 3.1.0 released.

BradGreig:
Hey @steven-murray thanks for the clarification re. versioning. I might push over the next day or so to have the functions to estimate memory usage included in this same minor version release then. I'll let you know in a day or so whether or not that actually looks like happening.

BradGreig:
This addresses #189. Also related to #134, however, that should probably remain open as a reminder of the need to keep thinking about how best to deal with this.

PR #224, opened 2021-04-08
Improve memory usage

### LINKED ISSUE #63: Minimise memory usage for InitialConditions class

BradGreig:
This PR aims to minimise memory usage. It takes in the flag that has been introduced in #220 (`MINIMIZE_MEMORY`).

It improves memory usage in two places, only one of which uses this new flag. The first is in generating the ICs, where instead of creating 6 new boxes for the components of the 2LPT, I instead temporarily store the i=j components in the hires_vi_2LPT boxes which have been allocated (but not populated) at this point (they get filled after) and only need a single box for the current i>j component. This results in 5 less boxes (of size DIM). As this minimises memory without additional CPU/IO this change is to always be implemented.

The one usage of `MINIMIZE_MEMORY` so far is in `SpinTemperatureBox.c`. It is only used when `USE_MASS_DEPENDENT_ZETA=True` and `USE_INTERPOLATION_TABLES=True`. It results in only allocating one box for the current smoothed density at the expense of `global_params.NUM_FILTER_STEPS_FOR_Ts` additional FFT's per redshift (but having `global_params.NUM_FILTER_STEPS_FOR_Ts - 1` less boxes in memory). This doesn't seem to add too much in terms of overall computations, especially if `USE_FFTW_WISDOM=True`.

Closes #63


src/py21cmfast/inputs.py
```diff
@@ -452,6 +452,9 @@ class UserParams(StructWithDefaults):
         is considerably faster than when performing integrals explicitly.
     FAST_FCOLL_TABLES: bool, optional
         Whether to use fast Fcoll tables, as described in Sec X of JBM XX. Significant speedup for minihaloes.
+    MINIMIZE_MEMORY: bool, optional
+        If set, the code will run in a mode that minimizes memory usage, at the expense
+        of some CPU/disk-IO. Good for large boxes / small computers.
```
qyx268, line 457: no expense of disk-IO since you are not reading in unfiltered_box from the disk again.

qyx268 (approved):
great improvement!

src/py21cmfast/inputs.py
BradGreig: But this is the same flag as created by @steven-murray with #220. Therefore, there will be additional IO.

steven-murray: Yup, this is used on the python side as well, and will involve extra IO.

steven-murray:
I'd approve if it was ready for it :-)

BradGreig:
Yeah, I'll only move it to a full PR once #220 gets pulled into master and I merge all that in.

BradGreig:
Hey @steven-murray, I added the test data and fixed up the original linting issue, however, it is still failing the linting. I tried re-running the linting tests, which gave the existing error. The previous error was a timing out issue. Might have been an issue on their server end or something...

Issue #226, opened 2021-04-19
Calculating LFs fails when interpolating between minihalos + atomic cooling halos


charlottenosam:
**Describe the bug:**

Using the `calculate_luminosity_function` wrapper, there is an error when when interpolating between the LFs minihalos and atomic cooling halos

`ValueError: x and y arrays must be equal in length along interpolation axis.` at [line 563](https://github.com/21cmfast/21cmFAST/blob/ec274c24502ac6dbe14a5a1e78a75d73c402edfd/src/py21cmfast/wrapper.py#L563) 

```
lfunc_all[iz] = np.log10(10 ** (interp1d(Muvfunc, lfunc, fill_value="extrapolate")(Muvfunc_all[iz]))
                + 10** (interp1d(Muvfunc_MINI, lfunc_MINI, fill_value="extrapolate")(Muvfunc_all[iz])))
```

Also fails when calculating `Mhfunc_all[iz]` just below.

**To Reproduce:**
```
import py21cmfast as p21c
muv, mhalo, lf = p21c.compute_luminosity_function(redshifts=[7, 8, 9], nbins=100, component=0,
                                                  flag_options={'USE_MINI_HALOS':True},
                                                  mturnovers=[7.,7.,7.], mturnovers_mini=[5.,5.,5.])
```

**Expected behavior:**
Expect to produce arrays for muv, mhalo, lf with shapes (len(redshifts), nbins)

NB the above does work fine for component=1 or =2, which calculate minihalos and ACHs only so it's just the interpolation between the two that's failing.

**Details:**
 - OS: MacOS 10.14.6
 - Python Version: 3.9.2
 - Relevant package versions: scipy 1.6.0, numpy 1.20.2

**Additional context**
The fix is just to use the `[iz]` index for everything inside the `interp1d`, and I can make a pull request #227, but I just want to check other people reproduce this error and it's not anything to do with my setup.

It would also be good to add some tests for `calculate_luminosity_function` with USE_MINI_HALOS=True

BradGreig:
Hi @charlottenosam, I can confirm I receive the same error that you are seeing (for `component=0`).

I had a quick look over #227 and this looks fine to me, but I'll leave it to @qyx268 to check that that is the correct thing to do (given he implemented it).

And yes, more tests are always welcomed!

PR #227, opened 2021-04-19
fix interpolation of LFs between ACHs and minihalos

charlottenosam:
Fix for bug #226 in making LFs with both ACHs + minihalos (`component=0`).

Added index for each redshift, and fixed array creation for Mhfunc_all to create the expected LFs. Added one test for LFs with minihalos.


qyx268:
This does look like a bug and the PR fixes it.

qyx268 (approved):

PR #230, opened 2021-04-28
A fix to smooth out the visible kinks in the Lyman-alpha flux

BradGreig:
This is a fix to smooth out visible kinks in the Lyman-alpha flux that can then appear in the spin and brightness temperature.

The cause of this is on large radii the contribution to the Lyman-alpha flux can go to zero. This isn't itself a problem, however, as we have a relative small number of finite (`NUM_FILTER_STEPS_FOR_Ts = 40`) filtering radii, across successive redshifts the number of filtering radii that can contribute to the total signal can increase causing a discontinuity.

This fix finds the two radii where the signal goes from non-zero to zero, and interpolates in between these two filtering radii (i.e. redshifts) to determine on what radii (redshift) the signal should go to zero. It then approximates the expected signal as being some weight of the previous radii contribution rather than being zero. The weight is determined by the number of filtering radii that produce a non-zero contribution between the two larger radii where the signal transitions from non-zero to zero. In effect, the kink is smoothed out over multiple redshift snapshots by having the weight gradually increase from zero to unity rather than suddenly increasing at a single redshift.

It does not notably alter any physical quantity (seems to be less than 0.1 per cent of even less), but smooths out the clearly visible kink.


BradGreig:
Clearly it has some impact on the 21-cm PS, I didn't check that as I was only looking at the globally averaged quantities.

Interestingly though not all tests using the spin temperature failed. Perhaps the failed tests just happened to coincide where this kink feature was most prominent. Will go away and double check things are sensible before we update the test data.

steven-murray (approved):
Thanks @BradGreig, this looks great, and thanks for the extensive in-code comments!

Yeah, let's inspect the tests that are failing just to make sure nothing untoward is happening. Unfortunately I haven't gotten #220 merged yet (been so busy!) because it introduces the possibility of plotting power spectra for many of the intermediate products. But I guess you can look at those yourself anyway.

BradGreig:
It can produce ~2-3 per cent changes in the 21-cm PS. The amplitude of the change is dependent on the location of the kink in the original test data. If the power spectrum is measured across the box where the kink is present in its maximum, the resultant deltaTb will differ (which filters into the 21-cm PS as it is divided through by the volume averaged deltaTb). You can see that even some tests failed the global brightness temperature check, as the original data would have a kink, whereas this fix will remove the kink. I think also the 21-cm PS varies more because the test data are generated without a RNG for the ICs. This kink would increase the fluctuations as it differs in amplitude across successive redshifts.

I think all test data which use the spin temperature calculation should be updated. Even if they passed the tests currently. I wasn't sure on the policy for generating new test data. @steven-murray, do we generate new test data on a specific machine for consistency? Or do we just do it locally and push it.

steven-murray:
Hi @BradGreig, you should just run it locally and push it. You can produce just the nums you want by using `--nums=0,1,10,...`. I will have to run these _again_ in #220 :-)

BradGreig:
Hmm, that's odd, it still fails after updating the test data. Will need to investigate locally...

qyx268 (approved):

Issue #195, opened 2020-12-03
velocity on y and z direction seems odd


qyx268:
**Describe the bug:**

Velocity components on z and y direction looks elongated. The co-evolve box of density, vz, vy and vx is shown below

100Mpc box:
![image](https://user-images.githubusercontent.com/15994713/101075885-e5203780-35a2-11eb-82fd-ca70e60ee273.png)
1Gpc box:
![image](https://user-images.githubusercontent.com/15994713/101076025-126ce580-35a3-11eb-9d04-ea11d01febef.png)

**To Reproduce:**
Just add new component in `PerturbedField` to also output velocity on y and x direction (see branch store_allV)

**Expected behavior:**

**Details:**
 - OS: Linux nefertem.sns.it 2.6.32-504.hz100.el6.x86_64 #1 SMP Fri Oct 31 20:08:27 CDT 2014 x86_64 x86_64 x86_64 GNU/Linux
 - Python Version: 3.6.5

**Additional context**

Better to first check with 2LPT off. 

Again, do we have similar outputs from previous version of 21cmFAST?

qyx268:
I dug a bit. looks like the pattern comes from the IC.

Top: hires_v[zyx]
Middle: hires_v[zyx]_2LPT
Bottom: hires_v[zyx] - hires_v[zyx]_2LPT
Left to Right: vz, vy, vx
![IC_highres_v](https://user-images.githubusercontent.com/15994713/101095088-ae0b4f80-35bd-11eb-9eb9-74b7a5eb8a2c.jpg)

The density looks okay though...
Left to Right: projection of yz, zx and xy
![IC_highres_density](https://user-images.githubusercontent.com/15994713/101095423-31c53c00-35be-11eb-9f3c-9bb7b3d56897.jpg)

andreimesinger:
this looks like it might be a memory /fft /initialization issue after the vx calculation…  vx is the first field to be computed, and then vy, then vz.  the same code is used.  but it could be that things are not being reset properly after the first vx

qyx268:
that's what i thought. but can't find any trace of it

andreimesinger:
is it also true if you turn off hi res?

the only other thing i can think of are the complex conjugate adjustments in init, but i doubt there is a bug that hasn't been found here...

qyx268:
[...]

PR #231, opened 2021-05-04
Perturb Field changes (possibly in a wrong way) when using different number of threads

qyx268:
Ivan has been seen changing velocity fields when using different `N_THREADS`. He is using the `all-vel-components` branch, which is slightly deviated from the master.

e.g. Here, the only parameter that changes is `N_THREADS`, but you see velocity y and z change in the two results.
image
image

I think we do expect the fields to change when using a different number of threads, because the random seed changes. i.e. line 156-185 in `src/GenerateICs.c`. However, the velocity output shown above when using 4 threads could potentially be incorrect, judging from e.g. that the small scale fluctuations seem to get washed out. @andreimesinger what's your opinion on this?

The PR only puts the two missing shared variables in `src/GenerateICs.c`, but I doubt this is where the problem come from. 

Before spending more time into this, here are something worth checking first for Ivan:
1. check the current output from `all-vel-components` to see how IC behaves;
2. re-run the Nthread=4 model with the same fix shown in this PR included in `all-vel-components`;
3. re-run the two comparison models with `NO_RNG=True`, to see if the problem stays when not varying the random seed;


andreimesinger:
this reminds me of the previous discussion here:
https://github.com/21cmfast/21cmFAST/issues/195 <https://github.com/21cmfast/21cmFAST/issues/195>

single-cell slices along the velocity fields are not particularly illuminating and are impacted a lot by cosmic variance, which should change when you change seed.  one can compute the variance of the box or PS if you want, but I doubt this is a bug…  note also that the color maps are different.

steven-murray:
I would definitely suggest running with `NO_RNG=True`. @JulianBMunoz has previously noticed that even with this on there can be some slight differences when using a different number of threads due to rounding errors, but I doubt that is the problem here.

IvanNikolic21:
Hi, to further comment on the issue. On the top, I'm showing a slice of the density cube for N_THREADS=1 on the left and N_THREADS=4 on the right. On the bottom, I'm showing a sum of the velocity_z cube along the z-axis. Both velocity plots share a colorbar, of course. I've tried using NO_RNG=True and a similar thing happens. I've also inspected the InitialConditions file and saw no problem (summing over the z-axis for velocity_z gives 0). I should also note that the same thing happens for all velocity components when I sum over their axis.
Screenshot from 2021-05-05 12-35-31

steven-murray:
Well. That does look like a bug :-/. 

Can you post the plots with `NO_RNG=True`?

BradGreig:
Hi @IvanNikolic21, I don't think I follow the plots you are presenting along the bottom row.

If I understand what you have written, you are presenting a 2D projection of the velocity field summed along the z-direction. What I don't understand is why this quantity is zero. The construction of the velocity field should be producing coherent large scale motions. Producing PDFs of the velocity return non-zero centred PDFs. Thus, I don't understand how you can be returning a uniform zero across a 2D image when summed along the z-direction. Could you perhaps explain in more detail what you are presenting along the bottom row to see if it can be replicated and thus investigated.

I have looked at both the master branch and the all-vel-components branches and I am not seeing any obvious indicators of bugs or anything to that fact. Everything appears to be functioning as it should be. I have only done this locally so far, so haven't produced as large a box as you have. But I haven't seen anything to indicate an issue.

qyx268:
I've only been checking the mater branch. @IvanNikolic21 found that the bug seems to be only in the all-vel-components branch, where 'k_direc' in line 536 is not set to private. This is the result he got after fixing it:
image

@BradGreig the mean/sum is not exactly zero. the colorbar scale was just too large to see the difference in the previous plot.

qyx268:
i'll just merge this PR now.

andreimesinger:
the z velocity component projected along z must sum to zero when summed over the whole box because of periodic boundary conditions.

PR #237, opened 2021-06-13
maint: update production branch to production from master

steven-murray:
This updates so that test deployments occur everywhere except merges to _production_, rather than master (since master is like our dev branch).


qyx268:
do we want to get #224 done before this?

steven-murray:
@qyx268 this is just some extra work that'll go into v3.3.2 or whatever. It can wait, although it's entirely separate from #224 so it might as well just go in when it's ready.

PR #247, opened 2021-07-14
Release v3.1.2

steven-murray:
### Internals

* ``MINIMIZE_MEMORY`` flag significantly reduces memory without affecting performance much,
  by changing the way some arrays are allocated and accessed in C (#224)


steven-murray:
@BradGreig this should essentially just be your PR getting released. Main things to check are that I got the CHANGELOG description right, and that you agree with bumping to 3.1.2 instead of 3.2.0

BradGreig (approved):
This all looks good to me.

Can we perhaps update the default behaviour of USE_INTERPOLATION_TABLES to be True in this same release? I don't see why this would be a problem with anything. Also it'll avoid more people running into these problems.

steven-murray:
Yeah good idea. I'll try do that tomorrow.

andreimesinger:
i agree with this

PR #249, opened 2021-07-15
fix: update use_interpolation_tables default with warning.

### LINKED ISSUE #186: update the default parameters for UserParams and FlagOptions

This issue aims to update the default parameters. Please post any other required updates below.

I think we have decided to set `PERTURB_ON_HIGH_RES=1` by default. 

Also, I strongly suggest to automatically set `USE_INTERPOLATION_TABLES=1` when `USE_MASS_DEPENDENT_ZETA==1`. I've been trapped by this many times. We can return a warning saying that 
```
You have set USE_MASS_DEPENDENT_ZETA to True but USE_INTERPOLATION_TABLES to False! "
                "Automatically setting USE_INTERPOLATION_TABLES to True."
```
If we really want to support running with `USE_MASS_DEPENDENT_ZETA=1` and `USE_INTERPOLATION_TABLES=0`, we can probably just give a warning without changing `USE_INTERPOLATION_TABLES`... What do people think?

**Additional context**

steven-murray:
Closes #186

I went the route of issuing a warning whenever the user doesn't set this explicitly (because in principle this is API breaking, so we should at least warn users that it's happening). This warning can be removed in the next major version.

### DIFF
--- a/.github/workflows/test_suite.yaml
+++ b/.github/workflows/test_suite.yaml
@@ -97,8 +97,7 @@ jobs:
           export PATH="$HOME/miniconda/bin:$PATH"
           source activate $ENV_NAME
           python -m pytest -l --cov=py21cmfast --cov-config=.coveragerc -vv --cov-report xml:./coverage.xml --durations=25 ${{ env.extra_pytest_args }}
-      - uses: codecov/codecov-action@master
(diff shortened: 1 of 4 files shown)


BradGreig (approved):
This makes sense to me

PR #251, opened 2021-07-17
Update rmax

qyx268:
add the EVOLVING_R_BUBBLE_MAX flag to use the redshift-dependent R_BUBBLE_MAX (37pMpc*0.7/hlittle*((1+z)/5)**-4.7 from Worseck+14 with adjusted index for Becker+21; cap at z>6, i.e. ~40cMpc).


qyx268:
@andreimesinger changing ALPHA_UVB (and possibly T_RE) leads to changes in the output duing test. This is why some of the tests failed in Github Action. Do you want to update these by default, meaning I will need to update the test files too, or keep them to their orignal values and update them when running 21cmfast.

PR #252, opened 2021-07-26
Adding ETHOS dark matter power spectrum to the code, plus smooth-k window function for making HMF

charlottenosam:
This PR adds the ETHOS DM power spectrum as a function of new AstroParams `h_PEAK` and `log10_k_PEAK` (see [Bohr+2020](https://ui.adsabs.harvard.edu/abs/2020MNRAS.498.3403B/abstract), Mason+in prep). 

This parameterized power spectrum spans WDM (h_PEAK=0) to CDM (log10_k_PEAK --> inf), and everything in between has dark acoustic oscillations. log10_k_PEAK [1-10] is the scale at which the matter power spectrum has its first oscillation (or cut-off for WDM) and h_PEAK [0-1] is the relative height of that oscillation. This should enable some easy explorations of a range of DM models.

Major code changes:
- Add new AstroParams `h_PEAK` and `log10_k_PEAK`. These are really cosmology parameters but they won't effect ICs on large scales -- they are really only changing the HMF, so it's fine to treat them as AstroParams
- Add `USE_ETHOS flag` to FlagOptions
- Move HMF window function `FILTER` from GlobalParams to FlagOptions so that it can be easily changed after ICs are calculated
- Add smooth-k window function for HMFs (see [Leo+2018](https://ui.adsabs.harvard.edu/abs/2018JCAP...04..010L/abstract)), which works better than tophat + sharp-k for truncated matter power spectra. The default parameters were calibrated to fit CDM, WDM and ETHOS DM models well at z=5-12.
- Some broadcasting of these new variables to functions that need them when calculating HMFs etc.

I guess we don't include any example notebooks though - would you guys like a notebook before we do the PR?


src/py21cmfast/inputs.py
```diff
@@ -727,6 +738,22 @@ def PHOTON_CONS(self):
...
 
+    @property
+    def FIX_VCB_AVG(self):
+        """FIX_VCB_AVG docstring TODO."""
+        return self._FIX_VCB_AVG
+
+    @property
+    def FILTER(self):
+        """The filter to use for the HMF calculations."""
+        if self.USE_ETHOS:
```
steven-murray, line 749: Please amend to `if self.USE_ETHOR and self._FILTER != 2:`

src/py21cmfast/inputs.py
```diff
@@ -821,6 +848,10 @@ class AstroParams(StructWithDefaults):
         Impact of the LW feedback on Mturn for minihaloes. Default is 22.8685 and 0.47 following Machacek+01, respectively. Latest simulations suggest 2.0 and 0.6. See Eq. XX.
     A_VCB, BETA_VCB: float, optional
         Impact of the DM-baryon relative velocities on Mturn for minihaloes. Default is 1.0 and 1.8, and agrees between different sims. See Eq. XX.
+    h_PEAK, log10_k_PEAK: double, optional
+        ETHOS parameters for dark acoustic oscillations (DAOs) in the matter power spectrum.
+        Location (10^log10_k_PEAK h/Mpc-1) and height (h_PEAK from 0 to 1) of the first DAO peak.
+        Warm dark matter (WDM) corresponds to h_PEAK=0, and log10_k_PEAK changes its mass. See Ref.~YY. (TODO!)
```
steven-murray, line 854: Just noting that there's a TODO here. Is there a plan to publish first then merge?

src/py21cmfast/src/UsefulFunctions.c
```diff
@@ -151,27 +151,30 @@ double HI_ion_crosssec(double nu);
 
 
 /* R in Mpc, M in Msun */
-double MtoR(double M){
+double MtoR(double M, struct FlagOptions *flag_options){
```
steven-murray, line 154: I think it would be cleaner to have `MtoR` depend on the int variable `filter` rather than all of `flag_options`.

src/py21cmfast/src/ps.c
```diff
@@ -740,17 +786,32 @@ double dsigmasq_dm(double k, void *params){
...
-        LOG_ERROR("No such filter: %i. Output is bogus.", global_params.FILTER);
+        LOG_ERROR("No such filter: %i. Output is bogus.", flag_options_sfrd->FILTER);
         Throw(ValueError);
     }
 
-//    return k*k*p*2*w*dwdr*drdm * d2fact;
+    if (flag_options_sfrd->USE_ETHOS == 1){
+      //add the suppression and DAOs due to ETHOS DM-DR interactions
+      // TODO: add debugging
+        // LOG_DEBUG("T=%e",ETHOS_DAOs(k, astro_params_sfrd));
```
steven-murray, line 811: Noting another TODO here. remove if unnecessary.

src/py21cmfast/src/ps.c
```diff
@@ -4292,3 +4370,51 @@ void FreeTsInterpolationTables(struct FlagOptions *flag_options) {
     LOG_DEBUG("Done Freeing interpolation table memory.");
 	interpolation_tables_allocated = false;
 }
+
+double ETHOS_DAOs(double k, struct AstroParams *astro_params_sfrd){
```
steven-murray, line 4374: again, better to have it depend just on `h_PEAK` and `log10_k_PEAK`, then pass those in directly when calling (since it's just two variables) in my opinion.

steven-murray (commented):
Thanks so much @charlottenosam, this is really impressive. 

I've made a couple of small comments on the code -- nothing major. The only major things to do, I think, are:

- [ ] Update the changelog
- [ ] Add a section to a tutorial (or a whole tutorial if you wish)
- [ ] Add tests (perhaps just an extra test in the integration test suite)

src/py21cmfast/src/21cmFAST.h
```diff
@@ -190,6 +195,7 @@ int CreateFFTWWisdoms(struct UserParams *user_params, struct CosmoParams *cosmo_
 void Broadcast_struct_global_PS(struct UserParams *user_params, struct CosmoParams *cosmo_params);
 void Broadcast_struct_global_UF(struct UserParams *user_params, struct CosmoParams *cosmo_params);
 void Broadcast_struct_global_HF(struct UserParams *user_params, struct CosmoParams *cosmo_params, struct AstroParams *astro_params, struct FlagOptions *flag_options);
+void Broadcast_struct_global_SFRD(struct UserParams *user_params, struct CosmoParams *cosmo_params, struct AstroParams *astro_params, struct FlagOptions *flag_options);
```
BradGreig, line 198: I think you should remove this function, and follow @steven-murray's suggestion of just passing the `FILTER`, `h_peak` and `log10_k_PEAK` parameters explicitly to the functions expecting them.

The idea of these functions are to broadcast everything to all functions in the respective PS (ps.c), UF (UsefulFunctions.c) and HF (heating_helper_progs.c) files. Thus this is essentially a duplication of `Broadcast_struct_global_PS`. As you only need to pass a couple of specific arguments once or twice, explicitly adding those as function arguments would be preferable.

BradGreig (changes requested):
Hey @charlottenosam, this all looks pretty good to me.

My main comment basically extends the suggestion by @steven-murray. In passing through the couple extra parameters as function arguments, then you shouldn't need the `Broadcast_struct_global_SFRD` function.

charlottenosam:
I am finally working on this again (slowly..!), so have converted to draft and will make the changes you have suggested :)

steven-murray:
@charlottenosam bumping this :-)

steven-murray:
@charlottenosam do you mean to revive this?

charlottenosam:
Oops, sorry guys. I don't think I'm going to have time to finish this, so happy for you to close this unless someone else really wants to finishing merging it! 

We ended up finishing this project with [Zeus21](https://github.com/JulianBMunoz/Zeus21) - ETHOS fork is here for future reference https://github.com/joverwohlt/Zeus21_ethos (@JulianBMunoz we can ask Jo to try to submit a pull request for Zeus).

steven-murray:
@charlottenosam in the absence of someone to take the lead in completing this, I'm recommending we close the PR for now, and let it be re-opened at a later stage if someone is interested.

I will note that much of the code will need to be refactored with v4 having come out recently, so it will not be a trivial exercise.

PR #256, opened 2021-10-05
Fixed ACG flag when using FAST_FCOLL_TABLES

JulianBMunoz:
It was only written for GaussLegendreQuad, but not for Nion_ConditionalM, and the latter was calling the former.

### DIFF
--- a/src/py21cmfast/src/21cmFAST.h
+++ b/src/py21cmfast/src/21cmFAST.h
@@ -29,7 +29,7 @@ struct UserParams{
     bool PERTURB_ON_HIGH_RES;
     bool NO_RNG;
     bool USE_INTERPOLATION_TABLES;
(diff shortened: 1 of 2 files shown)


steven-murray:
Hey @JulianBMunoz, I'm not sure I'm understanding what was wrong here?

steven-murray:
Ah, I see. That looks reasonable then. It is just the relvel integration tests that are failing (for coeval and lightcone). Does that help narrow it down?

BradGreig:
Hey @JulianBMunoz, it looks like you might just need to update the test data for the specific model that @steven-murray mentioned (relvel).

Though, looking at the integration test data, the "fast_fcoll" data may also need to be updated (given the data will have changed for similar reasons). Likely it was lucky that that didn't fail too.

PR #267, opened 2022-01-19
Fix FFT normalisation in FindHaloes

daviesje:
In the old version, the gaussian random field in k-space was passed to find_halos, which was normalised by VOLUME/TOT_NUM_PIXELS. Now the real field is passed to the function so the normalisation should be different after the iFFT.


BradGreig:
Hey @daviesje indeed you are correct. It seems that when I was using my own test suite to implement this I just so happened to end up with the unfortunate case of having set `BOX_LEN` and `DIM` to the same number! Lesson learnt...

In order to get this to pass you'll need to update all the test data for any test containing `USE_HALO_FIELD = True`. You can see which tests these are via all the failed tests above. Once the data is updated and the tests pass this is good to be merged.

steven-murray:
To create new test data, move to the `tests/` directory and run 

```
python produce_integration_test_data.py --names NAME --names ANOTHER_NAME ...
```

The names of the tests that you need to run, as @BradGreig said, are in the failing tests. You can use `--help` for more help.

daviesje:
Thanks, I've regenerated the files used in the halo tests.

BradGreig (approved):

BradGreig:
Hey @steven-murray the docs are failing along with some of the plotting routines. Are you able to look into these, or bypass them?

PR #273, opened 2022-02-21
Release v3.1.4

steven-murray:
This is what was meant to happen in #271 -- I merged to master instead of production :/


.github/workflows/test_suite.yaml
```diff
@@ -20,86 +20,85 @@ jobs:
...
+        python-version: [3.7, 3.8, 3.9, "3.10"]
+    defaults:
+     run:
+       # Adding -l {0} ensures conda can be found properly in each step
+       shell: bash -l {0}
     steps:
       - uses: actions/checkout@master
         with:
           fetch-depth: 0
-      - name: Get Miniconda MacOSX
```
qyx268, line 35: so nowadays, there is no need to distinguish macos and ubuntu?

qyx268 (approved):

steven-murray:
@qyx268 I think it's complaining about deploying to test pypi because this version already exists there. That makes sense because I already did this PR but to the wrong branch.

.github/workflows/test_suite.yaml
steven-murray: That's correct, it has its own action that will determine that.

PR #286, opened 2022-06-02
fix: R_BUBBLE_MAX set to BOX_LEN if it's too large

### LINKED ISSUE #112: [BUG] Box fails for R_BUBBLE_MAX > BOX_LEN
We should issue a warning when either `R_BUBBLE_MAX > BOX_LEN` or `HII_FILTER_RADIUS = 1` and `R_BUBBLE_MAX <= BOX_LEN/3`.

steven-murray:
Fixes #112 

In fixing this, I found that it was nice to have a standalone function where we put all the validation between different structs. This will make it easier in the future to do the validation between eg. user params and flag options.

I also found it useful to abstract out some of the logic in each of the high-level functions where we basically just configure/check inputs. Since each function had essentially the same pattern, it seemed easier just to make it a function (I called it `_setup_inputs`). This function now runs the new `validate_all_inputs` function.


src/py21cmfast/inputs.py
```diff
@@ -931,3 +918,47 @@ def t_STAR(self):
             raise ValueError("t_STAR must be above zero and less than or equal to one")
         else:
             return self._t_STAR
+
+
+class InputCrossValidationError(ValueError):
+    """Error when two parameters from different structs aren't consistent."""
```
BradGreig, line 924: Is this used anywhere? Or is it currently incomplete?

steven-murray: It's not used, but it is intended to be used in `validate_all_inputs` when something is wrong. At the moment, `validate_all_inputs` is just warning and setting things to appropriate values. I thought I'd leave this here for the future.

BradGreig (approved):

src/py21cmfast/inputs.py
```diff
@@ -931,3 +918,47 @@ def t_STAR(self):
...
+    if astro_params is not None:
+        if astro_params.R_BUBBLE_MAX > user_params.BOX_LEN:
+            astro_params.update(R_BUBBLE_MAX=user_params.BOX_LEN)
+            warnings.warn(
+                f"Setting R_BUBBLE_MAX to BOX_LEN (={user_params.BOX_LEN} as it doesn't make sense for it to be larger."
+            )
+
+        if (
+            global_params.HII_FILTER == 1
+            and astro_params.R_BUBBLE_MAX < user_params.BOX_LEN / 3
```
BradGreig, line 949: How does this behave with the inhomogeneous recombinations condition (that is, R_BUBBLE_MAX = 50 if set to true)? Does some weirdness occur here whereby R_BUBBLE_MAX != 50 if INHOMO_RECO = True?

Just thinking that in the case of INHOMO_RECO = True and BOX_LEN = 125 the result will be R_BUBBLE_MAX < 50. Not sure if this is desirable... Not sure if twice the size instead of three times the size will help mitigate this.

steven-murray: @BradGreig, these conditions don't reference `INHOMO_RECO` at all, so they'll execute regardless of its value. Nevertheless, if `R_BUBBLE_MAX = 50` and `BOX_LEN=125`, nothing should happen to `R_BUBBLE_MAX` here. It's only if `R_BUBBLE_MAX` was smaller than 42 that it would set it back up to ~42. 

I actually wasn't perfectly sure of the correct logic here (physically speaking). I was just following the bare bones of what I wrote in #112. What do you think the logic should be?

One more thing: auto-setting the `R_BUBBLE_MAX` like we do here is actually changing tests, so I'm wondering if it's a little to restrictive, or whether we should change our tests...

PR #287, opened 2022-06-02
feat: floats in hashes now written to given sig figs

### LINKED ISSUE #80: [BUG] Caching doesn't work unless float precision is exact
**Describe the bug:**

When reading from cache, if the original redshift was say 7.000001, and you ask for 7, it will recompute.

**To Reproduce:**
```
import logging
logger = logging.getLogger("21cmFAST")
logger.setLevel(logging.INFO)
p21c.run_coeval(redshift=7.00001)
p21cm.run_ceoval(redshift=7.0)
```
And notice that it doesn't say it's reading in redshift-dependent boxes.
This is annoying when the cubes are created by the lightcone scrolling at 
strange float redshifts.

**Expected behavior:**

I would expect that some tolerance in redshift should be acceptable. Maybe 10^-3?

steven-murray:
Fixes #80

This changes the string representation of `OutputStructs` and `StructWithDefaults` objects so that floats appear with a certain number of significant digits (configurable in global configuration). Primarily deals with the caching of redshifts (by default, it uses 4 significant digits, so the redshifts 6.001 and 6.000 would be considered different, but not 6.0001). 

A couple of things to think about:

1. I've applied the float rendering also to input structs, so eg. `HII_EFF_FACTOR=30.0` and `HII_EFF_FACTOR=30.001` would be considered the same. This could play a little havoc on MCMC routines (or other uses where a large database of randomly sampled parameters is chosen), so that the parameters change but the outputs don't, creating a "step-like" likelihood. To offset this, the user can set the significant digits much higher on a particular run. Also, they probably shouldn't be caching on such a run. But if we think that's not a good idea, I can turn off the significant figure truncation on the input params (but not redshift).
2. Unfortunately, this change could render a user's existing cache unusable. I've warned about this in the CHANGELOG, but they might not read that.
3. it strikes me that different parameters might need different precision. Redshifts probably don't need 4 digits, while astro/cosmo params might. I don't have a great way of enforcing this. I could hack it out though, if we think it's important.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,11 +9,13 @@ Changed
 
 * Floats are now represented to a specific number of significant digits in the hash of
   an output object. This fixes problems with very close redshifts not being read from
-  cache (#80). Note that this means that ver close astro/cosmo params will now be read
+  cache (#80). Note that this means that very close astro/cosmo params will now be read
   from cache. This could cause issues when creating large databases with many random
   parameters. The behaviour can modified in the configuration by setting the
-  ``cache_significant_figures`` parameter. **NOTE**: updating to this version will cause
-  your previous cached files to become unusable. Remove them before updating.
+  ``cache_param_sigfigs`` and ``cache_redshift_sigfigs`` parameters (these are 6 and
+  4 by default, respectively).
+  **NOTE**: updating to this version will cause your previous cached files to become
+  unusable. Remove them before updating.
 
 v3.1.5 [27 Apr 2022]
 ----------------------
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,17 @@ Changelog
 dev-version
 -----------
 
+Changed
+-------
+
+* Floats are now represented to a specific number of significant digits in the hash of
+  an output object. This fixes problems with very close redshifts not being read from
+  cache (#80). Note that this means that ver close astro/cosmo params will now be read
+  from cache. This could cause issues when creating large databases with many random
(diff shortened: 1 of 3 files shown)


BradGreig:
Hey @steven-murray, is it possible to do this as number of decimal places rather than significant figures? I feel like checking on the number of decimal places works better as it gives you better resolution on your parameters. Otherwise for some large parameters, you may only have one (or less) decimal place.

BradGreig:
Hey @steven-murray, yeah, I guess I was thinking of the counter example of having a parameter being say 1500 and 1499.5 and these being treated as being the same (depending on the rounding). While arguably they are likely to be pretty similar, you could probably come up with an example (e.g. certain parameter) where this could be a little problematic.

Perhaps a happy medium would be to use six significant figures over four? That way the increased precision helps for databases/MCMC runs. Though, I guess you wouldn't be saving anything in an MCMC where this could cause issues.

steven-murray:
@BradGreig yeah, that's a good point. Really, the precision required will depend on each parameter and the constraining power of whatever observation is being fit (though, as you say, in MCMC mode, one shouldn't be caching on astro params anyway). 

I've made a small update that might be a good medium: now we allow separate configuration options for parameters versus redshifts. By default they get 6 and 4 sig figs respectively. Let me know if you think they are good values, and whether it needs to be more fine-tuned than that.

BradGreig:
Hey @steven-murray, yeah, I think that is probably a reasonable compromise.

BradGreig (approved):

qyx268:
just want to shout that this is a great feature!

PR #288, opened 2022-06-02
feat: ability to save/read directly from an open file

### LINKED ISSUE #170: [Feature Req.] More flexible data saving/loading
At the moment, whenever `.save` or `.read` functions are called, the only option is to save/read directly from the filename. This is good as it is self-consistent, but is not very flexible. For example I would like to save a lightcone object, but in the same hdf5 would like to have additional data. 

SImple solution would be to enable the user to pass opened hdf5 `file` or `group`. In that way structure would still be consistent and user would have the ability to save and load objects as a part of their larger hdf5 files.

steven-murray:
Fixes #170 

Adds ability to use an open hdf5 file/group on `.save` and `.read`.


BradGreig (approved):

qyx268 (approved):

src/py21cmfast/_utils.py
```diff
@@ -939,7 +939,13 @@ def exists(self, direc=None):
         """
         return self.find_existing(direc) is not None
 
-    def write(self, direc=None, fname=None, write_inputs=True, mode="w"):
+    def write(
+        self,
+        direc=None,
+        fname=str | Path | None | h5py.File | h5py.Group,
```
dprelogo, line 964: This is problematic for python < 3.10 and some tests seem to be failing because of it.
It should be something like: `fname : Union[str, Path, None, h5py.File, h5py.Group] = None`

dprelogo (approved):

PR #295, opened 2022-09-08
update luminosity function to work with minihalos

qyx268:


steven-murray (commented):
Thanks @qyx268, this all looks good to me. Please just update the CHANGELOG, then I'll approve.

CHANGELOG.rst
```diff
@@ -4,6 +4,15 @@ Changelog
 dev-version
 -----------
 
+v3.2.1 [13 Sep 2022]
+----------------------
+
+Changed
+~~~~~
```
steven-murray, line 11: this needs to be the full length of the title. Also, you could update the deploy workflow to use the latest version of the uploader action, rather than `master` (see the error for the deploy script)

steven-murray (approved):

PR #297, opened 2022-09-25
Radio excess from ACG, MCG and PBH

JSC42:
Added excess radio background from: 1. Atomically cooled galaxies (ACG), 2. Molecularly cooled galaxies (MCG or minihalo), 3. Accreting PBH. Feature for Hawking radiation also included.


src/py21cmfast/src/BrightnessTemperatureBox.c
```diff
@@ -1,485 +1,561 @@
...
-                for (k=0; k<user_params->HII_DIM; k++){
-                    *((float *)v + HII_R_FFT_INDEX(i,j,k)) = perturb_field->velocity[HII_R_INDEX(i,j,k)];
+            for (i = 0; i < user_params->HII_DIM; i++)
+            {
+                for (j = 0; j < user_params->HII_DIM; j++)
+                {
+                    for (k = 0; k < user_params->HII_DIM; k++)
+                    {
+                        *((float *)v + HII_R_FFT_INDEX(i, j, k)) = perturb_field->velocity[HII_R_INDEX(i, j, k)];
+                    }
```
steven-murray, line 49: Not sure I like this style here.

src/py21cmfast/src/JSC.h
```diff
@@ -0,0 +1,571 @@
...
+#include "Hawking_Radiation.h"
+
+// ---- Knobs and Switches ----
+// nu0 is degenerate with fR so no reason to leave this as a param
+#define astro_nu0 0.15 // in GHz
+
+// Debuging options, print out gas temp and SFRD box, I am gonna keep these
+#define print_SFRD_box 1
+#define Reset_Radio_Temp_HMG 0
+#define debug_mode 1
```
steven-murray, line 14: Might not need this since we have LOG_DEBUG etc.

src/py21cmfast/src/SpinTemperatureBox.c
```diff
@@ -1,14 +1,15 @@
-
+#include "JSC.h"
 // Re-write of find_HII_bubbles.c for being accessible within the MCMC
+// Note for JSC: Anything that begins with Radio or Hawking are my modules
```
steven-murray, line 3: maybe calling the module "JSC" isn't the best idea for the long-term. Could we call it something more meaningful?

steven-murray (changes requested):
Thank you @JSC42, this is a very cool set of features! I am far from understanding the code/physics, but I do have a few suggestions (other than the few comments I left). The two big things that I think need to be done in this PR are:

1. The new stuff should all be made optional via one or more `FlagOptions` which are by default set to False. 
2. You should add one or more integration tests to the suite with these options turned on.

Something else to think about, maybe for @BradGreig / @qyx268: since there are *so many* parameters for this feature, would it be reasonable to admit another separate parameter struct? Eg. `PBHParams`? It would be annoying to implement this currently, because so many of the functions take in the four parameter structs, and this would have to be added everywhere. But this could also force us to be a bit more systematic in how we manage this, and open up the possibility of adding extensions more easily in the future.

Another thought: would it be possible to make the main part of this new calculation a new C function entirely, that takes in an `IonizedBox` and/or `TsBox` and computes the `TradBox` (optionally)? That would keep things a little more modular.

qyx268:
@steven-murray  yeah I agree. I'm also reviewing another work (although I'm not sure when if ever we will merge it into the master), which has introduced 11 new astro parameters and another 11 new flags... The developers were in fact including all astro parameters in UserParams, which py21cmmc cannot cope with. I agree a long term solution will be allowing new classes to be added for new and distinct features such as radio source / dark matter...

For short term, they have to be put in astro_params if they are not affecting InitialConditions...

src/py21cmfast/src/SpinTemperatureBox.c
JSC42: Hi Steven, the JSC signatures in previous PR has now been removed

JSC42: the JSC signatures in previous PR has now been removed

src/py21cmfast/src/BrightnessTemperatureBox.c
JSC42: Hi, I have added new FlagOptions for all the new features (USE_Radio_ACG, USE_Radio_MCG, USE_Radio_PBH, USE_Hawking_Radiation)

JSC42: FlagOptions are added for all the new features (USE_Radio_ACG, USE_Radio_MCG, USE_Radio_PBH, USE_Hawking_Radiation), defaults are set to False

src/py21cmfast/src/SpinTemperatureBox.c
JSC42: and the JSC.h has also been renamed as RadioExcess.h

src/py21cmfast/inputs.py
```diff
@@ -626,6 +626,19 @@ class FlagOptions(StructWithDefaults):
         Determines whether to use a fixed vcb=VAVG (*regardless* of USE_RELATIVE_VELOCITIES). It includes the average effect of velocities but not its fluctuations. See Muñoz+21 (2110.13919).
     USE_VELS_AUX: bool, optional
         Auxiliary variable (not input) to check if minihaloes are being used without relative velocities and complain
+    USE_Radio_ACG: bool, optional
+        Determines whether to use radio excess background from ACG, set this to True and AstroParams.fR > 0 to activate
+    USE_Radio_MCG: bool, optional
+        Determines whether to use radio excess background from MCG, set this to True and AstroParams.fR_mini > 0 to activate
+    USE_Radio_PBH: bool, optional
+        Determines whether to use radio excess background from PBH, set this to True and AstroParams.log10_fbh > -200 to activate
+    USE_Hawking_Radiation: bool, optional, see 2108.13256
```
steven-murray, line 635: For visual consistency with other parameters, can we make these all caps? I actually don't like the all-caps myself, but consistency is better than opinion ;-)

src/py21cmfast/src/SpinTemperatureBox.c
```diff
@@ -236,6 +235,33 @@ int ComputeTsBox(float redshift, float prev_redshift, struct UserParams *user_pa
...
+            LOG_ERROR("Mission aborted due to conflicting params: numerical params says Use_Radio_PBH but FlagOptions->USE_Radio_PBH=F, you need to set FlagOptions->USE_Radio_PBH=T to use Radio PBH.");
+            Throw(ValueError);
+        }
+        if ((!flag_options->USE_Hawking_Radiation) && (Use_Hawking_Radiation))
+        {
+            Use_Hawking_Radiation=false;
+            LOG_ERROR("Mission aborted due to conflicting params: numerical params says Use_Hawking_Radiation but FlagOptions->Use_Hawking_Radiation=F, you need to set FlagOptions->Use_Hawking_Radiation=T to use Hawking Radiation.");
+            Throw(ValueError);
+        }
+
```
steven-murray, line 264: We actually deal with this kind of logic on the Python side, where it is a little bit easier to do so. Use the `py21cmfast.inputs.validate_all_inputs()` function to do these checks.

Also, you should not have both `flag_options->USE_Hawking_Radiation` AND `Use_Hawking_Radiation` defined. Just have one source of truth (the flag options).

steven-murray (commented):
Thanks @JSC42, I think this is much closer -- just a few more small comments.

src/py21cmfast/inputs.py
JSC42: Thanks Steven, all the new FlagOptions are now in caps

src/py21cmfast/src/BrightnessTemperatureBox.c
JSC42: yeah I was using VS codes so the styling of BrightnessTemp.c is set by VS auto-indentation, anyway I will revert the code indentation style back to the original version

JSC42: I have reverted BrightnessTemperatureBox.c to the original indentation style

src/py21cmfast/src/SpinTemperatureBox.c
JSC42: Yeah in the earlier versions the Use_Hawking_Radiation (and other bool variables) were used to determine whether certain new features should be turned on, but with the new FlagOptions these old variables seem unnecessary. In the new version I have removed these internal switches in SP.c.

src/py21cmfast/src/JSC.h
JSC42: It might actually be more convenient to keep the print_SFRD_box feature, this is used to print out huge arrays to a file. Using other original debug features (e.g LOG_DEBUG) will also turn on a bunch of other debug infos that are irrelevant for the new features

src/py21cmfast/src/SpinTemperatureBox.c
JSC42: Inputs parameter checks are now moved from SP.c to inputs.py

steven-murray (approved):
Thanks @JSC42 this all looks good to me now. Let's just wait for someone to approve the physics, eg @qyx268.

andreimesinger:
I can take a look at the end of the week likely; and/or Zip here could also take a look (having done the PBH derivations)

steven-murray:
Great -- it looks like master has to be merged back into this branch, and then we can make sure tests pass.

src/py21cmfast/inputs.py
```diff
@@ -813,20 +813,17 @@ class AstroParams(StructWithDefaults):
...
         Power-law energy spectra index for mini-halos
-    log10_mbh: float, optional
-        log10 of PBH birth mass in msun
-    log10_fbh: float, optional
-        log10 of PBH fraction, i.e. rho_PBH/rho_dm
-        Allowed range: [-Inf, 0]. Set to below -100 to turn off PBH (both radio and Hawking), log10_fbh > 0.01 triggers error
+    mbh: float, optional
+        PBH birth mass in msun. Given in log10 units.
+    fbh: float, optional
+        PBH fraction, i.e. rho_PBH/rho_dm, allowed range: [0, 1]. Given in log10 units.
```
steven-murray, line 826: I don't mind it being given in log10 units, but I think it's much cleaner if the name includes the log10, so I prefer the original

PR #303, opened 2022-10-10
Improved implementation for perturbed density/velocity field

BradGreig:
This is a PR to fix an issue with the statistics of the perturbed density field as pointed out by @rajeshmondal18.

Previously, when calculating the perturbed density, all the mass of a perturbed cell was placed into a single cell. This features replaces this step with an implementation of cloud in cell to re-distribute the mass over the 8 neighbouring cells based on their distance from the centre of the perturbed cell.

This removes spiky features from the density field PDF and corresponding oscillatory features from the velocity PDF. The linear/log PDF of the density and velocity fields highlighting the issue are shown [here](https://github.com/21cmfast/21cmFAST/files/9742933/DensVelPDFs_z15_384_128_200Mpc_prefix.pdf) while following this fix the correspond PDFs [remove](https://github.com/21cmfast/21cmFAST/files/9742935/DensVelPDFs_z15_384_128_200Mpc_postfix.pdf) the sharp/oscillatory features.

Comparing the light-cones appears to show very little difference [(see here)](https://github.com/21cmfast/21cmFAST/files/9742936/lightcone_comparison.pdf). Reionisation histories and global signals remain unaffected. Looking at the 21-cm PS across the light-cone the largest [differences](https://github.com/21cmfast/21cmFAST/files/9742938/21cmPS_Comparison.pdf) occur at increasing z on the smallest scales. This difference is due to the fact that the previous implementation over estimated the small-scale power at increasing z. However, in all inference work, the thermal noise completely dominates at these scales and thus previous results should remain unaffected.

All test-data will need to be updated as a result of this fix.

I suspect this issue arose due to the `PERTURB_ON_HIGH_RES` option. The previous approach to just move all the mass into the new cell would work when perturbing directory onto the `HII_DIM` box. But since we now have the option to perturb directly onto the `DIM` box, the resolution difference would have amplified the effect.


steven-murray (approved):
Thanks @BradGreig! The changes all look good to me. We'll just have to get the tests working.

steven-murray:
Oh, I forgot to say that you need to update the CHANGELOG

steven-murray:
Tests are failing due to a compilation error from a missing `{`, so that should be easy to fix.

src/py21cmfast/src/PerturbField.c
```diff
@@ -257,13 +260,50 @@ int ComputePerturbField(
...
+                        t_z = 1. - d_z;
+
+                        // Determine the grid coordinates of the 8 neighbouring cells
+                        xp1 = xi + 1;
+                        if(xp1 >= dimension) { xp1 -= (dimension);}
+                        yp1 = yi + 1;
+                        if(yp1 >= dimension) { yp1 -= (dimension);}
+                        zp1 = zi + 1;
+                        if(zp1 >= dimension) { zp1 -= (dimension);}
+
```
andreimesinger, line 279: I think these should be defined wrt the midpoint of the cell, e.g. d_x = fabs(xf - (float)(xi+0.5)), and  xp1  being either xi+1 or xi-1, depending on if xf is gtr or lessthen xi+0.5.

andreimesinger (changes requested):
note the midpoint of the cell definition.

also, this isn't a "big-fix".  The previous version was implementing a "nearest cell" window function, correctly.  This changes "nearest cell" to "cloud in cell", which is less noisy in the limit of finite resolution.

src/py21cmfast/src/PerturbField.c
BradGreig: I have implemented these changes

BradGreig:
I've made some updates to the code. I noted that the original usage of floats could cause issues on the periodic boundary conditions when re-distributing the mass over neighbouring cells. Upon replacement of this with doubles, the spike in the density PDF at overdensities of -1 has gone away as shown here [here](https://github.com/21cmfast/21cmFAST/files/9751469/DensVelPDFs.pdf). Everything else has remained relatively unchanged.

Updating this double/float issue and correctly applying omp atomic statements has resolved almost all issues with the tests. One remaining test fail, which I haven't yet properly looked into. Though not immediately obvious why it is failing...

andreimesinger (approved):
looks good to me

andreimesinger:
Can we pls change the default min redshift for photon conservation to z=4 instead of z=5?  Might be easiest to do here since this PR is about to come in...

BradGreig:
Hey @steven-murray @andreimesinger, in order to get the final test to pass I had to reduce the relative tolerance for that test data (see my comments above).

This PR now also includes a modification following #305. Changelog has been updated to reflect these changes.

steven-murray:
Thanks @BradGreig, and apologies for the slow response on this (I'm out at the Global 21cm Workshop this week). I'm a little uncomfortable with the tolerance change as well. I mean, the different is still very small, and it probably not a problem, but it bugs me that there's any difference at all. There shouldn't be essentially any randomness, right? 

I think I'm happy for us to merge this, because it's not an urgent/significant problem, but I'd feel happier if we made an issue to remind ourselves to look into it a bit further.

andreimesinger:
What is the hold-up for merging this?

Issue #311, opened 2022-11-08
Segmentation Fault when using photon conservation


IvanNikolic21:
**Describe the bug:**
21cmFAST produces Segmentation Fault errors for certain values of parameters. This happens while running multinest within 21CMMC and only when using photon-conservation flag 'PHOTO_CONS=True'. Unfortunately, segmentation fault crashes the whole run when multinest is ran with mpirun.

**To Reproduce:**
Using the following code  with _mpirun --map-by ppr:nproc:node run.py_:

```
from astropy import cosmology
from py21cmmc import mcmc
from py21cmmc import LikelihoodNeutralFraction
from py21cmmc import CoreLightConeModule
from py21cmmc import LikelihoodLuminosityFunction, CoreLuminosityFunction
from py21cmmc import LikelihoodForest, CoreForest
from py21cmmc import LikelihoodPlanck
from py21cmmc import LikelihoodBase

user_params = {'HII_DIM':128, 'BOX_LEN':250.0, 'USE_INTERPOLATION_TABLES': True, 'USE_FFTW_WISDOM': True, "PERTURB_ON_HIGH_RES": True, "N_THREADS": 4, "OUTPUT_ALL_VEL": True}
cosmo_params = {'SIGMA_8': 0.8118, 'hlittle': 0.6688, 'OMm': 0.321, 'OMb':0.04952, 'POWER_INDEX':0.9626}
flag_options = {'USE_MASS_DEPENDENT_ZETA': True, "INHOMO_RECO": True, "PHOTON_CONS": True,}
global_params = {'Z_HEAT_MAX': 15.0, 'T_RE': 1e4, 'ALPHA_UVB': 2.0, 'PhotonConsEndCalibz':3.5}

model_name = "forest"

cosmo = cosmology.FlatLambdaCDM(H0=cosmo_params['hlittle']*100, Om0=cosmo_params['OMm'],Tcmb0=2.725,Ob0=cosmo_params['OMb'])

class Prior(LikelihoodBase):
    def reduce_data(self, ctx):
        params = ctx.getParams()
        return  [params.log10_f_rescale, params.f_rescale_slope]

    def computeLikelihood(self, model):
        return -0.5 * ((model[0] / 5)**2 + (model[1] / 2.5)**2)

[...]

Issue #296, opened 2022-09-23
Interpolation Table initialization is brittle


steven-murray:
Currently we have a bunch of interpolation tables that need to be initialized at various places before they are interpolated. These are typically global variables and are accessed from all over the place in different C files / functions. Therefore, the initialization code has to be called numerous times in different places.

This gets more tricky because the user could sometimes read in a file (eg. the spin temperature) and compute the ionization box from that spin temperature, which means all the initialization code in ComputeTsBox is not called. Therefore it has to be called again in IonizationBox, but we don't want to call it twice if we don't have to. 

Currently we're trying to do this by having a bunch of logical conditions (eg. `if (!flag_options.USE_TS_FLUCT`)...`) but these are bound to fail in corner cases.

I suggest we use a more foolproof system: let each table be a small struct that has a reference to its size, minimum and maximum values, the data itself, and whatever parameters are required to compute it. Whenever the initialize method for that struct is called, if all of its metadata is the same as the global struct, then just exit, otherwise do the initialization. That way we never have to worry about the current state in the actual code.

PR #312, opened 2022-11-11
Fix for the specific case in #311

BradGreig:
Photon conservation correction could miss the initialisation step under an extreme scenario of a very early EoR and with low `Z_HEAT_MAX`. This PR adds in a check for this occurrence and initialises the requisite function if it has not already been initialised.

A fix for the specific example in #311. But not necessarily the broader issue of dealing with the brittle photon conservation correction

### DIFF
--- a/src/py21cmfast/src/ps.c
+++ b/src/py21cmfast/src/ps.c
@@ -3984,6 +3984,14 @@ float adjust_redshifts_for_photoncons(
             adjusted_redshift = *redshift;
         }
         else {
+            // Initialise the photon non-conservation correction curve
+            // It is possible that for certain parameter choices that we can get here without initialisation happening.
+            // Thus check and initialise if not already done so
+            if(!photon_cons_allocated) {
+                determine_deltaz_for_photoncons();
+                photon_cons_allocated = true;
+            }
+
             // We have crossed the NF threshold for the photon conservation correction so now set to the delta z at the threshold
             if(required_NF < global_params.PhotonConsAsymptoteTo) {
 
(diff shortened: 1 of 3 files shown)


src/py21cmfast/src/ps.c
```diff
@@ -3984,6 +3984,14 @@ float adjust_redshifts_for_photoncons(
...
         }
         else {
+            // Initialise the photon non-conservation correction curve
+            // It is possible that for certain parameter choices that we can get here without initialisation happening.
+            // Thus check and initialise if not already done so
+            if(!photon_cons_allocated) {
+                determine_deltaz_for_photoncons();
+                photon_cons_allocated = true;
+            }
+
```
BradGreig, line 3994: Note, this is repetitive code with the other initialisation check further down. I thought about adding this earlier to only have this code once, but I'm not sure of the consequences of attempting to run this initialisation in other parts of this function (i.e. under certain conditions).

qyx268:
you have no idea how many lines of printf I wrote to discover these 2 lines of solution...

qyx268:
better to let the test finish. i'll approve afterwards...

qyx268:
Looks like jinja2==3.0.3 is needed?

steven-murray:
Thanks @BradGreig  -- I think this is also related to #296. This is a fine temporary solution in this case.

IvanNikolic21:
Hi, what's the progress with this pull request @steven-murray ? We're waiting for this to start running inference.

BradGreig:
Hi @IvanNikolic21, apologies, I hadn't realised this wasn't merged. I though it would have been merged straight away.

I've gone and fixed the dependency issue (temporarily pinning it to a specific version). So @qyx268 or @steven-murray you should be able to just approve it and merge.

qyx268 (approved):

PR #313, opened 2022-11-24
Angular lightcone

steven-murray:
This PR adds the ability to create lightcones in Angular coordinates (as well as a more flexible lightconing interface).

In total, it does the following:

* Makes it possible to checkpoint lightcones. 
* Adds a new `Lightconer` base class, and two implementations: a `RectilinearLightconer` (equivalent to the old way) and an `AngularLightconer`.
* Removes some excess warnings we were producing
* Makes a small fix to the `lightcone_coords` attribute of the `LightCone` class. 
* Adds an `AngularLightcone` class to deal with the outputs of these lightcones.


andreimesinger:
would it be possible to include some visualizations for this PR, showing the two options?

steven-murray:
@andreimesinger Ah, yes that was always my intention. I'll get to that ASAP.

JulianBMunoz:
I feel like adding a new section to the lightcone tutorial (or a new tutorial entirely) would also be great

steven-murray:
@JulianBMunoz that's a great section. In fact, I'm currently updating the lightcone tutorial with this (that was the notebook we were looking at last week with the weird aspect ratio plot).

steven-murray:
OK @andreimesinger and @JulianBMunoz, I'm gonna post some plots here, but also these plots are all available in the updated documentation [here](https://21cmfast--313.org.readthedocs.build/en/313/tutorials/lightcones.html).

First, I made an angular grid that was suppose to closely match the rectilinear grid (at least at the lowest redshift), and looking directly at the face of the lightcone, we get:

image

Then, looking at the redshift evolution (top is rectilinear, bottom is angular): 

image


Of course, the angular lightcone in the above is plotted as if it were rectilinear, so we can also plot it in polar coords:

image

And, just to prove that we're not limited to doing small areas of the sky corresponding to the rectilinear lightcone, I simulated a "lightplane" -- i.e. a single strip along the horizon, extending into the line of sight. Plotting that in polar coords: 

image

andreimesinger:
beautiful!  i love the lightplane!!On 13 Dec 2022, at 22:50, Steven Murray ***@***.***> wrote:﻿
OK @andreimesinger and @JulianBMunoz, I'm gonna post some plots here, but also these plots are all available in the updated documentation here.
First, I made an angular grid that was suppose to closely match the rectilinear grid (at least at the lowest redshift), and looking directly at the face of the lightcone, we get:

Then, looking at the redshift evolution (top is rectilinear, bottom is angular):

Of course, the angular lightcone in the above is plotted as if it were rectilinear, so we can also plot it in polar coords:

And, just to prove that we're not limited to doing small areas of the sky corresponding to the rectilinear lightcone, I simulated a "lightplane" -- i.e. a single strip along the horizon, extending into the line of sight. Plotting that in polar coords:

andreimesinger:
The "face" slices you posted look a bit weird..  The angular one looks smoothed and is missing some large structures compared to the rectilinear one....

steven-murray:
Yeah -- so I think it's all as expected. At the top left corner, the match is very very close, because the (0,0) pixel of the angular lightcone perfectly corresponds to the rectilinear. However, moving away from there, the angular lightcone bends away from the rectilinear, and so it is showing effectively a different slice at the bottom right. The smoothing I'm not quite sure -- probably an interpolation thing. Probably if I created the angular lightcone at double the resolution, it would look more crisp.

BradGreig:
This is unavoidable. An interpolated 21-cm signal will always be blurry as there is no way to interpolate between the sharp edges of the ionised regions. It inevitably always leads to blurry features around ionised regions. Increasing the resolution will not help this.

steven-murray:
In terms of power spectrum, though, this shouldn't be a big deal as it should only affect very small scales, right? I guess it's something we should test.

JulianBMunoz:
This is what I was going to ask, does the power spectrum change much
between the rectangular and angular coordinates?

steven-murray:
@JulianBMunoz I can take the (cartesian) power spectrum of the angular lightcone, which is slightly the wrong thing to do because the coordinates aren't exactly rectilinear. OR I can compute the angular power spectrum after tiling onto a full healpix map (this is a bit more compute heavy, obviously). For the latter, I can never remember exactly how to relate the angular PS back to the 3D PS in a simple way, so I'm less sure it'll be useful.

JulianBMunoz:
I think in order to compare more fairly it may make sense to compare angular power spectra, so C_\ells. In the case of the rectilinear lightcone one can take the flat-sky approximation or similar to compute the C_ells. Ping me on Slack if you want to chat more

steven-murray:
OK @JulianBMunoz @BradGreig @andreimesinger , I think this is now ready for final review. I have added RSDs in properly to the angular lightcone, based on this PR: https://github.com/steven-murray/cosmotile/pull/106 which will need to be merged before this one passes tests. 

Again, the lightcone tutorial shows how to use the new features, and has all the plots that show it's "working" properly.

steven-murray:
Note to self: I just merged master to this branch, and the non-cubic-factor PR seriously messes with it. At least two things will break: 

1. here I changed from using just `velocity` in perturbed field to keeping `velocity_x/y/z`. Since the NCF changed a bunch of calls to the `velocity` box (changing HII_DIM -> HII_MIDDLE_PARA), I'll have to go back and put those back in manually in some cases.
2. more importantly, the RSD calculation was updated in this PR (mostly moving around big chunks of code, not actually changing the calculation itself). Obviously parts of this were also changed in the NCF. In merging, I just kept the work from _this_ PR, so I'll have to go through and manually add back in the changes from the NCF PR (and, I guess, any other changes on master...)

steven-murray:
@andreimesinger + @JulianBMunoz this PR is now ready for final review/approval. It's kinda large and unwieldy, but all tests are passing and I've even added tests of the documentation tutorials. The validation of the new lightconing method is in the lightcone tutorial.

andreimesinger (approved):

PR #315, opened 2022-12-20
Added ubuntu installation instructions

DanielaBreitman:
Added installation instructions for ubuntu reflecting my recent struggles (thank you @steven-murray )
Updated the 21cmfast environment YAML file to install python 3.9 and removed the gcc library package inside (since we found it was better to use a system installation of gcc)
Added --verbose flags in build_cffi,py

### DIFF
--- a/INSTALLATION.rst
+++ b/INSTALLATION.rst
@@ -41,8 +41,8 @@ Most linux distros come with packages for the requirements, and also ``gcc`` by 
 which supports ``-fopenmp``. As long as these packages install into the standard location,
 a standard installation of ``21cmFAST`` will be automatically possible (see below).
 If they are installed to a place not on the ``LD_LIBRARY``/``INCLUDE`` paths, then you
-must use the compilation options (see below) to specify where they are. 
-For example, you can check if the header file for ``fftw3`` is 
+must use the compilation options (see below) to specify where they are.
+For example, you can check if the header file for ``fftw3`` is
 in its default location ``/usr/include/`` by running::
 
     cd /usr/include/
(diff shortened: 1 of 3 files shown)


INSTALLATION.rst
```diff
@@ -47,6 +47,39 @@ must use the compilation options (see below) to specify where they are.
           This is discussed below in the context of MacOSX, where it is often the
           easiest way to get the dependencies, but it is equally applicable to linux.
 
+Ubuntu
+^^^^^^
+It is best to use a system-wide installation of ``gcc, gsl,`` and ``fftw3`` rather than installing
```
steven-murray, line 52: separate gcc/gsl, like `gcc`, `gsl` and `ffw3`

steven-murray: I think you should first say something like 

```
If you are installing 21cmFAST just as a user, the very simplest 
method is `conda` -- with this method you simply need 
`conda install -c conda-forge 21cmFAST`, and all dependencies 
will be automatically installed. However, if you are going to use 
`pip` to install the package directly from the repository, there is 
a [bug in pip](https://stackoverflow.com/questions/71340058/conda-does-not-look-for-libpthread-and-libpthread-nonshared-at-the-right-place-w) 
that means it cannot find conda-installed shared libraries properly. 
In that case, it is much easier to install the basic dependencies 
(`gcc`, `gsl` and `fftw3`) with your system's package manager.
```

INSTALLATION.rst
```diff
@@ -47,6 +47,39 @@ must use the compilation options (see below) to specify where they are.
...
+them in a conda environment. ``gcc`` is by default available in Ubuntu.
+To check if ``gcc`` is installed, run ``gcc --version`` in your terminal.
+Install ``fftw3`` and ``gsl`` on your system with  ``sudo apt-get install libfftw3-dev libgsl-dev``.
+
+Once these are successfully installed, check where the ``fftw3`` library is actually located
+using ``locate libfftw3.so``. For example, it may be located in ``/usr/lib/x86_64-linux-gnu/``.
```
steven-murray, line 58: I don't *think* you need to say this, because I think `apt` will always place the packages on the search path.

INSTALLATION.rst
```diff
@@ -47,6 +47,39 @@ must use the compilation options (see below) to specify where they are.
...
+
+Make sure you can locate the header files as well, usually in
+``/usr/include/``. Check this by running::
+    cd /usr/include/
+    find fftw3.h
```
steven-murray, line 63: Same here.

steven-murray: Although, perhaps this info would be useful in the section above (i.e. generic "linux" instructions)

INSTALLATION.rst
```diff
@@ -47,6 +47,39 @@ must use the compilation options (see below) to specify where they are.
...
+or::
+
+    locate fftw3.h
+
+Lastly, locate your ``gcc`` installation using ``which gcc``.
+
+In your ``21cmfast`` environment, now install the ``21cmFAST`` package using::
+
+    cd /path/to/21cmFAST/
+    CC=/path/to/gcc LIB=/path/to/lib/ INC=/path/to/headers/include/ pip install
```
steven-murray, line 74: Again, you should not require the `CC` typically, unless you installed `gcc` with conda into this environment.

INSTALLATION.rst
```diff
@@ -47,6 +47,39 @@ must use the compilation options (see below) to specify where they are.
...
+
+If there is an issue during installation, add ``DEBUG=all`` or ``--DEBUG`` which may provide additional
+information.
+
+.. note:: You may choose to install ``gsl`` as an anaconda package as well, however, in that case,
+          you need to add the path to the gsl header in the environment e.g.
+          ``INC=/usr/include /path/to/conda/env/include/``
```
steven-murray, line 81: This shouldn't work... you'd need to specify two different `INC` statements (eg. `GSL_INC=/path/to/conda/env/include FFTW_INC=/usr/include`)

steven-murray (commented):
Thanks @DanielaBreitman! This is very useful. I've made a few comments that I think could be slightly better, so let's patch those up and then we'll be good to go.

steven-murray (approved):
Thanks @DanielaBreitman !

PR #322, opened 2023-03-17
The heating PR from debanjan, now a 21cmfast-native one

steven-murray:
This is the same PR, but we have a local branch of it now, so we can finish it off.


BradGreig:
Updates on this:

I've made some minor changes to the interpolation table to ensure its more stable and doesn't exceed its boundaries. I've also added som memory cleanup for the new arrays.

The plots below are for log10(L_X) = 38., and appear to retain their original behaviour after the implementation of this.
[deltaTb.pdf](https://github.com/21cmfast/21cmFAST/files/11049774/deltaTb.pdf)
[PowerSpectra.pdf](https://github.com/21cmfast/21cmFAST/files/11049776/PowerSpectra.pdf)

Tests are still failing though. I'm pretty confident that the failing of most tests is just numerical (rounding errors), with the tests being a little too rigid for noticing that zero and 10^-60 are effectively equivalent.

The one test of concern is "ts_fluct_no_tables". The test data is quite discrepant, which I have yet to be able to rectify. I'm hoping it turns out to be as simple as the test data was generated from an older commit pre changes.

BradGreig:
Hey @steven-murray, I think this PR is now ready. Not sure if this should go in before or after #320.

Additionally, this includes minor changes to address #325.

steven-murray:
@BradGreig great, thanks! I'll have a look today, and also try to finalize #320 .

steven-murray:
Adding my nominal approval to this PR (I can't approve since I made the PR technically).

Issue #371, opened 2024-03-28
Proper Implementation of High-Mass Stellar Fraction Turnover


daviesje:
In order to produce some predictions for galaxy number densities, we implemented a turnover in stellar mass at high halo masses within the `set_halo_properties` function in `HaloBox.c`. However this model is hardcoded and not optimised. It should be properly implemented with flags and parameters.

PR #379, opened 2024-04-08
Fixing some issues in the HaloBox struct and Mass Function Integrals.

daviesje:
This solves issue (#371) by improving the hardcoded double power law in the stellar to halo mass relation stellar mass, so that it uses a flag option `USE_UPPER_STELLAR_TURNOVER` to switch it on or off, and two global params `UPPER_STELLAR_TURNOVER_MASS` and `UPPER_STELLAR_TURNOVER_INDEX` to define the power law segment. This is currently only implmented in the halo model since it will only effect galaxy-based observables (the effect on the radiative backgrounds is minimal).

I have also fixed some issues with the function `set_fixed_grids`, a function used to estimate the average quantities of halo properties in each cell. This is used either when we do not want stochastic halo samples but we still want to use the halo-based radiative backgrounds (`USE_HALO_FIELD=True` `FIXED_HALO_GRIDS=True`), or when we want to add a population of halos below the minimum sampler mass (`AVG_BELOW_SAMPLER=True`). Since these mass function integrals are now used over a wider variety of upper/lower mass limits we now have to be more careful about how we set limits around the critical density, and so here I assume that if the cell is above critical density, the entire cell mass is within a single halo with mass exactly equal to the cell's Lagrangian mass (without `(1+delta)` since that is how the CMFs are defined).

This does involve a rather clumsy way of checking the mass limits (e.g line 1555 in ps.c `if(M_cond*(1-FRACT_FLOAT_ERR) <= exp(lnM2))`) since we only want to return the halo if it is within the mass integral. If the upper limit of the integral is defined exactly at the condition mass, floating point errors can result from taking the exponent and we end up on the wrong side of the conditional if we don't include the tolerance. If you think it's necessary, I can create a more robust way of checking the mass limit but I think this works well enough.

### DIFF
--- a/src/py21cmfast/inputs.py
+++ b/src/py21cmfast/inputs.py
@@ -315,12 +315,6 @@ class GlobalParams(StructInstanceWrapper):
     AVG_BELOW_SAMPLER:
         int flag to add the average halo proeprties in each cell between the source mass limit
         and the mass sampled by the halo sampler
-    UPPER_STELLAR_TURNOVER_MASS:
-        The pivot mass associated with the optional upper mass power-law of the stellar-halo mass relation
-        (see FlagOptions.USE_UPPER_STELLAR_TURNOVER)
-    UPPER_STELLAR_TURNOVER_INDEX:
-        The power-law index associated with the optional upper mass power-law of the stellar-halo mass relation
-        (see FlagOptions.USE_UPPER_STELLAR_TURNOVER)
     """
 
     def __init__(self, wrapped, ffi):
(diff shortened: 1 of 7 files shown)


src/py21cmfast/inputs.py
```diff
@@ -315,6 +315,12 @@ class GlobalParams(StructInstanceWrapper):
     AVG_BELOW_SAMPLER:
         int flag to add the average halo proeprties in each cell between the source mass limit
         and the mass sampled by the halo sampler
+    UPPER_STELLAR_TURNOVER_MASS:
+        The pivot mass associated with the optional upper mass power-law of the stellar-halo mass relation
+        (see FlagOptions.USE_UPPER_STELLAR_TURNOVER)
+    UPPER_STELLAR_TURNOVER_INDEX:
+        The power-law index associated with the optional upper mass power-law of the stellar-halo mass relation
+        (see FlagOptions.USE_UPPER_STELLAR_TURNOVER)
```
steven-murray, line 323: Is there a good reason these aren't in `AstroParams`? Eventually we want to move as many global params out as possible

tests/test_c_interpolation_tables.py
```diff
@@ -168,8 +169,10 @@ def test_Massfunc_conditional_tables(name):
         up.INTEGRATION_METHOD_HALOS,
     )
 
+    max_in_d = N_cmfi_cell[:, :1]
+    max_in_d[:, 0] = 1.0  # fix delta=-1
     N_cmfi_cell = (
-        N_cmfi_cell / N_cmfi_cell[:, :1]
+        N_cmfi_cell / max_in_d
```
steven-murray, line 175: something seems wrong here -- `max_in_d` is defined as all ones, then you're dividing by it, which does nothing?

steven-murray (commented):
Thanks @daviesje, this all seems good. Just a couple of small comments.

tests/test_c_interpolation_tables.py
daviesje: You're right, I reversed the indices. These are meant to be P(>M) (and so we divide by the first column at M = M_min) but with delta=-1 all entries are zero

src/py21cmfast/inputs.py
steven-murray: Yeah, I think we should move towards having ~no global parameters. However, we should think about this further in #9 and what criteria we want to have to potentially keep something as a global. But for now, let's assume we make all parameter-like things non-global.

daviesje: They've been moved to `AstroParams` now

PR #387, opened 2024-05-13
Large halo number fix

### LINKED ISSUE #384: Issue with halo sampler scaling with box length
Hello,

I closed my previous issue as it had to do with memory, but this seems like a separate concern. 

When generating lightcones, increasing the box length seems to reduce(?) the number of halos found. I discovered this while doing a parameter sweep of the global parameter 'SAMPLER_MIN_MASS' and found that settings which returned reasonable reionization histories at smaller box lengths would begin to break down at larger box lengths. Seemingly, no halos can be found once the box length exceeds a certain value, all other parameters being equal.

Attached is a doc showing lightcones generated at different values of 'SAMPLER_MIN_MASS' and box lengths. You can see that the larger boxes have a 0 halo field and no ionization history. 
[box length halo sampler test.pdf](https://github.com/21cmfast/21cmFAST/files/15249205/box.length.halo.sampler.test.pdf)

If you could look this over that would be appreciated.

daviesje:
Fixes #384. Some indices for halo number were still `int` type, which caused issues for boxes with billions of halos. They have been changed to `long long unsigned`.

### DIFF
--- a/src/py21cmfast/src/FindHaloes.c
+++ b/src/py21cmfast/src/FindHaloes.c
@@ -9,7 +9,7 @@
 // the virialized halos
 
 int check_halo(char * in_halo, struct UserParams *user_params, float R, int x, int y, int z, int check_type);
-void init_halo_coords(struct HaloField *halos, int n_halos);
+void init_halo_coords(struct HaloField *halos, long long unsigned int n_halos);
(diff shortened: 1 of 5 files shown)


steven-murray (approved):
Thanks @daviesje this looks good to me and tests pass. It would be nice to have a test that would *break* without this change, but I fear that would take more memory than we have on CI. So this is good. Just fix up the pre-commit and gtm.

Issue #374, opened 2024-04-02
Finalize Alternate Halo Sampling Methods


daviesje:
There currently exist four methods within the halos sampler for producing a population of halos from a condition, controlled by `global_params.HALO_SAMPLING_METHOD`:
- 0: (Default) Mass-limited sampling, where we draw from the CMF until the entire mass of the condition is in halos (including below the resolution
- 1: Number-limited sampling, where a Poisson variable is drawn for the total number of progenitors, and then the CMF is sampled that many times
- 2: Partitioning method from Sheth & Lemson 1999 (https://articles.adsabs.harvard.edu/pdf/1999MNRAS.305..946S). Where halos are partitioned from a condition based on the collapsed fraction distribution.
- 3: Binary split merger tree construction based on Qiu et al 2021 (https://arxiv.org/abs/2007.14624) and its predecessors (similar to GALFORM)

Since our goal was to abstract the model to any given HMF, methods 2 and 3 are fairly underdeveloped. They need to be tested under more parameter sets for accuracy and speed so we can provide better documentation on when each method may be useful.

This will also involve the examination of the mass tolerance implemented in methods 0 and 1. Where an entire sample is thrown out when it is too far from the target mass. It will likely be removed, or at least restricted to the initial sample, since it heavily disfavours certain samples resulting in the CMF skewing over time.

Issue #370, opened 2024-03-28
Alternative Conditional Mass Functions


daviesje:
Different halo mass functions have been handled previously by rescaling the integrals on the grid, which were calculated using the Extended Press-Schechter conditional mass function, to the mean values of a specified unconditional mass function. This was done separately for ionising emissivity and star formation rate density so that we always had the correct mean.

With a discrete halo catalogue this is not possible. We could apply the same corrections to the `HaloBox` grids, but then the emissivity grids would be inconsistent with the halo mass catalogues. Alternatives include:
- Implementing Sheth-Tormen (https://arxiv.org/pdf/astro-ph/0105113.pdf) and/or Delos (https://arxiv.org/pdf/2311.17986.pdf) conditional mass functions. In theory these should be much more accurate mass functions which could remove the necessity for mean fixing in all models. There are implementations of these CMFs in the code but they are not working. First samples drawn from them currently fit expected distributions very well, but drawing progenitors from these distributions quickly results huge errors. The Delos paper specifically warns that short timesteps (high delta) cause the conditional mass functions to not match random walk results, _however_ I would still expect samples drawn from the CMF to sum to the unconditional mass function across all conditions, as long as it always integrates to one. I am unsure at the moment if there is a bug in the implementation or if the approach itself is restricted to Extended Press-Schechter somehow.
- Rescaling the EPS mass function: This has been notably done in binary-split merger tree construction (Parkinson+08), where scaling terms are added to the EPS CMF (with condition scale and density) so that the final halo populations match expected results (usually an N-body).
- Rescaling unconditional mass functions: Different forms have been implemented in https://arxiv.org/pdf/0803.1954.pdf, https://arxiv.org/pdf/1702.01788.pdf, which rescale a given unconditional mass function, resulting in a CMF that can be used.

PR #388, opened 2024-05-15
Inverse table improvements

daviesje:
This PR mainly contains improvements to the HMF tables used in the halo sampler.

For the inverse CDF tables used in the Sampling:
- I replaced the interpolation on the CDF with a rootfind to get the inverse, this is more accurate so we require fewer bins.
- The output is no longer log-mass, but the ratio of mass to condtition mass, this is slightly faster and makes the table interpolation a little more accurate.
- Both tables for grid and halo conditions are now in log-probability, this should allow longer timesteps in the halo sampler (see #376).
- Below the minimum log-probability I have added an extrapolation function, which assumes exponential decay in p(M)

The Sheth-Tormen conditional mass function has been fixed and optimized (fixing #370), The Delos+24 conditional mass function is available for the default case but does not perform well in the sampler due to the small timestep.

Alternate sampling methods (Sheth+1999 Partition and Parkinson+08 binary split) have been fixed (#374), they are less conforming to the given CMF but as far as I can tell they are working.

Old global parameters have been removed and many sampler related global parameters have been moved to the input structures.


src/py21cmfast/inputs.py
```diff
@@ -532,6 +502,36 @@ class UserParams(StructWithDefaults):
...
+    SAMPLE_METHOD: int, optional
+        The sampling method to use in the halo sampler when calculating progenitor populations:
+        0: Mass-limited CMF sampling, where samples are drawn until the expected mass is reached
+        1: Number-limited CMF sampling, where we select a number of halos from the Poisson distribution
+        and then sample the CMF that many times
+        2: Sheth et al 1999 Partition sampling, where the EPS collapsed fraction is sampled (gaussian tail)
+        and then the condition is updated using the conservation of mass.
+        3: Parkinsson et al 2008 Binary split model as in DarkForest (Qiu et al 2021) where the EPS merger rate
+        is sampled on small internal timesteps such that only binary splits can occur.
+        NOTE: Sampling from the density grid will ALWAYS use number-limited sampling (method 1)
```
steven-murray, line 527: Can we make the python-side parameter either a string or an enum? This is more explicit. It's harder to read code that just has `SAMPLE_METHOD=2` rather than `SAMPLE_METHOD=samplers.partition` or `SAMPLE_METHOD='partition'`

src/py21cmfast/inputs.py
```diff
@@ -532,6 +502,36 @@ class UserParams(StructWithDefaults):
...
+    AVG_BELOW_SAMPLER: bool, optional
+        When switched on, an integral is performed in each cell between the minimum source mass and SAMPLER_MIN_MASS,
+        effectively placing the average halo population in each HaloBox cell below the sampler resolution
```
steven-murray, line 530: What happens if it's _not_ true? Those masses are just ignored totally?

src/py21cmfast/inputs.py
```diff
@@ -532,6 +502,36 @@ class UserParams(StructWithDefaults):
...
+    HALOMASS_CORRECTION: float, optional
+        This provides a corrective factor to the mass-limited (SAMPLE_METHOD==0) sampling, which multiplies the
+        expected mass from a condition by this number.
+        This also is used in the partition (SAMPLE_METHOD==2) sampler, multiplying sigma(M) of each sample drawn.
```
steven-murray, line 534: What's the usefulness of this option? Why is its default 0.9?

src/py21cmfast/src/FindHaloes.c
```diff
@@ -166,7 +162,7 @@ LOG_DEBUG("redshift=%f", redshift);
 
             //Pending a serious deep-dive into this algorithm, I will force DexM to use the fitted parameters to the
             //      Sheth-Tormen mass function (as of right now, We do not even reproduce EPS results)
```
steven-murray, line 164: is this comment still true?

src/py21cmfast/src/Stochasticity.c
```diff
@@ -401,121 +350,90 @@ int fix_mass_sample(gsl_rng * rng, double exp_M, int *n_halo_pt, double *M_tot_p
...
+ * Stochasticity is ignored below a certain mass threshold */
 int stoc_mass_sample(struct HaloSamplingConstants * hs_constants, gsl_rng * rng, int *n_halo_out, float *M_out){
-    //lnMmin only used for sampling, apply factor here
-    double mass_tol = global_params.STOC_MASS_TOL;
     double exp_M = hs_constants->expected_M;
-    if(hs_constants->from_catalog)exp_M *= 1.; //fudge factor for assuming that internal lagrangian volumes are independent
 
-    int n_halo_sampled, n_failures=0;
+    //fudge factor for assuming that internal lagrangian volumes are independent
+    exp_M *= user_params_stoc->HALOMASS_CORRECTION;
```
steven-murray, line 364: this comment and code are unclear to me

src/py21cmfast/src/Stochasticity.c
```diff
@@ -1252,13 +1148,38 @@ int single_test_sample(struct UserParams *user_params, struct CosmoParams *cosmo
                     }
                 }
                 //output descendant statistics
-                out_n_exp[j] = hs_constants_priv.expected_N;
-                out_m_exp[j] = hs_constants_priv.expected_M;
+                // out_n_exp[j] = hs_constants_priv.expected_N;
+                // out_m_exp[j] = hs_constants_priv.expected_M;
```
steven-murray, line 1152: could just delete these lines.

src/py21cmfast/src/UsefulFunctions.c
```diff
@@ -27,7 +27,8 @@ void seed_rng_threads(gsl_rng * rng_arr[], int seed){
     unsigned int seeds[user_params_ufunc->N_THREADS];
 
     // For multithreading, seeds for the RNGs are generated from an initial RNG (based on the input random_seed) and then shuffled (Author: Fred Davies)
-    int num_int = INT_MAX/16; //JD: this was taking a few seconds per snapshot so i reduced the number TODO: init the RNG once
+    // int num_int = INT_MAX/16;
+    int num_int = INT_MAX/256; //JD: this was taking a few seconds per snapshot so i reduced the number TODO: init the RNG once
```
steven-murray, line 31: is the comment here still applicable?

steven-murray (commented):
This looks great, @daviesje. Just a few small comments and I think we'll be gtm.

src/py21cmfast/inputs.py
daviesje: Yes, I would imagine that most of the time `SAMPLER_MIN_MASS` would be set low enough so that the averaging is unnecessary, and that this would be for saving memory on larger boxes.

steven-murray: Cool -- I think a sentence describing that would be helpful

src/py21cmfast/inputs.py
daviesje: The mass-limited sampler has a slight bias toward having too many halos, this bias is independent of delta or halo mass, so this factor is a correction to that. The value of 0.9 works well for the default timestep factor of 1.02. Since the partition method also uses a single correction factor (Mcquinn+ 2007 lowers sigma_8 slightly, I've implemented a correction in nu) we re-use the parameter

src/py21cmfast/src/FindHaloes.c
daviesje: Yes, I think it would take some time / code archaeology to fully understand why the excursion set halo finder gets the results it does, and find suitable corrections for any given CMF.

src/py21cmfast/src/Stochasticity.c
daviesje: done

src/py21cmfast/src/Stochasticity.c
daviesje: This is the implementation of the above correction in the mass-limited sampling. One of the approximations in our sample is that we sample the final CMF in an uncorrelated way, assuming each sampled halo is independent of the others. This means that the CMF doesn't change after each sample, like it does in the partition or binary split methods. This results in a bias where the last halo sampled is on average larger.

I'll change the comment to something more generic and expand the explanation in `inputs.py`

src/py21cmfast/src/UsefulFunctions.c
daviesje: Yes, I had un-done this change in the initial v4-prep commit, but it takes a few seconds per snapshot. In a future commit it would be useful to only initialise the RNG once, along with some interpolation tables like sigma.

daviesje: For now I just reduced the maximum seed number

src/py21cmfast/inputs.py
steven-murray: OK. Then I think this should be documented here (i.e. probably don't touch this unless you have good reason, and these are the values it should take in these circumstances)

src/py21cmfast/src/Stochasticity.c
steven-murray: Thanks!

src/py21cmfast/inputs.py
```diff
@@ -603,6 +617,38 @@ def HII_tot_num_pixels(self):
...
+        try:
+            val = int(val)
+        except (ValueError, TypeError) as e:
+            raise ValueError(f"{val} is an invalid value for {propname}") from e
+
+        if not 0 <= val < len(enum_list):
+            raise ValueError(f"HMF must be an int between 0 and {len(enum_list) - 1}")
+
+        return val
+
```
daviesje, line 651: Before the merge, do you think this is a good way to implement the enums? I found trying to work with the python enum.IntEnum a bit more cumbersome

steven-murray (approved):

PR #389, opened 2024-05-15
Allow halo catalogue cuts to output averages correctly

daviesje:
This is a small change which affects cases where a user wants to cut halos from a catalogue before making a grid. For convenience we allow the setting of the mass to zero to acheive this, rather than creating a whole new output struct which needs to be `.compute()`d. This change cleans up the implementation of that case a little bit. Previously, if halos were cut, certain average quantities would not be calculated correctly, which influences the debug output, as well as the averaging of the mass turnover fields. Now cut halos are no longer counted as zero-mass halos in these averages.


steven-murray (approved):
Looks good to me!

PR #392, opened 2024-05-21
Lightcone distance unit fix

daviesje:
Since the angular lightcone update, the `lightcone_distances` field has units. However these units are not saved and loaded from the HDF5 Files.

This is a quick fix which adds the units when required. It might be a better long-term idea to properly save the units of all quantities along with the values, however this will take more time, and currently you cannot call `lightcone_redshifts` on a lightcone loaded from file.

I have also added a `lightcone_redshift` check to `test_lightcone_roundtrip` to catch this in future.


steven-murray (approved):
Thanks @daviesje, this is very much necessary. One small comment and you're good.

PR #394, opened 2024-05-22
fix return value order in make_lightcone_slices

daviesje:
In some lightcone runs, I noticed a crash when the simulation was run at higher redshifts than the lightconer outputs. It seems that the case where we don't find any interpolation outputs in `make_lightcone_slices` yielded arguments in the wrong order, resulting in a KeyError when we index the lightcone. This has been fixed.

Before I merge, I would like to add to the tests to check for this, however we already run lightcones in the tests which have maximum redshifts below `Z_HEAT_MAX` and I'm having trouble figuring out why it doesn't crash there.

### DIFF
--- a/src/py21cmfast/lightcones.py
+++ b/src/py21cmfast/lightcones.py
@@ -162,11 +162,13 @@ class Lightconer(ABC):
 
         Returns
         -------
+        quantity
+            The field names of the quantities required by the lightcone.
+        lcidx
+            The indices of the lightcone to which these slices belong.
         scalar_field_slices
             The scalar fields evaluated on the "lightcone" slices that exist within
             the redshift range spanned by ``c1`` and ``c2``.
-        lcidx
-            The indices of the lightcone to which these slices belong.
         """
         if c1.user_params != c2.user_params:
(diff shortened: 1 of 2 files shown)


daviesje:
The remaining thing I can't figure out is that the tests (e.g. test_lightcone_quantities) have a default `max_redshift` of 25, and as far as I can tell don't change the default `Z_HEAT_MAX` from 35. So I would expect these to fail as well

steven-murray (approved):

daviesje:
In `produce_lc_power_spectra` for the integration tests:
```
# NOTE: this is here only so that we get the same answer as previous versions,
    #       which have a bug where the max_redshift gets set higher than it needs to be.
    flag_options = FlagOptions(options.pop("flag_options"))
    if flag_options.INHOMO_RECO or flag_options.USE_TS_FLUCT:
        max_redshift = options.get("z_heat_max", global_params.Z_HEAT_MAX)
        del options["redshift"]
    else:
        max_redshift = options.pop("redshift") + 2
```

This explicitly sets max redshift to be `Z_HEAT_MAX` when the bug would have occurred.

This should be deleted when we regenerate the test data for v4

Issue #85, opened 2020-01-31
[Feature Req.] Allow the user to set default settings which remain after session is closed


andreimesinger:
**Is your feature request related to a problem? Please describe.**

Current defaults are very "physics lite", which is ok maybe for the typical (observational) user, but users wanting more realistic calculations should have the default settings of, e.g. Park et al.

**Describe the solution you'd like**

Provide alternate default configurations, like a "quick and easy" one, and a "physically realistic" one.  Allow the user to save their configurations so that they are default whenever they open an new session.

**Describe alternatives you've considered**

**Additional context**

JulianBMunoz:
That's a neat idea. I'd also add a cosmic-dawn default setting, which does not need to do 2LPT (either regular or hi-res), since that doesn't matter for z>10, but instead defaults to using minihaloes, LW feedback, and relative velocities, all of which matter for cosmic dawn.

steven-murray:
Done in #444 thank you @daviesje

PR #397, opened 2024-06-04
change defaults in the branch to the halo sampler model

daviesje:
Eventually we will want to have default parameters which are conditional, dependent on the specified parameters with some kind of heirarchy. However this will take a while.

For now, I think It's safe to assume that anyone currently using the `v4-prep` branch is more likely to want to use the halo model than not, so I've set the defaults to the halo sampler, exponential filter for ionising emissivity, recombinations in-cell, and the spherical top-hat filter for density.


steven-murray (approved):
Agreed @daviesje. Actually, the idea will be to provide configuration "profiles" so that many settings can be changed with a single line, in a consistent way (this will of course be in tandem with dependent parameters and validation). See https://github.com/21cmfast/21cmFAST/issues/85

Issue #404, opened 2024-07-03
Add another turnover mass to model supernova feedback


daviesje:
Currently, minihalo turnover mass is set by the minimum of the LW feedback turnover and the reionization feedback turnover.

We want to include another turnover mass to represent supernova feedback, which would be a parameter which sets another floor on the turnover mass, possibly different for ACG/MCG components.

PR #406, opened 2024-07-05
Lx scatter and Nikolic+2024 scaling relations

daviesje:
Currently WIP PR for adding the xray luminosity scatter, I am currently testing the outputs to make sure we get what we expect.

#### Changes:
- Added another RNG field to `HaloField` and `PerturbHaloField` for the scatter in xray luminosity
- Added another field to the `HaloBox` and `XraySourceBox` to represent the xrays
- Added scaling relations for Metallicity, xray luminosity and mass-dependent SFR scatter to `HaloBox.c`
- Replaced the SFR field with the xray field in `SpinTemperature.c` for the xray calculations (lyman alpha / lyman werner still uses SFR)
- Refactored `HaloBox.c` to be more easily extendable and readable

### DIFF
--- a/src/py21cmfast/inputs.py
+++ b/src/py21cmfast/inputs.py
@@ -978,15 +978,16 @@ class AstroParams(StructWithDefaults):
         Power-law index of fraction of galactic gas in stars as a function of halo mass, for MCGs.
         See Sec 2 of Muñoz+21 (2110.13919).
     SIGMA_STAR : float, optional
-        Lognormal scatter of the halo mass to stellar mass relation.
+        Lognormal scatter (dex) of the halo mass to stellar mass relation.
         Uniform across all masses and redshifts.
     CORR_STAR : float, optional
         Self-correlation length used for updating halo properties. Properties are interpolated between
         a random sample at the current halo mass and one matching the point in the PDF of the
         previous sample, the interpolation point in [0,1] is given by exp(-dz/CORR_STAR)
-    SIGMA_SFR : float, optional
-        Lognormal scatter of the stellar mass to SFR relation.
-        Uniform across all masses and redshifts.
+    SIGMA_SFR_LIM : float, optional
+        Lognormal scatter (dex) of the stellar mass to SFR relation above a stellar mass of 1e10 solar.
+    SIGMA_SFR_INDEX : float, optional
+        index of the power-law between SFMS scatter and stellar mass below 1e10 solar.
     CORR_SFR : float, optional
         Self-correlation length used for updating halo properties. Properties are interpolated between
         a random sample at the current halo mass and one matching the point in the PDF of the
@@ -1054,6 +1055,13 @@ class AstroParams(StructWithDefaults):
     UPPER_STELLAR_TURNOVER_INDEX:
         The power-law index associated with the optional upper mass power-law of the stellar-halo mass relation
         (see FlagOptions.USE_UPPER_STELLAR_TURNOVER)
+    SIGMA_LX: float, optional
+        Lognormal scatter (dex) of the Xray luminosity relation.
+        Uniform across all masses and redshifts.
+    CORR_LX : float, optional
+        Self-correlation length used for updating halo properties. Properties are interpolated between
+        a random sample at the current halo mass and one matching the point in the PDF of the
+        previous sample, the interpolation point in [0,1] is given by exp(-dz/CORR_STAR)
     """
 
     _ffi = ffi
@@ -1064,9 +1072,10 @@ class AstroParams(StructWithDefaults):
         "F_STAR7_MINI": -2.0,
         "ALPHA_STAR": 0.5,
         "ALPHA_STAR_MINI": 0.5,
-        "SIGMA_STAR": 0.5,
+        "SIGMA_STAR": 0.25,
         "CORR_STAR": 0.5,
-        "SIGMA_SFR": 0.6,
+        "SIGMA_SFR_LIM": 0.19,
+        "SIGMA_SFR_INDEX": -0.12,
         "CORR_SFR": 0.2,
         "F_ESC10": -1.0,
         "F_ESC7_MINI": -2.0,
@@ -1087,7 +1096,9 @@ class AstroParams(StructWithDefaults):
         "A_VCB": 1.0,
         "BETA_VCB": 1.8,
         "UPPER_STELLAR_TURNOVER_MASS": 11.447,  # 2.8e11
-        "UPPER_STELLAR_TURNOVER_INDEX": -0.61,
+        "UPPER_STELLAR_TURNOVER_INDEX": -0.6,
+        "SIGMA_LX": 0.5,
+        "CORR_LX": 0.2,  # NOTE (Jdavies): It's difficult to know what this should be, ASTRID doesn't have the xrays and I don't know which hydros do
     }
 
     def __init__(
@@ -1112,7 +1123,9 @@ class AstroParams(StructWithDefaults):
             "X_RAY_Tvir_MIN",
             "UPPER_STELLAR_TURNOVER_MASS",
         ]:
-            return 10**val
+            return 10**val  # log10 to linear conversion
+        if key in ["SIMGA_STAR" "SIGMA_SFR_LIM" "SIGMA_LX"]:
+            return 2.3025851 * val  # dex to base e conversion
         else:
             return val
 
--- a/src/py21cmfast/inputs.py
+++ b/src/py21cmfast/inputs.py
@@ -1037,12 +1037,12 @@ class AstroParams(StructWithDefaults):
         Given in log10 units.
     L_X : float, optional
         The specific X-ray luminosity per unit star formation escaping host galaxies.
-        Cf. Eq. 6 of Greig+2018. Given in log10 units. For models which use metallicity this gives
-        the value at solar metallicity.
+        Cf. Eq. 6 of Greig+2018. Given in log10 units. For the double power-law used in the Halo Model
+        This gives the low-z limite.
     L_X_MINI: float, optional
         The specific X-ray luminosity per unit star formation escaping host galaxies for
-        minihalos. Cf. Eq. 23 of Qin+2020. Given in log10 units. For models which use metallicity this gives
-        the value at solar metallicity.
+        minihalos. Cf. Eq. 23 of Qin+2020. Given in log10 units. For the double power-law used in the Halo Model
+        This gives the low-z limite.
     NU_X_THRESH : float, optional
         X-ray energy threshold for self-absorption by host galaxies (in eV). Also called
         E_0 (cf. Sec 4.1 of Greig+2018). Typical range is (100, 1500).
@@ -1094,8 +1094,8 @@ class AstroParams(StructWithDefaults):
         "M_TURN": None,
         "R_BUBBLE_MAX": None,
         "ION_Tvir_MIN": 4.69897,
-        "L_X": 39.46,  # Kaur+22 and Brorby+16
-        "L_X_MINI": 39.46,
+        "L_X": 40.5,  # Kaur+22
+        "L_X_MINI": 40.5,
         "NU_X_THRESH": 500.0,
         "X_RAY_SPEC_INDEX": 1.0,
         "X_RAY_Tvir_MIN": None,
--- a/src/py21cmfast/inputs.py
+++ b/src/py21cmfast/inputs.py
@@ -1025,10 +1025,12 @@ class AstroParams(StructWithDefaults):
         Given in log10 units.
     L_X : float, optional
         The specific X-ray luminosity per unit star formation escaping host galaxies.
-        Cf. Eq. 6 of Greig+2018. Given in log10 units.
+        Cf. Eq. 6 of Greig+2018. Given in log10 units. For models which use metallicity this gives
+        the value at solar metallicity.
     L_X_MINI: float, optional
         The specific X-ray luminosity per unit star formation escaping host galaxies for
-        minihalos. Cf. Eq. 23 of Qin+2020. Given in log10 units.
+        minihalos. Cf. Eq. 23 of Qin+2020. Given in log10 units. For models which use metallicity this gives
+        the value at solar metallicity.
     NU_X_THRESH : float, optional
         X-ray energy threshold for self-absorption by host galaxies (in eV). Also called
         E_0 (cf. Sec 4.1 of Greig+2018). Typical range is (100, 1500).
@@ -1080,8 +1082,8 @@ class AstroParams(StructWithDefaults):
         "M_TURN": 8.7,
         "R_BUBBLE_MAX": None,
         "ION_Tvir_MIN": 4.69897,
-        "L_X": 38.44,  # Nikolic et al. 2024 L_X at 10^10 solar
-        "L_X_MINI": 38.44,  # Nikolic et al. 2024
+        "L_X": 39.46,  # Kaur+22 and Brorby+16
+        "L_X_MINI": 39.46,
         "NU_X_THRESH": 500.0,
         "X_RAY_SPEC_INDEX": 1.0,
         "X_RAY_Tvir_MIN": None,
--- a/src/py21cmfast/inputs.py
+++ b/src/py21cmfast/inputs.py
@@ -931,11 +931,23 @@ class FlagOptions(StructWithDefaults):
         return self._HALO_STOCHASTICITY
 
     @property
+    def USE_EXP_FILTER(self):
+        """Automatically setting USE_EXP_FILTER to False if not HII_FILTER==0."""
+        if self._USE_EXP_FILTER and global_params.HII_FILTER != 0:
+            warnings.warn(
+                "USE_EXP_FILTER can only be used with a real-space tophat HII_FILTER==0"
+                "Setting USE_EXP_FILTER to False"
+            )
+            return False
+        return self._USE_EXP_FILTER
+
+    @property
     def CELL_RECOMB(self):
         """Automatically setting CELL_RECOMB if USE_EXP_FILTER is active."""
         if self.USE_EXP_FILTER and not self._CELL_RECOMB:
             warnings.warn(
                 "CELL_RECOMB is automatically set to True if USE_EXP_FILTER is True."
+                "setting CELL_RECOMB to True"
             )
             return True
         return self._CELL_RECOMB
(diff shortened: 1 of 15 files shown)


src/py21cmfast/inputs.py
```diff
@@ -1054,6 +1055,13 @@ class AstroParams(StructWithDefaults):
     UPPER_STELLAR_TURNOVER_INDEX:
         The power-law index associated with the optional upper mass power-law of the stellar-halo mass relation
         (see FlagOptions.USE_UPPER_STELLAR_TURNOVER)
+    SIGMA_LX: float, optional
+        Lognormal scatter (dex) of the Xray luminosity relation.
+        Uniform across all masses and redshifts.
+    CORR_LX : float, optional
+        Self-correlation length used for updating halo properties. Properties are interpolated between
+        a random sample at the current halo mass and one matching the point in the PDF of the
+        previous sample, the interpolation point in [0,1] is given by exp(-dz/CORR_STAR)
```
steven-murray, line 1064: does this mean to say `CORR_LX` instead of `CORR_STAR`? 

I *think* I understand this explanation, but I'm not quite sure: which properties are interpolated? All of them? "Previous" means higher redshift?

src/py21cmfast/inputs.py
```diff
@@ -1054,6 +1055,13 @@ class AstroParams(StructWithDefaults):
     UPPER_STELLAR_TURNOVER_INDEX:
         The power-law index associated with the optional upper mass power-law of the stellar-halo mass relation
         (see FlagOptions.USE_UPPER_STELLAR_TURNOVER)
+    SIGMA_LX: float, optional
+        Lognormal scatter (dex) of the Xray luminosity relation.
```
steven-murray, line 1059: Xray luminosity with what? SFR?

src/py21cmfast/inputs.py
```diff
@@ -1087,7 +1096,9 @@ class AstroParams(StructWithDefaults):
         "A_VCB": 1.0,
         "BETA_VCB": 1.8,
         "UPPER_STELLAR_TURNOVER_MASS": 11.447,  # 2.8e11
-        "UPPER_STELLAR_TURNOVER_INDEX": -0.61,
+        "UPPER_STELLAR_TURNOVER_INDEX": -0.6,
+        "SIGMA_LX": 0.5,
+        "CORR_LX": 0.2,  # NOTE (Jdavies): It's difficult to know what this should be, ASTRID doesn't have the xrays and I don't know which hydros do
```
steven-murray, line 1101: This note reminds me a good point: we should whenever possible note down where the default value comes from

src/py21cmfast/src/HaloBox.c
```diff
@@ -1,105 +1,276 @@
...
+};
+
+//struct holding each halo property we currently need.
+//This is only used for both averages over the box/catalogues
+//  as well as an individual halo's properties
+struct HaloProperties{
+    double halo_mass;
+    double stellar_mass;
+    double halo_sfr;
+    double stellar_mini;
```
steven-murray, line 51: in the spirit of using more explicit names, I think this one is a little cryptic

src/py21cmfast/src/HaloBox.c
```diff
@@ -1,105 +1,276 @@
...
+    double sfr_mini;
+    double fescweighted_sfr;
+    double n_ion;
+    double halo_xray;
+    double m_turn_a;
```
steven-murray, line 56: these (`m_turn_a/m/r`) are also not obvious what they refer to.

src/py21cmfast/src/HaloBox.c
```diff
@@ -1,105 +1,276 @@
...
+    //intermediates
+    double fstar_mean;
+    double f_sample, f_sample_mini;
+    double sm_sample,sm_sample_mini;
+
+    double baryon_ratio = cosmo_params_stoc->OMb / cosmo_params_stoc->OMm;
+
+    /* Simply adding lognormal scatter to a delta increases the mean (2* is as likely as 0.5*)
+    * We multiply by exp(-sigma^2/2) so that X = exp(mu + N(0,1)*sigma) has the desired mean */
+    star_rng = -sigma_star*sigma_star/2 + star_rng*sigma_star;
```
steven-murray, line 137: Hold on -- is `star_rng` a random number _generator_ here? Or is it a random number drawn from N(0,1)? Also, I'm not sure I quite understand the notation -- is `-sigma^2/2` equal to your `mu`? Also, is adjusting the distribution so its mean in linear space is correct the right thing to do? Are we sure that the given mean doesn't represent the log-space mean?

src/py21cmfast/src/HaloBox.c
```diff
@@ -179,93 +351,71 @@ int set_fixed_grids(double redshift, double norm_esc, double alpha_esc, double M
...
-                                                            alpha_esc, norm_star_mini, norm_esc_mini, Mlim_Fstar_mini,
-                                                            Mlim_Fesc_mini, user_params_stoc->INTEGRATION_METHOD_MINI),
-                                    Nion_ConditionalM_MINI(growth_z, lnMmin, lnMmax, M_cell, sigma_cell,
-                                                dens, M_turn_m, M_turn_a, alpha_star_mini,
-                                                0., norm_star_mini, 1., Mlim_Fstar_mini, 0.,
-                                                user_params_stoc->INTEGRATION_METHOD_MINI));
-                }
-                LOG_SUPER_DEBUG("ACG MTURN %.3e (%.3e) MCG MTURN %.3e (%.3e) REION %.3e",M_turn_a_nofb,M_turn_a,M_turn_m_nofb,M_turn_m,M_turn_r);
-            }
+            grids->halo_mass[i] = mass_intgrl * prefactor_mass * (1+dens);
```
steven-murray, line 395: not a big thing but `(1+dens)` could be pre-computed and used for all these lines

src/py21cmfast/src/HaloBox.c
```diff
@@ -329,32 +477,219 @@ int get_box_averages(double redshift, double norm_esc, double alpha_esc, double
...
+            m = halos->halo_masses[i_halo];
+            //It is sometimes useful to make cuts to the halo catalogues before gridding.
+            //  We implement this in a simple way, if the user sets a halo's mass to zero we skip it
+            if(m == 0.){
+                n_halos_cut++;
+                continue;
+            }
+            x = halos->halo_coords[0+3*i_halo]; //NOTE:PerturbHaloField is on HII_DIM, HaloField is on DIM
+            y = halos->halo_coords[1+3*i_halo];
+            z = halos->halo_coords[2+3*i_halo];
```
steven-murray, line 578: are you hard-coding `DIM = 3*HII_DIM` here?

src/py21cmfast/src/HaloBox.c
```diff
@@ -329,32 +477,219 @@ int get_box_averages(double redshift, double norm_esc, double alpha_esc, double
...
+                if(!flag_options_stoc->FIX_VCB_AVG && user_params_stoc->USE_RELATIVE_VELOCITIES)
+                    curr_vcb = ini_boxes->lowres_vcb[HII_R_INDEX(x,y,z)];
+
+                J21_val = redshift < global_params.Z_HEAT_MAX ? previous_spin_temp->J_21_LW_box[HII_R_INDEX(x,y,z)] : 0.;
+                Gamma12_val = redshift < global_params.Z_HEAT_MAX ? previous_ionize_box->Gamma12_box[HII_R_INDEX(x, y, z)] : 0.;
+                zre_val = redshift < global_params.Z_HEAT_MAX ? previous_ionize_box->z_re_box[HII_R_INDEX(x, y, z)] : 0.;
+                M_turn_m = lyman_werner_threshold(redshift, J21_val, curr_vcb, astro_params_stoc);
+                M_turn_r = reionization_feedback(redshift, Gamma12_val, zre_val);
+                M_turn_a = consts->mturn_a_nofb < M_turn_r ? M_turn_r : consts->mturn_a_nofb; //since ACG turnover not reassigned
+                if(M_turn_m < M_turn_r)M_turn_m = M_turn_r;
```
steven-murray, line 592: why use the `if` here instead of the `?`?

steven-murray: Actually, the set of three lines above comparing `redshift` to `Z_HEAT_MAX` would probably be easier to read if you just had:

```c
J21_val=0.; Gamma12_val =0.; zre_val = 0;
if(redshift < global_params.Z_HEAT_MAX){
    J21_val = J_21_LW_box[HII_R_INDEX(x,y,z)];
   etc.
```

src/py21cmfast/src/HaloBox.c
```diff
@@ -329,32 +477,219 @@ int get_box_averages(double redshift, double norm_esc, double alpha_esc, double
...
+            nion = out_props[2];
+            wsfr = out_props[3];
+            stars = out_props[4];
+            stars_mini = out_props[5];
+            xray = out_props[6];
+
+            if(x+y+z == 0){
+                LOG_ULTRA_DEBUG("Cell 0 Halo %d: HM: %.2e SM: %.2e (%.2e) NI: %.2e SF: %.2e (%.2e) WS: %.2e",i_halo,m,stars,stars_mini,nion,sfr,sfr_mini,wsfr);
+                LOG_ULTRA_DEBUG("Mturn_a %.2e Mturn_m %.2e",M_turn_a,M_turn_m);
+            }
```
steven-murray, line 613: again small thing, but since you have to do a computation to check whether you should log this, it'd be better if it was inside a pragma thing, like `#if LOG_LEVEL>=ULTRA_DEBUG`

src/py21cmfast/src/Stochasticity.c
```diff
@@ -232,17 +237,20 @@ int set_prop_rng(gsl_rng *rng, int from_catalog, double *interp, float * input,
...
     double prop2 = gsl_ran_ugaussian(rng);
+    double prop3 = gsl_ran_ugaussian(rng);
 
     //Correlate properties by interpolating between the sampled and descendant gaussians
     //THIS ASSUMES THAT THE SELF-CORRELATION IS IN THE LOG PROPRETY, NOT THE PROPERTY ITSELF
     //IF IT SHOULD BE IN LINEAR SPACE, EXPONENTIATE THE RANDOM VARIABLES
     if(from_catalog){
         prop1 = (1-interp[0])*prop1 + interp[0]*input[0];
         prop2 = (1-interp[1])*prop1 + interp[1]*input[1];
+        prop3 = (1-interp[2])*prop1 + interp[2]*input[2];
```
steven-murray, line 248: I don't see why this (and below) shouldn't just be in a loop over the inputs

steven-murray (commented):
Thanks @daviesje, with a naive look at this, it all looks good. I made a few comments about documentation and small performance things, but otherwise should be gtg.

Well, I suppose the other thing needed is some tests...

src/py21cmfast/src/HaloBox.c
daviesje: Nope, the `halo_coords` field is laid out (x0,y0,z0,x1,y1,z1...) the 3 is for the 3 dimensions

src/py21cmfast/src/HaloBox.c
daviesje: I would think any compiler should delete the line `if(x+y+z==0){}` but your way would be clearer

src/py21cmfast/src/Stochasticity.c
daviesje: I'm actually planning on going the other direction with this, and changing `prop1`,`prop2`... to `rng_star`,`rng_sfr`...

src/py21cmfast/src/HaloBox.c
daviesje: This is the adjustment from the mean of the halo property to the mean log of the halo property. I've set the parameters up to be the mean of whatever distribution you want, so that the mean doesn't change when adjusting the sigmas.

as for the `_rng` notation, you're right this is a little confusing, these are the correlated random numbers drawn from the normal distribution, so that when we multiply `(mean * exp(X_rng))` we get the lognormal scatter.

This function is not super clear, since the steps were all shuffled around to minimise exp/log calls.

src/py21cmfast/src/HaloBox.c
daviesje: In general, I use ternaries `a = X ? b : c`  when neither `b==a` or `c==a`, when one of the cases leaves the variable alone I use `if`.

I'll change both sets to initialisation + `if` for readability.

src/py21cmfast/inputs.py
daviesje: The mean is a function of SFR, stellar mass and redshift, but the lognormal scatter is constant. I've added this to the description.

src/py21cmfast/src/HaloBox.c
daviesje: One can argue whether the parameter should represent the mean in linear or log. I'm open to suggestion either way but zeroth order I would think that someone just changing the sigma shouldn't expect the global history to shift dramatically.

src/py21cmfast/inputs.py
steven-murray: OK, that is good -- however I still feel like it should be `CORR_LX` instead of `CORR_STAR`? You're describing the parameter `CORR_LX`.

src/py21cmfast/inputs.py
steven-murray: Thanks for adding the refs!

src/py21cmfast/src/HaloBox.c
```diff
@@ -784,5 +806,105 @@ int ComputeHaloBox(double redshift, struct UserParams *user_params, struct Cosmo
         return(status);
     }
     LOG_DEBUG("Done.");
-    return 0.;
+    return 0;
+}
+
+//test function for getting halo properties from the wrapper, can use a lot of memory for large catalogs
+int halo_props_exposed(double redshift, struct UserParams *user_params, struct CosmoParams *cosmo_params, struct AstroParams *astro_params,
```
steven-murray, line 813: I don't love having `exposed` in the name of the function -- I would just call this `get_halo_properties` or similar, and _expose_ it.

src/py21cmfast/src/UsefulFunctions.c
```diff
@@ -716,7 +716,7 @@ float ComputeTau(struct UserParams *user_params, struct CosmoParams *cosmo_param
     return tau;
 }
 
-
+//TODO: THESE NEED UPDATING
```
steven-murray, line 719: OK, let's update it :-)

src/py21cmfast/inputs.py
steven-murray: Ah, I see all is good, sorry.

src/py21cmfast/src/HaloBox.c
steven-murray: Thanks for clarifying these!

src/py21cmfast/src/HaloBox.c
steven-murray: Yes, I guess that is the correct criterion -- that the change in observables should be small given a change in the sigma(s). Thanks for adding some docs to `inputs.py` to clarify it.

src/py21cmfast/src/HaloBox.c
steven-murray: thanks!

src/py21cmfast/src/HaloBox.c
steven-murray: That sounds like the right approach

src/py21cmfast/src/Stochasticity.c
steven-murray: That sounds even better, thanks!

src/py21cmfast/src/HaloBox.c
daviesje: After talking with Andrei I've gone with your option (no -sigma^2, F_STAR10 is the mean of the log), the main reason being that all the observational scaling relation fits are done in log-log space anyway.

src/py21cmfast/src/HaloBox.c
daviesje: Honestly I think it would hurt readability a little, and it would probably compile to the same thing anyway.

steven-murray: OK, fair enough.

src/py21cmfast/src/UsefulFunctions.c
daviesje: Apart from testing, what is the point of these functions? After all, we have much better parameter printing in the wrapper.

steven-murray: I think the original idea was to be able to print the structs from C-code for debugging -- checking that the parameters all got passed through to C OK from the python. Since adding them we've added better logging in the python, so maybe it's not so important anymore, but I can imagine situations in which you want to debug this and then you'd have to rewrite the function if we remove it now.

steven-murray (approved):
All looks good to me, thanks @daviesje. It would be good to have a couple of tests in there to make sure things aren't wonky, because I don't think I would have picked up all the bugs if any exist.

daviesje:
I've written a quick test which just compares the outputs of the scaling relations to the power-laws we expect over a few halo masses. In the meantime I was working on #404 and it made sense to combine the changes now (and it also avoided the need to expose the lyman werner turnover function or rewrite it in the test). A summary of these changes is below:

- Add a floor to the turnover mass for both MCG and ACG, controlled by the existing parameter `M_TURN`, which had the same function before but only worked when minihalos were switched off
- Add the option to simply extend the ACG scaling relations to MCG by setting the MCG parameters to `None`, this works by having conditional defaults. The defaults have been changed such that the SHMR is the same for both ACG and MCG. Due to our treatment of MCGs as 100% PopIII stars I have left the escape fraction defaults the same until we examine this model.
- Fixed a few bugs in the minihalo case and scaling relations

src/py21cmfast/src/HaloBox.c
```diff
@@ -830,20 +818,20 @@ int ComputeHaloBox(double redshift, struct UserParams *user_params, struct Cosmo
...
+            // //NOTE: in the default mode, global averages are fixed separately for minihalo and regular parts
+            // #pragma omp parallel for num_threads(user_params->N_THREADS)
+            // for(idx=0;idx<HII_TOT_NUM_PIXELS;idx++){
+            //     grids->halo_mass[idx] *= averages_global.halo_mass/averages_box.halo_mass;
+            //     grids->halo_stars[idx] *= averages_global.stellar_mass/averages_box.stellar_mass;
+            //     grids->halo_stars_mini[idx] *= averages_global.stellar_mass_mini/averages_box.stellar_mass_mini;
+            //     grids->halo_sfr[idx] *= averages_global.halo_sfr/averages_box.halo_sfr;
+            //     grids->halo_sfr_mini[idx] *= averages_global.sfr_mini/averages_box.sfr_mini;
+            //     grids->halo_xray[idx] *= averages_global.halo_xray/averages_box.halo_xray;
+            //     grids->n_ion[idx] *= averages_global.n_ion/averages_box.n_ion;
```
steven-murray, line 832: So, we're not doing the mean adjustment any more? If so we should just delete these lines.

steven-murray (approved):
OK, I think I'm convinced that things are correct :-) thanks @daviesje

src/py21cmfast/src/HaloBox.c
daviesje: I really don't like the mean adjustments, and I've removed *some* need for it by including the Sheth-Tormen CMF a while back. This made the testing a lot easier but I'm doing some comparison runs now with/without

PR #409, opened 2024-07-24
remove unneeded param structures before validation

daviesje:
In `_setup_params` which is called for every output structure, if a certain parameter structure is not provided it will not be returned, however the default parameter structure is still passed into `validate_inputs`. This means that for structures which don't require some of the input parameters (InitialConditions and PertrubedField which don't need astro params). If the other provided structures conflict with the __defaults__ (not the actual provided parameters), then an error will be raised or unexpected behaviour will occur.

This PR makes the small change that parameter structures which are not requested from the funcion are not passed into `validate_inputs`.


steven-murray (approved):

steven-murray:
@daviesje I'm worried that this small change is breaking so many tests. What parameter is now getting passed through differently?

daviesje:
> @daviesje I'm worried that this small change is breaking so many tests. What parameter is now getting passed through differently?

Yeah this was unexpected, I'm looking into the way the parameters are setup in the integration tests, something might be fetching a default when it isn't meant to

Issue #361, opened 2024-03-27
Migration toward double precision floating point numbers


daviesje:
There are many parts of the C backend which use the single precision `float` without there being a clear reason (and in some cases converting between the two several times). While it makes sense to store the largest grids in single precision, most of the internal calculations (notably the interpolation tables) should be able to be in `double` without taking a performance hit.

We should run some performance tests replacing some of the `float` calculations with `double` throughout the code. Having a more uniform data size throughout will make development a little easier and the double precision should be a little more forgiving in terms of floating point errors that could accumulate.

Issue #359, opened 2024-03-27
C build reorganisation


daviesje:
This issue is essentially mirror of #135 for the C backend. Currently the C backend is all compiled from `GenerateICs.c` which directly includes all of the `.c` and `.h` files sequentially, and specifically exposes everything in `21cmFAST.h` and `Globals.h` to the python package. This makes it difficult to control dependencies between files, specify the scope of variables and functions, and will also make most linters fairly upset. We should reorganize the C backend to be separated into files with narrower scope (for example both `ps.c` and `UsefulFunctions.c` are way too broad at the moment), and we should define header files for each to control how they are shared amongst the backend.

We should discuss the specifics of what this will look like, but as the model grows it will become important that the codebase is easy to navigate and extend.

Issue #23, opened 2019-08-21
[Feature Req.] Automatic C linting


steven-murray:
**Is your feature request related to a problem? Please describe.**

Not a problem per se, but it would be good to have consistent C linting, as well as the Python linting.

**Describe the solution you'd like**

I have not done any C linting before, but I will look into it. The proper solution should preferably be able to run automatically, and be useable on Travis CI.

steven-murray:
It looks like https://clang.llvm.org/docs/ClangFormat.html and/or https://clang.llvm.org/extra/clang-tidy/ should be useful here. They have a number of options for style, and I'd suggest we just choose one and stick to it.

steven-murray:
I'll decide on a template and show a couple of examples before I make the change.

steven-murray:
Done

PR #417, opened 2024-08-08
C restructure

daviesje:
A draft pull request for a restructure of the C backend. #359 

### Changes:
- We no longer include all headers and code in one file `GenerateICs.c`
- Instead, each file includes only the headers it uses, and all `#include "X.c"` have been removed
- `build.cffi` has been changed to make it clear which functions and definitions are available to the python wrapper.
    - CFFI requires all functions, global variables, and type definitions directly used in python to be present in `cdef()` calls *as well* as in the `set_source()` header files.
    - The `cdef()` code *cannot* contain compiler directives, as a result new header files were created labelled `_something_wrapper.h` to contain exposed definitions. these are also included in the relevant header files for the backend so we only have to edit one file to expose a function properly.
- C files `UsefulFunctions.c` and `ps.c` have been split up into functions grouped together by their themes e.g `cosmology.c` `thermochem.c` `hmf.c` etc...
- Some unused functions have been removed

### Todos:
- Make directories for different categories of files, perhaps one for the output structures and one for lower-level functions?
- Remove the many unused variables declared throughout the code
- Stricter typing:
    - Indices for the entire box should already be `unsigned long long`, but single axis indices are sometimes `int` or `unsigned long long`.
    - Many lower-level functions still use `float`, most of these should be `double` unless they are a grid which takes up significant memory/disk space #361 
- Think about splitting up `heating_helper_progs.c` as well
- Look into another build system i.e meson
- Set up C linting #23


build_cffi.py
```diff
@@ -60,29 +66,44 @@
...
+    sources=c_files,
     include_dirs=include_dirs,
     library_dirs=library_dirs,
     libraries=libraries,
     extra_compile_args=extra_compile_args,
 )
 
-# This is the Header file
-with open(os.path.join(CLOC, "21cmFAST.h")) as f:
+# Header files containing types, globals and function prototypes
```
steven-murray, line 92: If we have a format for the wrapper files, could we just do:

```python
wrapper_files = Path(CLOC).glob("_*_wrapper.h")
for fl in wrapper_files:
    with open(fl, 'r') as f:
        ffi.cdef(f)
```

build_cffi.py
```diff
@@ -51,6 +40,23 @@
...
+# =================================================================
+# Set compilation arguments dependent on environment... a bit buggy
+# =================================================================
+extra_compile_args = ["-fopenmp", "-Wall", "--verbose", f"-DLOG_LEVEL={log_level:d}"]
+if "DEBUG" in os.environ:
+    extra_compile_args += ["-g", "-O0"]
+else:
+    extra_compile_args += ["-Ofast"]
+
+libraries = ["m", "gsl", "gslcblas", "fftw3f_omp", "fftw3f"]
```
steven-murray, line 52: Did you pull through the clang changes on master?

tests/test_c_interpolation_tables.py
```diff
@@ -885,9 +872,7 @@ def test_Nion_conditional_tables(name, R, mini, intmethod, plt):
         INHOMO_RECO=True,
         USE_TS_FLUCT=True,
     )
-    lib.Broadcast_struct_global_PS(up(), cp())
-    lib.Broadcast_struct_global_UF(up(), cp())
-    lib.Broadcast_struct_global_IT(up(), cp(), ap(), fo())
+    lib.Broadcast_struct_global_all(up(), cp(), ap(), fo())
```
steven-murray, line 875: This is not your task necessarily (certainly not in this PR) but I'd love to have a nicer lower level wrapper that just exposes these functions directly in Python style, so you could call like

````python
broadcast_structs(up, cp, ap, fo)
```

steven-murray (commented):
Couple of comments but overall this is SUPER nice. Thank you!!!

tests/test_c_interpolation_tables.py
daviesje: do you mean just adding something like the following to the wrapper? these functions are already exposed

```
def broadcast_inputstructs(up,cp,ap=None,fo=None):
    if not isinstance(up,UserParams) or not isinstance(cp,CosmoParams):
        raise ValueError
    if isinstance(ap,AstroParams) and isinstance(fo,FlagOptions):
        lib.Broadcast_struct_global_all(up(), cp(), ap(), fo())

    lib.Broadcast_struct_global_noastro(up(), cp())
```

build_cffi.py
daviesje: Not yet, I'll merge them in soon

build_cffi.py
daviesje: I'm not sure about this, I think it might be a good idea to strictly keep these wrapper file categories as input parameters, output structures and function prototypes, with the possible addition of global variables (although we would want to minimize these). If there is a use-case for exposing more than this we can change it but I think we should be keeping the entry/exit points to the backend minimal.

steven-murray: Yeah, you make a good point. For a small number like this, explicit is better than implicit.

tests/test_c_interpolation_tables.py
steven-murray: Exactly.

PR #420, opened 2024-08-13
fix: compilation for clang

steven-murray:
Fixes clang compilation on v4-prep branch.


daviesje (approved):
I actually made this change on `c-restructure` but if you need to get things working before merging that go ahead

steven-murray:
Ah, I was afraid of that. OK, let's just wait for c-restructure.

PR #421, opened 2024-08-13
fix out-of-bounds access in TF_CLASS

daviesje:
There was an out-of-bounds array access in the TF_CLASS interpolations. I don't think this was ever actually tripped but I found it during the c-restructure since we can actually get good compiler warnings now.


steven-murray (approved):
Great, thanks!

PR #422, opened 2024-08-26
Fixing bugs in HaloBox.c

daviesje:
This fixes a few bugs in the lognormal galaxy distributions which erroneously reduced the amount of scatter. I've also added the Mo & White 1996 conversion between Eulerian and Lagrangian density to use with the below-resolution halos and fixed grids. I find this produces better results near delta=-1 and delta=delta_crit.


steven-murray (approved):
This all looks good to me. I went and double-checked the formula in Mo+White and it checks out.

PR #423, opened 2024-08-29
A few small bug fixes

daviesje:
This PR fixes two small bugs:

- There was an issue where the halobox outputs were not properly saved to file in Coeval objects
    - this was partly due to the naming of fields and variables in the boxes, which rely on the class names converted to snake case. There are a couple instances of classes having specific kwarg strings that I don't think is great long-term so hopefully we can change that in the near future
- If there were no halos (for example above Z_HEAT_MAX), the mean turnover mass for minihalos was set to zero. Since this global was used for the optical depths in spin temperature, the impact of this numerically was minimal, but it's *possible* that it caused some integration errors in certain circumstances


steven-murray (approved):
THis looks good, thanks @daviesje

PR #424, opened 2024-09-03
Filter func tests

daviesje:
This PR combines the three filtering functions present in v4-prep, and makes some tests for the filter functions which:

- ensures that each filter doesn't change the mean of the box (or changes it the right amount)
- ensures that the profile produced by a single cell matches what we would expect from a filtered delta function, within some tolerance

Here are the test plots produced for each filter with R=10:
[tests.test_filtering.py--test_filters[2-10].pdf](https://github.com/user-attachments/files/16850336/tests.test_filtering.py--test_filters.2-10.pdf)
[tests.test_filtering.py--test_filters[1-10].pdf](https://github.com/user-attachments/files/16850337/tests.test_filtering.py--test_filters.1-10.pdf)
[tests.test_filtering.py--test_filters[0-10].pdf](https://github.com/user-attachments/files/16850338/tests.test_filtering.py--test_filters.0-10.pdf)
[tests.test_filtering.py--test_filters[4-10].pdf](https://github.com/user-attachments/files/16850339/tests.test_filtering.py--test_filters.4-10.pdf)
[tests.test_filtering.py--test_filters[3-10].pdf](https://github.com/user-attachments/files/16850340/tests.test_filtering.py--test_filters.3-10.pdf)

Still TODO:
- fix the first test for the exponential filter
- set a reasonable passing criteria for the second test

### DIFF
--- a/src/py21cmfast/src/filtering.c
+++ b/src/py21cmfast/src/filtering.c
@@ -17,64 +17,57 @@
 #include "indexing.h"
 #include "dft.h"
 
-void filter_box_annulus(fftwf_complex *box, int RES, float R_inner, float R_outer){
-    int n_x, n_z, n_y, dimension, midpoint;
-    float k_x, k_y, k_z, k_mag, kRinner, kRouter;
-    float f_inner, f_outer;
-
-    switch(RES) {
-        case 0:
-            dimension = user_params_global->DIM;
-            midpoint = MIDDLE;
-            break;
-        case 1:
-            dimension = user_params_global->HII_DIM;
-            midpoint = HII_MIDDLE;
-            break;
-        default:
-            LOG_ERROR("Resolution for filter functions must be 0(DIM) or 1(HII_DIM)");
-            Throw(ValueError);
-            break;
-    }
-    // loop through k-box
-
-#pragma omp parallel shared(box) private(n_x,n_y,n_z,k_x,k_y,k_z,k_mag,kRinner,kRouter,f_inner,f_outer) num_threads(user_params_global->N_THREADS)
-    {
-#pragma omp for
-        for (n_x=0; n_x<dimension; n_x++){
-            if (n_x>midpoint) {k_x =(n_x-dimension) * DELTA_K;}
-            else {k_x = n_x * DELTA_K;}
-
-            for (n_y=0; n_y<dimension; n_y++){
-                if (n_y>midpoint) {k_y =(n_y-dimension) * DELTA_K;}
-                else {k_y = n_y * DELTA_K;}
-
-                for (n_z=0; n_z<=(int)(user_params_global->NON_CUBIC_FACTOR*midpoint); n_z++){
-                    k_z = n_z * DELTA_K_PARA;
-
-                    k_mag = sqrt(k_x*k_x + k_y*k_y + k_z*k_z);
-
-                    kRinner = k_mag*R_inner;
-                    kRouter = k_mag*R_outer;
-
-                    if (kRinner > 1e-4){
-                        f_inner = 3.0/(pow(kRouter, 3) - pow(kRinner, 3)) * (sin(kRinner) - cos(kRinner)*kRinner);
-                        f_outer = 3.0/(pow(kRouter, 3) - pow(kRinner, 3)) * (sin(kRouter) - cos(kRouter)*kRouter);
-                        if(RES==1) { box[HII_C_INDEX(n_x, n_y, n_z)] *= (f_outer - f_inner); }
-                        if(RES==0) { box[C_INDEX(n_x, n_y, n_z)] *= (f_outer - f_inner); }
-                    }
-
-                }
-            }
-        } // end looping through k box
-    }
-    return;
+double real_tophat_filter(double kR){
+    if (kR < 1e-4)
+        return 1;
+    return 3.0*pow(kR, -3) * (sin(kR) - cos(kR)*kR);
 }
 
-void filter_box(fftwf_complex *box, int RES, int filter_type, float R){
-    int n_x, n_z, n_y, dimension, midpoint; //TODO: figure out why defining as ULL breaks this
-    float k_x, k_y, k_z, k_mag, kR;
+//TODO: TEST USING kR^2 INSTEAD FOR SPEED
+//  ALSO TEST ASSIGNMENT vs MULTIPLICATION
+double sharp_k_filter(double kR){
+    // equates integrated volume to the real space top-hat (9pi/2)^(-1/3)
+    if (kR*0.413566994 > 1)
+       return 0.;
+    return 1;
+}
 
+double gaussian_filter(double kR_squared){
+    return exp(-0.643*0.643*kR_squared/2.);
+}
+
+double exp_mfp_filter(double k, double R, double mfp, double R_constant, double limit){
+    double kR,kR2,f;
+
+    kR = k*R;
+    if(kR < 1e-4)
+        return limit;
+
+    kR2 = k*mfp;
+    //Davies & Furlanetto MFP-eps(r) window function
+    //The filter no longer approaches 1 at k->0, so we can't use the limit
+    f = (kR2*kR2*R + 2*mfp + R)*kR2*cos(kR);
+    f += (-kR2*kR2*mfp + kR2*kR2*R + limit + R)*sin(kR);
+    f *= R_constant;
+    f -= 2*kR2*mfp;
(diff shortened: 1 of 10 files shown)


src/py21cmfast/src/filtering.c
```diff
@@ -17,64 +17,57 @@
...
-    // loop through k-box
-
-#pragma omp parallel shared(box) private(n_x,n_y,n_z,k_x,k_y,k_z,k_mag,kRinner,kRouter,f_inner,f_outer) num_threads(user_params_global->N_THREADS)
-    {
-#pragma omp for
-        for (n_x=0; n_x<dimension; n_x++){
-            if (n_x>midpoint) {k_x =(n_x-dimension) * DELTA_K;}
-            else {k_x = n_x * DELTA_K;}
+double real_tophat_filter(double kR){
+    if (kR < 1e-4)
```
steven-murray, line 21: To get slightly higher precision, you could use the taylor expansion around zero, even to second order (i.e. 1 - x^2/10). This gives accuracy to the full extent of double-precision at kR=1e-4

src/py21cmfast/src/filtering.c
```diff
@@ -17,64 +17,57 @@
...
+    return exp(-0.643*0.643*kR_squared/2.);
+}
 
-                    k_mag = sqrt(k_x*k_x + k_y*k_y + k_z*k_z);
+double exp_mfp_filter(double k, double R, double mfp, double R_constant, double limit){
+    double kR,kR2,f;
+
+    kR = k*R;
+    if(kR < 1e-4)
+        return limit;
```
steven-murray, line 44: Using the taylor expansion again might get slightly better precision here.

tests/test_filtering.py
```diff
@@ -0,0 +1,251 @@
...
+        up(),
+        cp(),
+        ap(),
+        fo(),
+        ffi.cast("float *", input_box_centre.ctypes.data),
+        R,
+        R_param,
+        filter_flag,
+        ffi.cast("double *", output_box_centre.ctypes.data),
+    )
```
steven-murray, line 120: can we write a low-level wrapper function for this in the main code?

steven-murray: In fact, if instead we write the `filter_box` function as a low-level wrapper, we could perhaps write the `test_filter` function directly in python?

tests/test_filtering.py
```diff
@@ -0,0 +1,251 @@
...
+                binned_truth_centre,
+            ],
+            truths=[
+                exp_output_centre,
+            ],
+            r_bins=r_bins,
+            r_grid=r_from_centre,
+            slice_index=up.HII_DIM // 2,
+            slice_axis=0,
+            plt=plt,
```
steven-murray, line 167: Something weird happened with the formatting here -- leave off the trailing commas within the lists so that they stay on one line.

tests/test_filtering.py
```diff
@@ -0,0 +1,251 @@
...
+        norm_factor = 1
+    # firstly, make sure the filters are normalised
+    np.testing.assert_allclose(
+        input_box_centre.sum() * norm_factor, output_box_centre.sum(), rtol=1e-4
+    )
+
+    # then make sure we get the right shapes (more lenient for aliasing)
+    np.testing.assert_allclose(
+        output_box_centre, exp_output_centre, rtol=RELATIVE_TOLERANCE
+    )
```
steven-murray, line 187: I think using an absolute tolerance makes more sense here (at least for the normalized filters), because you have a bunch of cells that are really close to zero.

tests/test_filtering.py
```diff
@@ -0,0 +1,251 @@
...
+            yerr=[stats_o["err1l"], stats_o["err1u"]],
+            label="filtered grid",
+        )
+        axs[idx, 3].plot(r_cen, t, "m:", label="Expected")
+        axs[idx, 3].grid()
+        axs[idx, 3].set_xlabel("dist from centre")
+        axs[idx, 3].set_ylabel("cell value")
+
+        axst = axs[idx, 3].twinx()
+        axst.plot(r_cen, stats_o["mean"] / t - 1, "r-")
```
steven-murray, line 249: Instead of dividing by the truth -- which could be very very close to zero, you might want to add some epsilon to the denominator here to account for machine-precision uncertainties (e.g. add 1e-16 to t so you don't get infs)

steven-murray (commented):
Thanks @daviesje  -- I have a couple of suggestions that _might_ help, but let's see

tests/test_filtering.py
daviesje: Would the second option not require us to expose a bunch of fftw functions to the python wrapper?

steven-murray: Not necessarily -- we could just use `np.fft` or even `pyfftw` to do the FFT stuff in the wrapper function

daviesje: Wouldn't we be testing different funcitons at that point?

daviesje: If you're okay with it, I would prefer to write wrappers for all the testing functions together in a separate PR

daviesje:
So the tests now pass. I've significantly widened the tolerances on the pixel-to-pixel test, and included a radially binned test in order to mitigate some problems with sharp real-space filters. Below are all the plots from these tests

[tests.test_filtering.py--test_filters[4-1.5].pdf](https://github.com/user-attachments/files/17294841/tests.test_filtering.py--test_filters.4-1.5.pdf)
[tests.test_filtering.py--test_filters[4-5].pdf](https://github.com/user-attachments/files/17294842/tests.test_filtering.py--test_filters.4-5.pdf)
[tests.test_filtering.py--test_filters[4-10].pdf](https://github.com/user-attachments/files/17294843/tests.test_filtering.py--test_filters.4-10.pdf)
[tests.test_filtering.py--test_filters[4-20].pdf](https://github.com/user-attachments/files/17294844/tests.test_filtering.py--test_filters.4-20.pdf)
[tests.test_filtering.py--test_filters[0-1.5].pdf](https://github.com/user-attachments/files/17294845/tests.test_filtering.py--test_filters.0-1.5.pdf)
[tests.test_filtering.py--test_filters[0-5].pdf](https://github.com/user-attachments/files/17294846/tests.test_filtering.py--test_filters.0-5.pdf)
[tests.test_filtering.py--test_filters[0-10].pdf](https://github.com/user-attachments/files/17294847/tests.test_filtering.py--test_filters.0-10.pdf)
[tests.test_filtering.py--test_filters[0-20].pdf](https://github.com/user-attachments/files/17294848/tests.test_filtering.py--test_filters.0-20.pdf)
[tests.test_filtering.py--test_filters[1-1.5].pdf](https://github.com/user-attachments/files/17294849/tests.test_filtering.py--test_filters.1-1.5.pdf)
[tests.test_filtering.py--test_filters[1-5].pdf](https://github.com/user-attachments/files/17294851/tests.test_filtering.py--test_filters.1-5.pdf)
[tests.test_filtering.py--test_filters[1-10].pdf](https://github.com/user-attachments/files/17294852/tests.test_filtering.py--test_filters.1-10.pdf)
[tests.test_filtering.py--test_filters[1-20].pdf](https://github.com/user-attachments/files/17294853/tests.test_filtering.py--test_filters.1-20.pdf)
[tests.test_filtering.py--test_filters[2-1.5].pdf](https://github.com/user-attachments/files/17294854/tests.test_filtering.py--test_filters.2-1.5.pdf)
[tests.test_filtering.py--test_filters[2-5].pdf](https://github.com/user-attachments/files/17294855/tests.test_filtering.py--test_filters.2-5.pdf)
[tests.test_filtering.py--test_filters[2-10].pdf](https://github.com/user-attachments/files/17294856/tests.test_filtering.py--test_filters.2-10.pdf)
[tests.test_filtering.py--test_filters[2-20].pdf](https://github.com/user-attachments/files/17294857/tests.test_filtering.py--test_filters.2-20.pdf)
[tests.test_filtering.py--test_filters[3-1.5].pdf](https://github.com/user-attachments/files/17294858/tests.test_filtering.py--test_filters.3-1.5.pdf)
[tests.test_filtering.py--test_filters[3-5].pdf](https://github.com/user-attachments/files/17294859/tests.test_filtering.py--test_filters.3-5.pdf)
[tests.test_filtering.py--test_filters[3-10].pdf](https://github.com/user-attachments/files/17294860/tests.test_filtering.py--test_filters.3-10.pdf)
[tests.test_filtering.py--test_filters[3-20].pdf](https://github.com/user-attachments/files/17294861/tests.test_filtering.py--test_filters.3-20.pdf)

tests/test_filtering.py
steven-murray: That's fine with me

PR #425, opened 2024-09-10
Reading compatibility

daviesje:
There is currently a behaviour in `global_params.use()` which, unlike other parameter structures, raises an error when a keyword not in the parameter set is passed. This made it impossible to read in boxes from different versions that had differing parameters. I've made it behave like the other parameter sets now, which simply send out a warning.

However, to keep things safer, I've added a flag to the high-level output reading functions which ensure that all the parameters are there, if the flag is on and the parameters are mismatched it raises an error. if the flag is off it emits a warning detailing which parameters are missing / extra.

Tests have also been written to ensure the desired behaviour.

### DIFF
--- a/src/py21cmfast/outputs.py
+++ b/src/py21cmfast/outputs.py
@@ -983,6 +983,10 @@ class _HighLevelOutput:
             If fname, is relative, the directory in which to find the file. By default,
             both the current directory and default cache and the  will be searched, in
             that order.
+        safe : bool
+            If safe is true, we throw an error if the parameter structures in the file do not
+            match the structures in the `inputs.py` module. If false, we allow extra and missing
+            items, setting the missing items to the default values and ignoring extra items.
 
         Returns
         -------
--- a/src/py21cmfast/outputs.py
+++ b/src/py21cmfast/outputs.py
@@ -1294,11 +1294,11 @@ class LightCone(_HighLevelOutput):
                 ("astro_params", AstroParams),
             ]:
                 dct = dict(fl[k].attrs)
-                if kls.__defaults__.keys() != dct.keys():
+                if kls._defaults_.keys() != dct.keys():
                     message = (
                         f"There are extra or missing global params in the file to be read.\n"
-                        f"EXTRAS: {[(k,v) for k,v in dct.items() if k not in kls.__defaults__.keys()]}\n"
-                        f"MISSING: {[(k,v) for k,v in kls.__defaults__.items() if k not in dct.keys()]}\n"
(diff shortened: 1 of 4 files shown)


tests/test_high_level_io.py
```diff
@@ -56,6 +57,81 @@ def ang_lightcone(ic, lc):
     )
```
daviesje, line 59: Ideally I would also test for gaps and gluts in the data structures themselves, but the different way coevals and lightcones are structured make this difficult.

Lightcones are agnostic to the types of the lightcone fields, so missing fields are ignored and extra ones are simply loaded in.

Coevals on the other hand rely on the output structs, so both missing and extra output fields will raise errors, and changing this behaviour requires adding `safe` flags to all the lower level reading as well.

src/py21cmfast/outputs.py
```diff
@@ -936,10 +936,23 @@ def save(self, fname=None, direc=".", clobber: bool = False):
         return self._write(direc=direc, fname=fname, clobber=clobber)
 
     @classmethod
```
daviesje, line 938: I had a question here. Where does the Coeval object read in the parameter structures? The Coeval `__init__` takes them directly from the `BrightnessTemp` struct, but these don't seem to be saved with the structures looking at the `_write_particulars` function.

tests/test_high_level_io.py
```diff
@@ -56,6 +57,81 @@ def ang_lightcone(ic, lc):
...
+    with pytest.warns(UserWarning, match="There are extra or missing"):
+        lc2 = LightCone.read(fname, direc=test_direc, safe=False)
+
+    # check that the fake fields didn't show up in the struct
+    assert not hasattr(lc2.user_params, "NotARealParameter")
+    assert "NotARealGlobal" not in lc2.global_params.keys()
+
+    # check that missing fields are set to default
+    assert lc2.user_params.BOX_LEN == UserParams._defaults_["BOX_LEN"]
+    assert lc2.global_params["OPTIMIZE_MIN_MASS"] == global_params.OPTIMIZE_MIN_MASS
```
daviesje, line 88: This assert would be more useful if I changed the global_params in the lc/coeval fixtures

tests/test_high_level_io.py
```diff
@@ -56,6 +57,81 @@ def ang_lightcone(ic, lc):
...
+
+    # check that the fields which are good are read in the struct
+    lc2.user_params.BOX_LEN = lc.user_params.BOX_LEN
```
daviesje, line 91: Is there a nicer way to compare all but one field?

steven-murray (approved):

PR #427, opened 2024-09-16
Ionbox refactor

daviesje:
This is a coarse modularisation of the ionised box calculation which separates the large function into a number of smaller steps. This should make future development easier. I have tested that we get the same results under a few variations of parameters, and it has not slowed down.

Key differences:

 - constants are calculated at the beginning and held in a struct which is passed into each function.
 - Radii are explicitly calculated at the beginning of the run, instead of on-the-fly in a while loop
 - the RNG for partial ionisations has been explicitly removed (it only affected the !USE_MASS_DEPENDENT_ZETA case)

Further modularisation is possible, but best left to another PR


src/py21cmfast/src/IonisationBox.c
```diff
@@ -27,1477 +27,1407 @@
...
-                }
-                if(flag_options->USE_MINI_HALOS){
-                    prev_deltax_unfiltered[ct]          /= (HII_TOT_NUM_PIXELS+0.0);
-                    log10_Mturnover_unfiltered[ct]      /= (HII_TOT_NUM_PIXELS+0.0);
-                    log10_Mturnover_MINI_unfiltered[ct] /= (HII_TOT_NUM_PIXELS+0.0);
-                }
+            #pragma omp for
+            for (ct=0; ct<HII_TOT_NUM_PIXELS; ct++) {
+                box->Gamma12_box[ct] = 0.0;
+                box->MFP_box[ct] = 0.0;
```
steven-murray, line 1211: Yes, this is all wrapper stuff.

src/py21cmfast/src/IonisationBox.c
```diff
@@ -27,1477 +27,1407 @@
...
-            if ( ((R/(global_params.DELTA_R_HII_FACTOR) - cell_length_factor*(user_params->BOX_LEN)/(float)(user_params->HII_DIM)) <= FRACT_FLOAT_ERR) || \
-                    ((R/(global_params.DELTA_R_HII_FACTOR) - global_params.R_BUBBLE_MIN) <= FRACT_FLOAT_ERR) ) {
-                LAST_FILTER_STEP = 1;
-                R = fmax(cell_length_factor*user_params->BOX_LEN/(double)(user_params->HII_DIM), global_params.R_BUBBLE_MIN);
-            }
+    // perform a very rudimentary check to see if we are underresolved and not using the linear approx
+    if ((user_params->BOX_LEN > user_params->DIM) && !(global_params.EVOLVE_DENSITY_LINEARLY)){
+        LOG_WARNING("Resolution is likely too low for accurate evolved density fields\n It Is recommended \
+                    that you either increase the resolution (DIM/BOX_LEN) or set the EVOLVE_DENSITY_LINEARLY flag to 1\n");
+    }
```
steven-murray, line 1253: I feel like this also should be in the wrapper.

steven-murray (commented):
Nice work @daviesje, this is going to make things much easier!

src/py21cmfast/src/Stochasticity.c
```diff
@@ -299,6 +299,46 @@ int stoc_halo_sample(struct HaloSamplingConstants *hs_constants, gsl_rng * rng,
     return 0;
 }
 
+//same as stoc_halo_sample but with a mass tolerance for comparison with Ivan
+int stoc_halo_sample_tol(struct HaloSamplingConstants *hs_constants, gsl_rng * rng, int *n_halo_out, float *M_out){
```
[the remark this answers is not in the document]
steven-murray: Thanks!

steven-murray (approved):
Looks great, thanks @daviesje !

steven-murray:
@daviesje most recent changes look good -- I didn't understand the switch from `true` to `false` but I assume it was for a good reason. The increase in the halo buffer seems perhaps a little arbitrary, though not really a problem. 

The two failing tests with `nomdz` -- these segfault here but pass locally? Can you run with valgrind or something?

daviesje:
> @daviesje most recent changes look good -- I didn't understand the switch from `true` to `false` but I assume it was for a good reason.
This was a bug introduced in the refactor where the wrong interpolation tables were being initialised (the previous redshift instead of the current).
> The increase in the halo buffer seems perhaps a little arbitrary, though not really a problem.
This is entirely arbitrary, but doesn't effect the test results. I just need to make sure that it can hold all the halos that the tests produce, I thought about 500MB should be enough.
> The two failing tests with `nomdz` -- these segfault here but pass locally? Can you run with valgrind or something?
This has just been fixed, it was an error on my part incorrectly setting interpolation table limits when USE_MASS_DEPENDENT_ZETA=False. For some reason running tests with `-v` made it not segfault, although the results would surely have been garbage

steven-murray (approved):

PR #430, opened 2024-09-18
add a minimum z to lightcone_sliceplot

daviesje:
Small change allowing `plotting.lightcone_sliceplot` to have a minimum redshift without messing with the lightcone object itself. For when you want to only plot a subset of cosmic history.


steven-murray (approved):
Great! Thanks!

PR #431, opened 2024-09-19
change default behaviour of from_file and read methods

daviesje:
File loading keywords were confusing and the docstrings were incorrect.

I've changed the default reading behaviour to not load any arrays, and the keyword `arrays_to_load` or `keys` for `from_file` or `read` respectively works as follows:

Passing empty array `[]`: load nothing
Passing `None`: load everything
Passing a list of array names: load that subset of arrays

This is _technically_ API breaking so I'm placing this PR in v4-prep

### DIFF
--- a/src/py21cmfast/_utils.py
+++ b/src/py21cmfast/_utils.py
@@ -1135,7 +1135,7 @@ class OutputStruct(StructWrapper, metaclass=ABCMeta):
         self,
         direc: Union[str, Path, None] = None,
         fname: Union[str, Path, None, h5py.File, h5py.Group] = None,
-        keys: Optional[Sequence[str]] = [],
+        keys: Optional[Sequence[str]] = (),
     ):
         """
         Try find and read existing boxes from cache, which match the parameters of this instance.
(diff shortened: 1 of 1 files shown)


src/py21cmfast/_utils.py
```diff
@@ -1135,7 +1135,7 @@ def read(
         self,
         direc: Union[str, Path, None] = None,
         fname: Union[str, Path, None, h5py.File, h5py.Group] = None,
-        keys: Optional[Sequence[str]] = None,
+        keys: Optional[Sequence[str]] = [],
```
steven-murray, line 1138: Ah, you should not set the default as a mutable object -- make it instead an empty tuple.

src/py21cmfast/_utils.py
```diff
@@ -1214,9 +1216,8 @@ def from_file(
         cls,
         fname,
         direc=None,
-        load_data=True,
         h5_group: Union[str, None] = None,
-        arrays_to_load=None,
+        arrays_to_load=[],
```
steven-murray, line 1220: same here

steven-murray (commented):
just two little things to update

steven-murray (approved):

PR #433, opened 2024-10-04
small changes for the upcoming database

daviesje:
A few changes mainly for the upcoming database:

- Setting a constant mfp instead of the power-law fit, which resulted in kinks in reionisation history at z=6
- Using the filtered density instead of local when using the exponential filter, as done in (Davies+22)
    - There are some questions here about what our model should be with this filter + recombinations but this choice seems better for now
- For unresolved halos in the sampler, use the Lagrangian density instead of Eulerian to be more consistent with the halo sampler. In future we also want to move these as particles with the 2LPT but that will take some time to develop so for now we assume they don't move.

### DIFF
--- a/src/py21cmfast/src/HaloBox.c
+++ b/src/py21cmfast/src/HaloBox.c
@@ -83,7 +83,7 @@ struct HaloProperties{
 
 void set_hbox_constants(double redshift, AstroParams *astro_params, FlagOptions *flag_options, struct HaloBoxConstants *consts){
     consts->redshift = redshift;
-    
+
     //whether to fix *integrated* (not sampled) galaxy properties to the expected mean
     //Set on for the fixed grid case since we are missing halos above the cell mass
     consts->fix_mean = flag_options->FIXED_HALO_GRIDS;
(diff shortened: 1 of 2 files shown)


src/py21cmfast/src/HaloBox.c
```diff
@@ -85,7 +85,7 @@ void set_hbox_constants(double redshift, AstroParams *astro_params, FlagOptions
     consts->redshift = redshift;
     //whether to fix *integrated* (not sampled) galaxy properties to the expected mean
     //  constant for now, to be a flag later
```
steven-murray, line 87: can remove this comment now

src/py21cmfast/src/IonisationBox.c
```diff
@@ -1216,12 +1218,14 @@ LOG_SUPER_DEBUG("excursion set normalisation, mean_f_coll_MINI: %e", box->mean_f
...
-                                curr_dens = perturbed_field->density[HII_R_INDEX(x,y,z)];
-                            else
-                                curr_dens = *((float *)deltax_filtered + HII_R_FFT_INDEX(x,y,z));
+                            //TODO: There may be a reason to use unfiltered density for CELL_RECOMB case,
+                            // since the "Fcoll" represents photons reaching the central cell rather than
+                            // photons in the entire sphere, but it makes reionisation look stranger so
+                            // more investigation is needed
+                            // if(flag_options->CELL_RECOMB)
+                            //     curr_dens = perturbed_field->density[HII_R_INDEX(x,y,z)];
+                            // else
```
steven-murray, line 1227: I think we need to make an issue about this so we don't lose it

steven-murray (approved):
Conditional approval pending two small comments.

PR #436, opened 2024-10-14
don't save HaloField by default in high level functions

daviesje:
Since halo fields can get quite large, and since the classes `HaloFIeld` and `PerturbHaloField` hold similar information, we really only want to cache one most of the time. So I've changed the default behaviour in `run_coeval` and `run_lightcone` to only write `PerturbHaloField`. This could save up to ~10% time and halve the cache requirements.

- `_get_config_options` is now only called on the lower-level structures, this was done to have more control over writing behaviour.
- If `hooks` are passed or `write` is set to true or false, the behaviour is the same as before
- if `write==None` (default) we write everything except for `HaloField`.

This also fixes an unrelated problem where we overpurged the ICs for the halofield since we need the velocity fields for `perturb_halo_list`.

### DIFF
--- a/src/py21cmfast/outputs.py
+++ b/src/py21cmfast/outputs.py
@@ -129,15 +129,6 @@ class InitialConditions(_OutputStruct):
 
         self.prepare(keep=keep, force=force)
 
-    def prepare_for_halos(self, flag_options: FlagOptions, force: bool = False):
-        """Ensure ICs have all boxes required for the halos, and no more."""
-        keep = ["hires_density"]  # for dexm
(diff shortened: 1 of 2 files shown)


daviesje:
> This looks good, thanks @daviesje. BTW does this fix the "writing when it's not supposed to" issues we've been having, by moving the config into the function instead of as the default?

I'm not sure, if `hooks==None` and `write==False` the first config call will set `hooks == {}`and pass in the default `write == None` since that keyword isn't in `iokw`. After that, since `len({}) == 0`, it shouldn't fill the hooks with the default write.

PR #438, opened 2024-11-05
Fixing Tests

daviesje:
This PR will achieve two things:
- Moves most of the C backend access in the tests to wrapper/cfuncs.py to make them more intuitive
- changes the tolerances of tests to reflect what we actually expect from our functions. this mostly involves restricting the comparisons between interpolation tables and integrals to domains we actually care about, since most of these tests had <10^-3 errors in all bins except one near the critical collapse density, where it had 10^-1 error.

As such, tests (apart from integration tests of course) should all be passing.

### DIFF
--- a/src/py21cmfast/wrapper/cfuncs.py
+++ b/src/py21cmfast/wrapper/cfuncs.py
@@ -373,7 +373,14 @@ def construct_fftw_wisdoms(
 def get_delta_crit(user_params, cosmo_params, mass, redshift):
     """Gets the critical collapse density given a mass, redshift and parameters."""
     sigma, _ = evaluate_sigma(user_params, cosmo_params, mass)
+    # evaluate_sigma already broadcasts the paramters so we don't need to repeat
     growth = lib.dicke(redshift)
+    return get_delta_crit_nu(user_params, sigma, growth)
+
+
+def get_delta_crit_nu(user_params, sigma, growth):
+    """Uses the nu paramters (sigma and growth factor) to get critical density."""
+    # None of the parameter structs are used in this function so we don't need a broadcast
     return np.vectorize(lib.get_delta_crit)(user_params.cdict["HMF"], sigma, growth)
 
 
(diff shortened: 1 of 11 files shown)


steven-murray (commented):
Thanks @daviesje, this is looking good. I have one thing to consider: currently you're putting these interpolation table functions into `cfuncs.py`. I was intending `cfuncs.py` to be very low-level wrappers around functions in the C (so, essentially just wrappers that have the same name as the C-function, with things like `np.vectorize` applied, and any setup required performed, like broadcasting global structs). 

It might be useful to keep `cfuncs.py` as this very low-level thing, and then have a module `interpolation_tables.py` which implements the slightly higher-level functions you've written here.

daviesje:
If you're happy for the cleanup of `cfuncs.py` to be in a quick followup PR, we can push a version of this branch with passing tests. Especially since I think that would require a bit more conversation.

Also, the test_cache_exists is passing locally so I'm not sure what's happening there

src/py21cmfast/wrapper/cfuncs.py
```diff
@@ -370,6 +370,13 @@ def construct_fftw_wisdoms(
 #  They are mostly used for testing but may be useful in some post-processing applications
 
 
+def get_delta_crit(user_params, cosmo_params, mass, redshift):
+    """Gets the critical collapse density given a mass, redshift and parameters."""
+    sigma, _ = evaluate_sigma(user_params, cosmo_params, mass)
+    growth = lib.dicke(redshift)
+    return np.vectorize(lib.get_delta_crit)(user_params.cdict["HMF"], sigma, growth)
```
steven-murray, line 377: It might be useful to have a version of `get_delta_crit` to which you just pass `sigma` directly (instead of mass).

steven-murray (approved):
Thanks @daviesje -- I've approved, though with one small comment that you can take or leave

PR #439, opened 2024-11-07
run lc tests with correct params, stop purging ics for halos

daviesje:
I had previously included a third purging method of the ICs, `prepare_for_halos`. However since I needed the perturbing fields for the halos anyway, this was over-purging the arrays, resulting in unnecessary i/o.

This also includes a small fix to one of the tests which was taking longer than it should by using the wrong flag options.


steven-murray (approved):
LGTM!

PR #440, opened 2024-11-09
Fix some bugs from most recent PR

daviesje:
This fixes two bugs which slipped through the tests when we removed recursion

- the hacky `XraySourceBox` construction has been changed to work with the new cached properties in the structs.
- `HaloField` required arrays have been updated

### DIFF
--- a/src/py21cmfast/drivers/single_field.py
+++ b/src/py21cmfast/drivers/single_field.py
@@ -576,7 +576,8 @@ def interp_halo_boxes(
     hbox_out = HaloBox(
         inputs=inputs,
     )
-    # initialise hbox memory
+
+    # initialise the memory
     hbox_out()
 
     # interpolate halo boxes in gridded SFR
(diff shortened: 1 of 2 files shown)


src/py21cmfast/drivers/single_field.py
```diff
@@ -587,20 +588,24 @@ def interp_halo_boxes(
...
+        + f" interp {getattr(hbox_out, fields[0]).mean()}"
+        + f" prog ({idx_prog}) {getattr(hbox_prog, fields[0]).mean()}"
     )
 
-    # pass the arrays to the c struct by calling again
-    hbox_out()
+    # HACK: this passes the field pointers to the backend,
+    # NOTE: the arrays are initialised in the call above so they shouldn't
+    # be re-initialised causing a memory leak
+    hbox_out._init_cstruct()
```
steven-murray, line 608: @daviesje I think I need more of an explanation of what's going on here :-) Maybe pop in when you get to the office

steven-murray (approved):
Thanks @daviesje !

PR #441, opened 2024-11-13
Fixing mac build and clang compilation on v4-prep

daviesje:


daviesje:
By moving the global validation to its `use()` method, the Mock module for the docs doesn't complain anymore, and I think it's probably better to validate there.

It is still *technically* possible to pass an invalid global struct, if the user alters the `global_params` object directly and then calls a backend function *without* going through one of the wrapper functions (since they all have the `@set_globals` decorator).

.github/workflows/test_build_macos.yaml
```diff
@@ -60,11 +61,11 @@ jobs:
...
-      - name: Setup GCC
-        if: matrix.compiler == 'gcc'
-        uses: Dup4/actions-setup-gcc@v1
-        with:
-          version: latest
+      # - name: Setup GCC
+      #   if: matrix.compiler == 'gcc'
+      #   uses: Dup4/actions-setup-gcc@v1
+      #   with:
+      #     version: latest
```
steven-murray, line 68: Let's just remove the dead code entirely.

.github/workflows/test_suite.yaml
```diff
@@ -79,10 +79,10 @@ jobs:
...
           sudo apt-get install libgsl0-dev
 
-      - name: Setup GCC
-        uses: Dup4/actions-setup-gcc@v1
-        with:
-          version: latest
+      # - name: Setup GCC
+      #   uses: Dup4/actions-setup-gcc@v1
+      #   with:
+      #     version: latest
```
steven-murray, line 85: and here

.github/workflows/test_suite_nointegration.yaml
```diff
@@ -78,10 +78,10 @@ jobs:
...
           sudo apt-get install libgsl0-dev
 
-      - name: Setup GCC
-        uses: Dup4/actions-setup-gcc@v1
-        with:
-          version: latest
+      # - name: Setup GCC
+      #   uses: Dup4/actions-setup-gcc@v1
+      #   with:
+      #     version: latest
```
steven-murray, line 84: and here

steven-murray (commented):
Thanks @daviesje  -- I think it's totally fine that it's *technically* possible to break things. Calling backend C-code without using a wrapper is totally outside the scope of the API, and should only be undertaken by developers, if ever, so this is good by me.

Just remove that dead code, and LGTM.

PR #442, opened 2024-11-14
fix trapezoidal bug, make error messages more clear

daviesje:
I accidentally broke the trapezoidal integrals in (z,Mturn) in the Ionisation box refactor (USE_HALO_FIELD=False, USE_MINI_HALOS=True). This fixes the issue and makes the error messages regarding those integrals clearer


steven-murray (approved):
This looks good -- thanks for adding more helpful errors!

PR #444, opened 2024-11-22
Run templates

daviesje:
Since we have a lot of parameters, it would help to provide a set of run templates which correspond to previous papers and common running modes of 21cmfast. To do this I have implemented TOML template files, along with functions which generate an `InputParameters` instance, which is now passed into `run_coeval` and `run_lightcone` instead of the four `InputStruct` instances


steven-murray (commented):
Thanks @daviesje, I really like this implementation! Just a couple of simple comments.

src/py21cmfast/run_templates.py
```diff
@@ -0,0 +1,96 @@
...
+
+    Parameters
+    ----------
+    template_name: str,
+        defines the name/alias of the native template (see templates/manifest.toml for a list)
+        alternatively, is the path to a TOML file containing tables titled [CosmoParams],
+        [UserParams], [AstroParams] and [FlagOptions] with parameter settings
+
+    kwargs,
+        names and values of parameters in the ``InputStruct`` subclasses to override the template values with.
```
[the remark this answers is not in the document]
daviesje: also docstrings for all the drivers need to be updated

src/py21cmfast/run_templates.py
```diff
@@ -0,0 +1,96 @@
...
+    )
+    logger.error("Available native templates are:")
+    for manf_entry in manifest:
+        logger.error(f" {manf_entry['name']}: {manf_entry['aliases']}")
+        logger.error(f"     {manf_entry['description']}:")
+        logger.error("")
+
+    raise ValueError(
+        f"template {template_name} not found on-disk or in native template aliases"
+    )
```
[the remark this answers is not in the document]
daviesje: done

src/py21cmfast/drivers/param_config.py
```diff
@@ -207,6 +223,49 @@ def from_output_structs(
...
+            cosmo_params=cosmo_params,
+            user_params=user_params,
+            astro_params=astro_params,
+            flag_options=flag_options,
+        )._add_seed(seed)
+
+    @classmethod
+    def from_defaults(cls, seed=None):
+        """Construct full InputParameters instance from default values."""
+        # TODO: add kwargs to modify values e.g run_templates._construct_param_objects?
```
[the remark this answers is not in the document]
daviesje: done

src/py21cmfast/drivers/lightcone.py
```diff
@@ -883,13 +872,20 @@ def run_lightcone(
...
     # For the high-level, we need all the InputStruct initialised
-    if cosmo_params is None and initial_conditions is None:
-        cosmo_params = CosmoParams.from_astropy(lightconer.cosmo)
+    # NOTE: the random seed is now set here instead of in the OutputStruct init
+    #   if it's passed as None
+    # TODO: make sure cosmo_params is consistent with lightconer.cosmo
+    if isinstance(inputs, str):
+        inputs = InputParameters.from_template(inputs, seed=random_seed)
+    elif inputs is None:
+        inputs = InputParameters.from_defaults(seed=random_seed)
```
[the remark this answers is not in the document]
daviesje: done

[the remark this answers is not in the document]
daviesje: We should update the tutorials as well

src/py21cmfast/drivers/coeval.py
```diff
@@ -605,44 +598,41 @@ def run_coeval(
...
-    cosmo_params = inputs.cosmo_params
-    user_params = inputs.user_params
-    flag_options = inputs.astro_params
-    astro_params = inputs.flag_options
-
-    random_seed = inputs.random_seed
+    # if no random seed was given to the inputs, and no output structs were passed,
+    #   raise an error
+    if not isinstance(inputs.random_seed, int):
+        raise ValueError("An integer seed, or initial conditions must be given")
```
steven-murray, line 619: Let's instead just make `random_seed` a required property on the `InputParameters` class.

src/py21cmfast/drivers/coeval.py
```diff
@@ -605,44 +598,41 @@ def run_coeval(
...
-        if initial_conditions is not None
-        else random_seed
+    # ensure inputs are compatible with ICs/Perturbedfields
+    inputs = InputParameters.from_output_structs(
+        (initial_conditions,) + perturbed_field,
+        cosmo_params=inputs.cosmo_params,
+        user_params=inputs.user_params,
+        astro_params=inputs.astro_params,
+        flag_options=inputs.flag_options,
+        random_seed=inputs.random_seed,
```
steven-murray, line 613: I reckon we should just write a new method on `InputParameters` that does the compatibility check (even if it itself just wraps the `.from_output_structs`). Something with a better name, like `inputs.check_compatibility(outputs)`.

src/py21cmfast/drivers/coeval.py
```diff
@@ -833,17 +823,18 @@ def run_coeval(
             st = st2
 
         perturb_files.append((z, os.path.join(direc, pf2.filename)))
-        if flag_options.USE_HALO_FIELD:
+        if inputs.flag_options.USE_HALO_FIELD:
             hbox_files.append((z, os.path.join(direc, hb2.filename)))
-            pth_files.append((z, os.path.join(direc, ph2.filename)))
-        if flag_options.USE_TS_FLUCT:
+            if not inputs.flag_options.FIXED_HALO_GRIDS:
+                pth_files.append((z, os.path.join(direc, ph2.filename)))
```
steven-murray, line 829: nitpicky, but I don't love `pth_files` as a name -- I associate `pth` with "path" not perturbed halos.

src/py21cmfast/drivers/lightcone.py
```diff
@@ -860,32 +843,31 @@ def run_lightcone(
...
+        user_params=inputs.user_params,
+        astro_params=inputs.astro_params,
+        flag_options=inputs.flag_options,
+        random_seed=inputs.random_seed,
+    )
+
+    # if no random seed was given to the inputs, and no output structs were passed,
+    #   raise an error
+    if not isinstance(inputs.random_seed, int):
+        raise ValueError("An integer seed, or initial conditions must be given")
```
steven-murray, line 870: this would be removed also if we move the random seed validation to the input parameters class.

src/py21cmfast/wrapper/structs.py
```diff
@@ -677,7 +660,7 @@ def write(
                 f"Non-computed boxes: {[k for k, v in self._array_state.items() if not v.computed]}"
             )
 
-        if not self._random_seed:
+        if not self.random_seed:
```
steven-murray, line 663: this would also be removed if we have validation in the `InputParameters`

src/py21cmfast/wrapper/structs.py
```diff
@@ -960,7 +943,7 @@ def __repr__(self):
         # This is the class name and all parameters which belong to C-based input structs,
         # eg. InitialConditions(HII_DIM:100,SIGMA_8:0.8,...)
         # eg. InitialConditions(HII_DIM:100,SIGMA_8:0.8,...)
-        return f"{self._seedless_repr()}_random_seed={self._random_seed}"
+        return f"{self._seedless_repr()}_random_seed={self.random_seed}"
```
steven-murray, line 946: I think this too, but maybe better if I deal with this in my own PR

steven-murray (commented):
Almost there!

src/py21cmfast/drivers/param_config.py
```diff
@@ -55,18 +78,48 @@ class InputParameters:
...
+        converter=float, default=global_params.Z_HEAT_MAX, eq=False, repr=False
+    )
+    _redshift_step = attrs.field(
+        converter=float, default=global_params.ZPRIME_STEP_FACTOR, eq=False, repr=False
+    )
+
+    # passed to the converter, TODO: this can be cleaned up
+    node_redshifts = attrs.field(
+        converter=attrs.Converter(_node_redshifts_converter, takes_self=True)
+    )
```
steven-murray, line 100: I think this is OK. It's a little convoluted, but does everything we need it to. I think this would be cleaner again if we were to remove Z_HEAT_MAX and ZPRIME_STEP_FACTOR as global variables. In that case, we'd just let the user construct the array every time (and provide a convenience method that they could call to do that easily).

src/py21cmfast/drivers/param_config.py
```diff
@@ -172,63 +209,124 @@ def is_compatible_with(self, other: InputParameters) -> bool:
...
+            if len(values) > 1:
+                raise ValueError(
+                    f"InputParameters.from_output_struct got multiple values for {field}: {values}"
+                )
+            elif len(values) == 0:
+                if attrs.fields_dict(cls)[field].default:
+                    # If the parameter has a default, we want to remove them
+                    #    from the dict (after the loop)
+                    # TODO: this is messy, try to clean it up
+                    default_fields.append(field)
```
steven-murray, line 255: why not just do `del input_params[field]` here, and then remove the line below that does the popping in the list comprension?

src/py21cmfast/drivers/param_config.py
```diff
@@ -172,63 +209,124 @@ def is_compatible_with(self, other: InputParameters) -> bool:
...
+            output_structs,
+            node_redshifts=self.node_redshifts,
+            cosmo_params=self.cosmo_params,
+            user_params=self.user_params,
+            astro_params=self.astro_params,
+            flag_options=self.flag_options,
+            random_seed=self.random_seed,
+        )
+
+    def evolve_inputstructs(self, **kwargs):
```
steven-murray, line 282: I think `evolve_input_structs`

src/py21cmfast/drivers/param_config.py
```diff
@@ -172,63 +209,124 @@ def is_compatible_with(self, other: InputParameters) -> bool:
...
+        """Return an altered clone of the `InputParamters` structs.
```
steven-murray, line 283: typo in `InputParameters`

src/py21cmfast/drivers/param_config.py
```diff
@@ -172,63 +209,124 @@ def is_compatible_with(self, other: InputParameters) -> bool:
...
+
+        Unlike clone(), this function takes fields from the constituent `InputStruct` classes
+        and only overwrites those sub-fields instead of the entire field
+        """
+        cosmo_params = self.cosmo_params.clone(
+            **{
+                k: v
+                for k, v in kwargs.items()
+                if k in self.cosmo_params.asdict().keys()
```
steven-murray, line 292: not too important, but I think doing the `.asdict()` inside the comprehension here might call the method for each iteration of the loop. You could save `cdict = self.cosmo_params.asdict()` outside the loop and use it inside the comprehension (and you don't need the `.keys()`)

src/py21cmfast/drivers/param_config.py
```diff
@@ -172,63 +209,124 @@ def is_compatible_with(self, other: InputParameters) -> bool:
...
+
+        Takes `InputStruct` fields as keyword arguments which overwrite the template/default fields
+        """
+        return cls(
+            **create_params_from_template(name),
+            **kwargs,
+        )
 
     @classmethod
     def from_inputstructs(
```
steven-murray, line 332: again I prefer `from_input_structs`

src/py21cmfast/drivers/param_config.py
```diff
@@ -246,46 +345,21 @@ def from_inputstructs(
...
-            }
-        )
-        flag_options = FlagOptions.new(
-            {
-                k: v
-                for k, v in kwargs.items()
-                if k in [f.name for f in attrs.fields(FlagOptions)]
-            }
-        )
+    def from_defaults(cls, **kwargs):
```
steven-murray, line 352: just wanted to check if this is still necessary, since we can just give the actual fields defaults now?

src/py21cmfast/drivers/param_config.py
```diff
@@ -306,14 +380,16 @@ def __repr__(self):
             + f"flag_options: {repr(self.flag_options)}\n"
         )
 
-    # TODO: methods for equality: w/o redshift, w/o seed
+    # TODO: methods for equality: w/o redshifts, w/o seed
 
 
-def check_redshift_consistency(inputs: InputParameters, output_structs):
+def check_redshift_consistency(redshift, output_structs):
     """Check the redshifts between provided OutputStruct objects and an InputParameters instance."""
```
steven-murray, line 387: docstring needs updating

src/py21cmfast/wrapper/outputs.py
```diff
@@ -27,6 +27,10 @@
 logger = logging.getLogger(__name__)
 
 
+# NOTE: The `inputs` arguments to the __init__ methods are set this way such that the
+#   required fields (`_inputs`) can be read either from the file
+#   (done in structs.OutputStruct.__init__) or the input struct (done here)
+# TODO: there is certainly a better way to organise it
```
steven-murray, line 33: Noted -- I will try to make this smoother in my PR

src/py21cmfast/wrapper/structs.py
```diff
@@ -134,7 +134,7 @@ def new(cls, x: dict | InputStruct | None = None, **kwargs):
         if isinstance(x, dict):
             return cls(**x, **kwargs)
         elif isinstance(x, InputStruct):
-            return x
+            return x  # x.clone(**kwargs)????
```
steven-murray, line 137: yes, I think you're right

steven-murray (commented):
Thanks again @daviesje, this is a decent chunk of work! I think your implementation is good. I could see that it might be able to be cleaner once we rid ourselves of globals (at least most of them), but maybe we do that in a different PR. For now, I think this is the right API.

src/py21cmfast/drivers/param_config.py
daviesje: We can't change the length of the dict inside it's own loop

src/py21cmfast/drivers/param_config.py
daviesje: I suppose not, there is a slightly different behaviour when someone tries to define an incomplete structure (via `from_output_structs` using an `InitialConditions` or `__init__` missing some arguments). Currently it refuses to let you build the structure at all, and you have to explicitly set the default structure if you want it. If we set default `InputStruct` fields here it will fill those in with the defaults, which may be more convenient (i.e the only required argument is the random seed). The only possible downside is that bad usage will be caught later: i.e ```InputParameters.from_output_struct((perturbed_field),...) --> ionize_box(astro_params=...)``` will catch errors during the ionize step instead of the parameter instantiation.

src/py21cmfast/drivers/param_config.py
daviesje: I think I can remove this method entirely now that we require every field

PR #447, opened 2024-12-09
Fixed Template Filepath Reading

samgagnon:
`InputParameters.from_template()` failed when supplied a filepath rather than template name due to the `create_params_from_template()` callable passing a string to `tomllib.load()` rather than a stream object. This has been fixed so that custom templates can now be supplied via their filepath.


steven-murray (approved):
Thanks @samgagnon !

PR #451, opened 2025-01-06
Fix recombinations error flag

daviesje:
The inhomogeneous recombinations model has an error check which exits when a non-finite recombination rate is found. since the refactor of `Ionisationbox.c`, this error flag was used uninitialised, which caused problems when debug mode was enabled, throwing an error occasionally without a non-finite value.

I have also added a log entry when the error is triggered. Since the error causes an exit after the loop, this shouldn't print too many times.


steven-murray (approved):
Thanks @daviesje

PR #452, opened 2025-01-22
remove global_params

daviesje:
I'm setting up a draft PR for removing the global parameter structure, as many of these are under/unused.

For now I've removed the parameters which were completely unused and I've commented on the others with suggestions of what to do with them.

I will begin to go through the parameters and move/delete them as necessary.

### DIFF
--- a/src/py21cmfast/src/InputParameters.c
+++ b/src/py21cmfast/src/InputParameters.c
@@ -18,30 +18,35 @@ CosmoParams *cosmo_params_global;
 AstroParams *astro_params_global;
 FlagOptions *flag_options_global;
 
+// These need to be removed, I am commenting below roughly where/when
+//  they are used, and my tentative decision on what to do with them
 GlobalParams global_params = {
-    .ALPHA_UVB = 5.0,
-    .EVOLVE_DENSITY_LINEARLY = 0,
-    .SMOOTH_EVOLVED_DENSITY_FIELD = 0,
-    .R_smooth_density = 0.2,
-    .HII_ROUND_ERR = 1e-5,
-    .FIND_BUBBLE_ALGORITHM = 2,
-    .N_POISSON = 5,
-    .T_USE_VELOCITIES = 1,
-    .MAX_DVDR = 0.2,
-    .DELTA_R_HII_FACTOR = 1.1,
-    .DELTA_R_FACTOR = 1.1,
-    .HII_FILTER = 0,
-    .INITIAL_REDSHIFT = 300.,
-    .R_OVERLAP_FACTOR = 1.,
-    .DELTA_CRIT_MODE = 1,
-    .HALO_FILTER = 0,
-    .OPTIMIZE = 0,
-    .OPTIMIZE_MIN_MASS = 1e11,
+    .ALPHA_UVB = 5.0, // IonisationBox, for Gamma (move to AstroParams)
+    .EVOLVE_DENSITY_LINEARLY = 0, //PerturnbedField, uses linear growth instead of 1/2LPT (move to UserParams)
+    .SMOOTH_EVOLVED_DENSITY_FIELD = 0, //PerturbedField, smooth field after perturb (delete)
+    .R_smooth_density = 0.2, //With above, radius for smoothing (delete)
+    .HII_ROUND_ERR = 1e-5, // IonisationBox, Nion threshold to run the ES (move to define)
+    .FIND_BUBBLE_ALGORITHM = 2, // IonisationBox central pixel vs sphere (move to UserParams)
+    .MAX_DVDR = 0.2, //BrightnessTemp, maximum velocity gradient for RSDS (move to AstroParams?)
+    .DELTA_R_HII_FACTOR = 1.1, //IonisationBox, radius factor for reion ES (move to AstroParams)
+    .DELTA_R_FACTOR = 1.1, //HaloField, radius factor for DexM halo ES (move to UserParams)
 
+    .HII_FILTER = 0, //IonisationBox, filter for reion ES (move to AstroParams)
+    .HALO_FILTER = 0, //HaloField, filter for halo ES (move to UserParams)
+    .HEAT_FILTER = 0, //SpinTemperature, filter for Ts model (move to AstroParams)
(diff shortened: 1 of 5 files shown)


src/py21cmfast/src/InputParameters.c
```diff
@@ -18,72 +18,77 @@ CosmoParams *cosmo_params_global;
...
+    .EVOLVE_DENSITY_LINEARLY = 0, //PerturnbedField, uses linear growth instead of 1/2LPT (move to UserParams)
+    .SMOOTH_EVOLVED_DENSITY_FIELD = 0, //PerturbedField, smooth field after perturb (delete)
+    .R_smooth_density = 0.2, //With above, radius for smoothing (delete)
+    .HII_ROUND_ERR = 1e-5, // IonisationBox, Nion threshold to run the ES (move to define)
+    .FIND_BUBBLE_ALGORITHM = 2, // IonisationBox central pixel vs sphere (move to UserParams)
+    .MAX_DVDR = 0.2, //BrightnessTemp, maximum velocity gradient for RSDS (move to AstroParams?)
+    .DELTA_R_HII_FACTOR = 1.1, //IonisationBox, radius factor for reion ES (move to AstroParams)
+    .DELTA_R_FACTOR = 1.1, //HaloField, radius factor for DexM halo ES (move to UserParams)
 
+    .HII_FILTER = 0, //IonisationBox, filter for reion ES (move to AstroParams)
```
steven-murray, line 34: I think anything that is not a continuous parameter should be in flag options, not astro params

daviesje:
A few notes here:

- I have removed (mostly commented) the WDM model that existed via `global_params.P_CUTOFF` since it seemed inconsistent.
- I have forced the `global_params.USE_ADIABATIC_FLUCTUATIONS` to True. This seemed like something we always wanted to be true.
- I am thinking about removing the `DEXM_OPTIMIZE` model, but this should probably go in another PR

Paths and memory allocation factors were moved into the config, and the config file (along with its warnings) are gone, substituted by a single instance which automatically passes necessary information to the backend.

src/py21cmfast/_cfg.py
```diff
@@ -16,68 +19,59 @@ class ConfigurationError(Exception):
...
     """Simple over-ride of dict that adds a context manager."""
 
     _defaults = {
-        "direc": "~/21cmFAST-cache",
+        "direc": Path("~/21cmFAST-cache").expanduser(),
         "regenerate": False,
         "write": True,
         "cache_param_sigfigs": 6,
         "cache_redshift_sigfigs": 4,
         "ignore_R_BUBBLE_MAX_error": False,
```
steven-murray, line 32: I actually don't think any of these settings are being used any more :/ We probably _should_ be using the sigfigs -- we shouldn't lose that functionality. But I don't think the new output struct framework actually accesses this info any more.

src/py21cmfast/_cfg.py
```diff
@@ -16,68 +19,59 @@ class ConfigurationError(Exception):
...
+        for k in self._c_config_settings.keys():
+            self.pass_to_backend(k, self[k])
+
+    def __setitem__(self, key, value):
+        """Set an item in the config. Also updating the backend if it exists there."""
+        super().__setitem__(key, value)
+        if key in self._c_config_settings.keys():
+            self.pass_to_backend(key, value)
+
+    def pass_to_backend(self, key, value):
```
steven-murray, line 66: Let's make this a semi-private method

steven-murray (commented):
@daviesje thanks, this is super nice. I think I agree with everything you've done here. I would say that the config class has actually become almost unnecessary now, so maybe we should htink about whether it is still reuqired (or whether it should _only_ be kept around for these lingering globals). 

Also, I'm not sure I _totally_ agree with all the places you're putting parameters. But I could be convinced. In my head, `AstroParams` are for actual physical parameters that can be constrained, while `FlagOptions` are for discrete options (True/False or choices from a list). This leaves no actual place for options that are continuous, but not a physical parameter (e.g. MAX_DVDR). So I'm not sure what the best place for those are. We should perhaps chat about this.

src/py21cmfast/_cfg.py
daviesje: I can remove these as necessary, but I think it's a good idea to have *some* place we can put variables which don't affect the run output, and may need to be passed in.

src/py21cmfast/_cfg.py
daviesje: Done.

daviesje:
As for the placement of the globals, I only considered whether they are utilised before or after the PerturbField finishes. Since we tie the parameter sets to these structs I thought it was more important to use that distinction, even if categorically they make less sense. We can talk about alternate ways to organise the parameters.

src/py21cmfast/_cfg.py
steven-murray: Yeah, I think having this structure around makes sense. I think for this PR, let's leave everything in there. Once we merge the output struct PR, we'll have a better sense of what is still required.

steven-murray:
Yeah, the before/after PerturbedField is obviously a very important distinction.

daviesje:
More notes:

- I've updated the `upload-artifact` action since version the was deprecated by Github
- I've folded in some fixes that were in a separate PR #454 since they were needed to fix some tests
- A bunch of issues were found and fixed by removing the `**global_kwargs` arguments from all the functions, since previously removed arguments to functions were being lumped in there and errors weren't being thrown if these were not global parameters

src/py21cmfast/_cfg.py
```diff
@@ -19,6 +18,73 @@ class ConfigurationError(Exception):
...
+            + "("
+            + ";".join(
+                f"{k}={str(v)}"
+                for k, v in sorted(self.items())
+                if k not in filter_params
+            )
+        ) + ")"
+
+
 # TODO: force config to be a singleton
```
steven-murray, line 88: An easy way to achieve some of this is just to make `Config` semi-private... i.e. `_Config`. It actually doesn't matter if someone creates a new instance of `_Config` -- it'll just never be used.

steven-murray (approved):
Thanks @daviesje. I made a couple of comments, but none of them are absolutely necessary. I am a little scared of how this is going to interact with #445. Might be a nasty merge.

daviesje:
Note to move some of the AstroParams which we won't want to infer into FlagOptions for a future renaming

PR #454, opened 2025-02-10
Fixing photoncons

daviesje:
Some of the InputParamteters changes were breaking the photon conservation models. The biggest issue was that in v3, the conditional defaults set `R_BUBBLE_MAX` to 15 on the calibration simulation if `INHOMO_RECO` was True. When the conditional defaults were changed and the parameter structs were frozen the calibrations were done with `R_BUBBLE_MAX=50` which messed with the photon conservation. This has been set back to the original behaviour, however we may want to think about this model a bit more in the future, each of the photon nonconservation corrections are unstable in certain areas of parameter space.

I've also snuck in a fix to the HaloBox caching, since the objects were not being saved correctly. I'm guessing the pending output PR fixes this anyway but I needed it for testing. 

Todos:
- figure out why the tests in `test_segfaults.py` didn't catch these errors and update the tests

### DIFF
--- a/src/py21cmfast/drivers/coeval.py
+++ b/src/py21cmfast/drivers/coeval.py
@@ -89,7 +89,7 @@ class _HighLevelOutput:
         kinds = {
             "init": InitialConditions,
             "perturb_field": PerturbedField,
-            "halobox": HaloBox,
+            "halo_box": HaloBox,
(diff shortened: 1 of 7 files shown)


daviesje:
Closing this PR as I have merged it into #452

PR #455, opened 2025-02-20
Fix gamma12 in sub-sampler HaloBox

daviesje:
The prefactors to the integrated portions of the HaloBox field (either sub-sampler or fixed halo grids) were missing a term. This will have a minor effect on recombinations. Thanks @jordanflitter for finding this


steven-murray (approved):
Thanks!

PR #456, opened 2025-02-24
Debug dimensions

daviesje:
Many of the debug printing functions were out of date, having integers which could go above 2 billion, indexing grids incorrectly etc. This PR just updates them so that they work as intended.

### DIFF
--- a/src/py21cmfast/src/IonisationBox.c
+++ b/src/py21cmfast/src/IonisationBox.c
@@ -1414,7 +1414,7 @@ int ComputeIonizedBox(float redshift, float prev_redshift, UserParams *user_para
         if (flag_options->INHOMO_RECO){
             set_recombination_rates(box,previous_ionize_box,perturbed_field,&ionbox_constants);
         }
-        
+
         if(!ionbox_constants.fix_mean){
             //if we don't fix the mean, make the mean_f_coll in the output reflect the box
             //since currently it is the global expected mean from the Unconditional MF


src/py21cmfast/src/PerturbField.c
```diff
@@ -527,9 +529,9 @@ int ComputePerturbField(
 
         LOG_SUPER_DEBUG("density_perturb: ");
         if(user_params->PERTURB_ON_HIGH_RES){
-            debugSummarizeBoxComplex(HIRES_density_perturb, dimension, user_params->NON_CUBIC_FACTOR, "  ");
+            debugSummarizeBox((float *)HIRES_density_perturb, dimension, dimension, 2*(dimension_z/2 + 1), "  ");
```
steven-murray, line 532: I'm a little perplexed about the `2*(dimension_z/2+1)` here... is it FFT padding?

steven-murray (commented):
Thanks @daviesje, this is well-needed. I just had the one question.

src/py21cmfast/src/PerturbField.c
daviesje: yeah, comparing `R_INDEX`, `R_FFT_INDEX`, and `C_INDEX` we see that the fft padding adds 2 to the z-dimension on even-dimensioned arrays and 1 to odd, and so the indexing uses `dimension_z`, `(dimension_z / 2 + 1)`  and `2*(dimension_z/2 + 1)` respectively. Since the fftw box is in real-space here, I use the float version with the padded z-index.

https://fftw.org/fftw3_doc/Multi_002dDimensional-DFTs-of-Real-Data.html

steven-murray: OK, great!

PR #464, opened 2025-03-07
Minor fixes to plotting scripts, adding turnover masses to global variables

daviesje:
- Updated global plotting script for the new output structures, which were still using lightcone.node_redshifts, And added a test to check for it.
- Added the turnover masses back into the possible lightcone global quantities
- Changed the default HALOMASS_CORRECTION variable to produce slightly better results with the default timestep

### DIFF
--- a/src/py21cmfast/drivers/lightcone.py
+++ b/src/py21cmfast/drivers/lightcone.py
@@ -459,14 +459,14 @@ def _run_lightcone_from_perturbed_fields(
     ):
         # Save mean/global quantities
         for quantity in global_quantities:
-            if "log10_mturn_acg" == quantity:
-                lightcone.global_quantities[quantity][
-                    iz
-                ] = coeval.ionized_box.log10_Mturnover_ave
-            elif "log10_mturn_mcg" == quantity:
-                lightcone.global_quantities[quantity][
-                    iz
-                ] = coeval.ionized_box.log10_Mturnover_MINI_ave
+            if quantity == "log10_mturn_acg":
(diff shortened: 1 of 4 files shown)


src/py21cmfast/drivers/lightcone.py
```diff
@@ -459,9 +459,18 @@ def _run_lightcone_from_perturbed_fields(
...
         for quantity in global_quantities:
-            lightcone.global_quantities[quantity][iz] = np.mean(
-                getattr(coeval, quantity)
-            )
+            if quantity == "log10_mturn_acg":
+                lightcone.global_quantities[quantity][iz] = (
+                    coeval.ionized_box.log10_Mturnover_ave
+                )
+            elif quantity == "log10_mturn_mcg":
+                lightcone.global_quantities[quantity][iz] = (
```
steven-murray, line 467: Why can't we treat these like other globals and just take the mean of the output field?

steven-murray (commented):
Thanks for the plotting fixes! I have one question about the global Mturnover, but otherwise this is good

src/py21cmfast/drivers/lightcone.py
daviesje: These fields aren't part of the output structs at the moment. Currently they are hidden in the backend, with the exception of the `XraySourcebox` which saves the MCG part when halos are used.

steven-murray: Ahh I see. No probs.

steven-murray (approved):

Issue #450, opened 2024-12-11
Improve testing suite


daviesje:
A few issues with the automated tests should be addressed.

A short/incomplete wishlist of extra tests:
- tests of spin-temperature calculations i.e how a particular radiative background effects Ts, Tk, x_e
- tests of mass function integrals / values, perhaps against `hmf`.
- Performance tests on a few simple cases to ensure nothing has slowed down.

We also want to rein in the time taken on the tests, some interpolation tables and example runs (usually the minihalo ones) run very slowly, and this can impact development

daviesje:
Adding a note here to consider that running lightcones with `USE_MINI_HALOS=True` and `USE_HALO_FIELD=False` in the integration tests or the segfault tests is taking the majority of the test time, up to 20 minutes. We should try to reduce this

PR #466, opened 2025-03-08
Fixing Some Issues with New Mturn Paramterisation

daviesje:
Recently, we allowed the free parameter M_TURN to function in the minihalo model, as a lower limit set by supernovae feedback. Previously, the atomic cooling threshold was used both as the lower turnover for ACGs and the upper turnover for MCGs.

The correct behaviour would have been to separate these two turnovers entirely, where the upper turnover for MCGs is *always* the atomic cooling threshold, and the lower turnover for ACGs is set by the maximum of M_TURN and the cooling threshold. When implementing this new lower limit, I did not do this, meaning that the maximum of the cooling and M_TURN was used for both turnovers.

This PR fixes this issue by separating the two turnovers properly, as well as trying to clean up the table/integral functions a little to make them less confusing. The latter involves doing some repeated computation once per snapshot (`atomic_cooling_threshold`,`dicke`,`Mass_limit_bisection` etc) but this should be insignificant, and worth the readibility.

Missing this bug in the first place exposes some gaps in our tests. Where we do have internal consistency checks for whether the tables match the integrals, but we cannot catch:
- when the integral functions are called incorrectly in computation functions
- when the integral functions themselves have errors

Both of these tests would require comparing some known cases of an output function or integral, which can be difficult to construct, but maybe we can store values for a couple of known integrals, maybe we can some numbers for specific cases `delta=0`, `mturn=1e8` etc. However this could have some issues similar to the integration tests, where changes to the code require updates to the comparison data.

### DIFF
--- a/src/py21cmfast/src/HaloBox.c
+++ b/src/py21cmfast/src/HaloBox.c
@@ -126,7 +126,7 @@ int get_box_averages(double M_min, double M_max, double M_turn_a, double M_turn_
                                             1., consts->Mlim_Fstar_mini, 0.);
     }
     if(flag_options_global->USE_TS_FLUCT){
-        integral_xray = Xray_General(consts->redshift, lnMmin, lnMmax, M_turn_m, M_turn_a,
+        integral_xray = Xray_General(consts->redshift, lnMmin, lnMmax, M_turn_a, M_turn_m,
                                             consts->alpha_star, consts->alpha_star_mini, consts->fstar_10,
                                             consts->fstar_7, consts->l_x, consts->l_x_mini, consts->t_h, consts->t_star,
                                             consts->Mlim_Fstar, consts->Mlim_Fstar_mini);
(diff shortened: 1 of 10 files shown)


steven-murray (approved):
@daviesje this all looks good to me, though I'm confident that if the new threshold wasn't implemented everywhere it needed to be, I would miss it. 

It would be really good to have better tests of this so we're more sure of ourselves, but I also don't have great ideas along those lines.

daviesje:
> @daviesje this all looks good to me, though I'm confident that if the new threshold wasn't implemented everywhere it needed to be, I would miss it.
> 
> It would be really good to have better tests of this so we're more sure of ourselves, but I also don't have great ideas along those lines.

That's true, and part of solving #450 will Involve a check against some known HMF integral values. I think it also would help to make the interfaces for these tables a bit more transparent, by shifting more of the logic to the constant structures. But I don't think that's for this PR.

PR #468, opened 2025-03-16
Backend interface overhaul

daviesje:
This PR is a cleanup of the interface between the backend and frontend, which makes `cfuncs.py` module as well as the testing much cleaner.

This has been implemented largely by creating functions in the backend which handle all the necessary initialisation so that we no longer have to make multiple backend calls to calculate a single integral. A few additional features include:

- HMF integrals now use the scaling relation stuctures in `scaling_relations.c` so that they are easier to manage and take fewer arguments.
- USE_INTERPOLATION_TABLES is now ternary, `no-interpolation` `sigma-interpolation` and `hmf-interpolation` control the two levels of interpolation table that exist in the code. This change was mostly done to make testing easier and faster, I don't think too many people will use the `sigma-interpolation` level.
- Tests for galaxy properties from the halo sampler have been activated and tests for integrals within HaloBox have been added


src/py21cmfast/src/photoncons.c
```diff
@@ -159,15 +159,15 @@ int InitialisePhotonCons(UserParams *user_params, CosmoParams *cosmo_params,
             z0 = 1./(a+delta_a) - 1.;
             z1 = 1./(a-delta_a) - 1.;
 
+            sc_i = scaling_consts_z_copy(zi,astro_params,flag_options,&sc_i,false);
```
steven-murray, line 162: What does this function do? Copy over contents of a ScalingConstants struct and then update for a new redshift? If so, would a better name be something like `evolve_scaling_constants_to_redshift()`?

src/py21cmfast/wrapper/cfuncs.py
```diff
@@ -16,11 +17,82 @@
...
+
+# TODO: a lot of these assume input as numpy arrays via use of .shape, explicitly require this
+
+
+def broadcast_params(func: Callable) -> Callable:
+    """Decorator to broadcast the parameters to the C library before calling the function."""
+
+    def wrapper(*args, **kwargs):
+        inputs = kwargs.get("inputs")
+        if inputs.astro_params:
```
steven-murray, line 31: In our current implementation, `astro_params` is always set on `inputs`. It's never `None`. So this if/else is unnecessary

src/py21cmfast/wrapper/cfuncs.py
```diff
@@ -16,11 +17,82 @@
...
+                inputs.user_params.cstruct,
+                inputs.cosmo_params.cstruct,
+            )
+        return func(*args, **kwargs)
+
+    return wrapper
+
+
+def init_backend_ps(func: Callable) -> Callable:
+    """Decorator to initialise the backend PS before calling the function."""
```
steven-murray, line 49: Can we add some extra context to the docstring -- like that while this is not always necessary, it is very fast and means that we can confidently call any backend function knowing that the cosmo etc is up-to-date?

src/py21cmfast/wrapper/cfuncs.py
```diff
@@ -16,11 +17,82 @@
...
+            != "no-interpolation"
+        ):
+            lib.initialiseSigmaMInterpTable(sigma_min_mass, sigma_max_mass)
+        return func(*args, **kwargs)
+
+    return wrapper
+
+
+def init_gl(func: Callable) -> Callable:
+    """Decorator to initialise the Gauss-Legendre if required before calling the function."""
```
steven-murray, line 77: I'm not sure what "initialise the Gauss-Legendre" means -- is this setting up GSL integration methods?

src/py21cmfast/wrapper/cfuncs.py
```diff
@@ -16,11 +17,82 @@
...
+# Ideally, backend functions that we access here should do all the broadcasting/initialisation themselves
+# These decorators are for lower functions which are called directly in one or two lines, like delta_crit
+
+# TODO: a lot of these assume input as numpy arrays via use of .shape, explicitly require this
+
+
+def broadcast_params(func: Callable) -> Callable:
+    """Decorator to broadcast the parameters to the C library before calling the function."""
+
+    def wrapper(*args, **kwargs):
```
steven-murray, line 29: For this and other decorators that require `inputs`, its probably easier to have the argspec be `wrapper(*args, inputs: InputParameters, **kwargs):` so that you can directly refer to it (it also makes it more explicit that the `inputs` argument is _required_ in such functions, and can't be passed as a position arg).

src/py21cmfast/wrapper/cfuncs.py
```diff
@@ -366,931 +434,393 @@ def construct_fftw_wisdoms(
...
-def get_delta_crit_nu(user_params, sigma, growth):
-    """Uses the nu paramters (sigma and growth factor) to get critical density."""
-    # None of the parameter structs are used in this function so we don't need a broadcast
-    return np.vectorize(lib.get_delta_crit)(user_params.cdict["HMF"], sigma, growth)
-
-
 def evaluate_sigma(
-    user_params: UserParams,
-    cosmo_params: CosmoParams,
+    inputs,
```
steven-murray, line 438: remember typing

src/py21cmfast/wrapper/cfuncs.py
```diff
@@ -366,931 +434,393 @@ def construct_fftw_wisdoms(
...
-        np.log(M_min),
-        np.log(M_max),
-        M_cond,
-        sigma,
-        delta,
-        0,  # GSL-QAG
+def integrate_chmf_interval(
+    inputs: InputParameters,
+    redshift: float,
+    lnM_lower: Sequence[float],
```
steven-murray, line 535: I would prefer to maintain python conventions wherever possible, so use lower-case names `lnm_lower` etc. Also, should there be a check inside the function like `assert np.all(lnM_upper > lnM_lower)`?

src/py21cmfast/wrapper/cfuncs.py
```diff
@@ -366,931 +434,393 @@ def construct_fftw_wisdoms(
...
+        ffi.cast("double *", ffi.from_buffer(densities)),
+        ffi.cast("double *", ffi.from_buffer(log10mturns)),
+        ffi.cast("double *", ffi.from_buffer(xray)),
     )
-    return Xray
+
+    return xray
 
 
 def halo_sample_test(
```
steven-murray, line 806: I'm a little confused with the name of this function -- is it purely a function for testing? If so, should it be defined in this module or in the tests themselves?

steven-murray (commented):
Thanks @daviesje, I think this looks quite nice. I have a couple of small comments. Otherwise, I think we need to:

- Get the code coverage here to ~100%
- Add some higher-level documentation (maybe developer-level) about how to do these integrals and what they mean.
- Have a switch to turn off plotting inside the tests, because I think it's adding a bit to the testing time (and is really only necessary when things are _failing_)

src/py21cmfast/src/HaloBox.c
```diff
@@ -322,22 +307,17 @@ int set_fixed_grids(double M_min, double M_max, InitialConditions *ini_boxes,
...
             h_count = EvaluateNhalo(dens, growth_z, lnMmin, lnMmax, M_cell, sigma_cell, dens);
             mass_intgrl = EvaluateMcoll(dens, growth_z, lnMmin, lnMmax, M_cell, sigma_cell, dens);
-            intgrl_fesc_weighted = EvaluateNion_Conditional(dens,l10_mturn_a,growth_z,M_min,M_max,M_cell,sigma_cell,
-                                            consts->Mlim_Fstar,consts->Mlim_Fesc,false);
-            intgrl_stars_only = EvaluateSFRD_Conditional(dens,growth_z,M_min,M_max,M_cell,sigma_cell,
-                                            pow(10.,l10_mturn_a),consts->Mlim_Fstar);
+            intgrl_fesc_weighted = EvaluateNion_Conditional(dens,l10_mturn_a,growth_z,M_min,M_max,M_cell,sigma_cell,consts,false);
+            intgrl_stars_only = EvaluateSFRD_Conditional(dens,growth_z,M_min,M_max,M_cell,sigma_cell,consts);
+            //TODO: SFRD tables still assume no reion feedback, this should be fixed
+            //  although it doesn't affect the histories (only used in Ts) it makes outputs wrong for post-processing
```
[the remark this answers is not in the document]
daviesje: I've turned this one into an issue. We should do a sweep before releasing for the rest

daviesje:
I added a test to compare our matter field (HMF, sigma, power) with `hmf`, and there are some errors which should be looked into. I've xfailed the test for now. All MFs show similar (in magnitude, different shape) errors as this one
[tests.test_cfuncs.py--test_hmf_values[BBKS-PS].pdf](https://github.com/user-attachments/files/19323238/tests.test_cfuncs.py--test_hmf_values.BBKS-PS.pdf)

steven-murray:
@daviesje what is your spectral index that you're using in both cases? It looks like the low-k behaviour of the power spectrum is a little off, which should be purely n_s.

steven-murray (approved):

PR #469, opened 2025-03-17
test: add tests for the RunCache

steven-murray:
This adds a few improvements to the `RunCache` class, as well as a bunch of tests.

I wrote this because I realized that there were errors in the `get_all_boxes_at_z` method, and I also wanted an easy way to output a `Coeval` from this.

### DIFF
--- a/src/py21cmfast/io/caching.py
+++ b/src/py21cmfast/io/caching.py
@@ -14,6 +14,7 @@ import attrs
 import numpy as np
 
 from .._cfg import config
+from ..wrapper import outputs as op
 from ..wrapper.inputs import InputParameters
 from ..wrapper.outputs import OutputStruct
 from .h5 import read_inputs, read_output_struct, write_output_to_hdf5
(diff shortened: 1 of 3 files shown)


src/py21cmfast/io/caching.py
```diff
@@ -232,13 +233,13 @@ class RunCache:
...
-    HaloBox: tuple[Path] = _dict_of_paths_field()
-    PerturbHaloField: tuple[Path] = _dict_of_paths_field()
-    XraySourceBox: tuple[Path] = _dict_of_paths_field()
+    PerturbedField: dict[float, Path] = _dict_of_paths_field()
+    TsBox: dict[float, Path] = _dict_of_paths_field()
+    IonizedBox: dict[float, Path] = _dict_of_paths_field()
+    BrightnessTemp: dict[float, Path] = _dict_of_paths_field()
+    HaloBox: dict[float, Path] | None = _dict_of_paths_field()
+    PerturbHaloField: dict[float, Path] | None = _dict_of_paths_field()
+    XraySourceBox: dict[float, Path] | None = _dict_of_paths_field()
```
daviesje, line 242: I'm guessing since this is constructed directly from the inputs, which is frozen, indexing via a float won't run into any precision issues?

daviesje (approved):
Looks good!

src/py21cmfast/io/caching.py
steven-murray: Yeah, that is my thought. If we run into trouble later, we can truncate the float or something

PR #474, opened 2025-03-27
Rare halo Truncation and some extra parameter checks

daviesje:
This is a cleaner implementation of the rare-halo truncation I implemented in the `2025-database-runs` branch. It uses the same criteria as DexM (>7 sigma) for determining halos which have been successively sampled at too high a mass. It then gives them a single progenitor at it's expected mass, given by it's CHMF integral and any corrective factors.

I have also disabled the previous behaviour of extrapolating the CHMF using an exponential below the table minimum probability. The rng *can* sample exactly zero and this causes too many progenitors to be at the condition mass, furthermore the exponential was simply not steep enough.

A few other changes were introduced here regarding parameter clashes. We are now forcing PS or ST mass functions when using halos, and throwing errors instead of warnings when trying to enable modules which have no effect without halos (upper stellar turnover and exponential filter), to avoid confusion.

I have also corrected some of the templates, the previous default `HII_FILTER` was the sharp-k, and so I've added that to the templates without discrete halos.


tests/test_halo_sampler.py
```diff
@@ -34,7 +34,8 @@
 @pytest.mark.parametrize("cond_type", ["cat", "grid"])
 @pytest.mark.parametrize("cond", range(len(options_delta)))
 def test_sampler(name, cond, cond_type, plt):
-    redshift, kwargs = cint.OPTIONS_HMF[name]
+    _, kwargs = cint.OPTIONS_HMF[name]
+    redshift = 7.0  # we need the largest condition mass to not be >7 sigma
```
steven-murray, line 38: Why does the largest condition mass need to be <7? Because it will trigger the rare halo case?

steven-murray (approved):

tests/test_halo_sampler.py
daviesje: Yeah, if the condition mass trips the rare halo check it's not a very useful test

PR #475, opened 2025-03-28
fixed segfault bug without halos but with minihalos and no lya heating

jordanflitter:
fixed segfault bug without halos but with minihalos and no lya heating


daviesje (approved):
Looks good to me. It would be nice to add this case to our segfault tests but this would end up doubling test time (see #450). I'm happy to put this in now

PR #479, opened 2025-04-02
dNrec_box exists only if INHOMO_RECO=True

jordanflitter:
This change was made to reduce used memory when `INHOMO_RECO=False`.


steven-murray:
To get the tests passing, I think all you need to do is remove `dNrec_box` from the lightcone fields in the `test_drivers_lc.py` test file.

tests/test_drivers_lc.py
```diff
@@ -51,7 +51,7 @@ def test_lightcone_quantities(
     assert "log10_mturn_mcg" in lc.global_quantities
 
     # dNrec is not filled because we're not doing INHOMO_RECO
-    assert lc.lightcones["dNrec_box"].max() == lc.lightcones["dNrec_box"].min() == 0
+    # assert lc.lightcones["dNrec_box"].max() == lc.lightcones["dNrec_box"].min() == 0
```
daviesje, line 54: Just remove these comments. We know that dNrec isn't always present

daviesje (approved):
g2g after getting pre-commit and tests to pass

PR #480, opened 2025-04-02
Detect c compiler more interactively

moll-re:
Hello, this is a tentative PR to fix an installation issue when simply using pip to install the package.

During the installation this version tries to infer the compiler from the environment variable `CC` and only then falls back onto the compiler returned by `sysconfig.get_config_var("CC")`.

Since binary distributions of programs are so common, it is often the case that the compiler used to build python is not present on the system. But `sysconfig.get_config_var("CC")` returns just that - the compiler that was used to build this particular distribution. Using environment variables instead is a more flexible approach in my opinion.

Let me know if you are interested in integrating this PR and whether you require changes/tests/documentation.

### DIFF
--- a/build_cffi.py
+++ b/build_cffi.py
@@ -6,7 +6,13 @@ import sysconfig
 from cffi import FFI
 
 # Get the compiler. We support gcc and clang.
-_compiler = sysconfig.get_config_var("CC")
+_compiler = os.environ.get("CC", None)
+if _compiler is None:
+    # If CC is not set, use the default compiler from sysconfig - the one used to build Python
+    _compiler = sysconfig.get_config_var("CC")
+    print(f"No environment variable CC set, trying to use compiler from sysconfig: {_compiler}")
+else:
+    print(f"Compiler specified as environment variable CC: {_compiler}")
 
 if "gcc" in _compiler:
     compiler = "gcc"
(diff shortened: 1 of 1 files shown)


build_cffi.py
```diff
@@ -6,7 +6,15 @@
...
-_compiler = sysconfig.get_config_var("CC")
+_compiler = os.environ.get("CC", None)
+if _compiler is None:
+    # If CC is not set, use the default compiler from sysconfig - the one used to build Python
+    _compiler = sysconfig.get_config_var("CC")
+    print(
+        f"No environment variable CC set, trying to use compiler from sysconfig: {_compiler}"
+    )
+else:
+    print(f"Compiler specified as environment variable CC: {_compiler}")
```
steven-murray, line 17:
```suggestion
_compiler = os.environ.get("CC", sysconfig.get_config_var("CC"))
```

steven-murray (changes requested):
Thank you very much @moll-re -- I like this suggestion a lot, but I think we should not be printing during the build. If you think it's important, we _could_ add a `logger.info()` statement?

moll-re:
I don't have any strong opinions on that. My thought process is that it would make it easier to retrace the issue should there be a problem. But in that case it might be sufficient to adapt the exception that is being raised instead.
If you want I can adapt the code accordingly or we can keep it as is.

steven-murray:
Thanks @moll-re -- it would be nice to be able to print this info if e.g. `-vv` was passed on the cmd line to `pip install ...`. I'm not quite sure how to do that though :-P

steven-murray:
Oh I see! I'd accept a PR to bring this back :-)

PR #481, opened 2025-04-02
J_21_LW exists only if USE_MINI_HALOS=True

jordanflitter:
This change was made to reduce used memory when `USE_MINI_HALOS=False`.

### DIFF
--- a/src/py21cmfast/wrapper/outputs.py
+++ b/src/py21cmfast/wrapper/outputs.py
@@ -1093,9 +1093,9 @@ class TsBox(OutputStructZ):
             int(inputs.user_params.NON_CUBIC_FACTOR * inputs.user_params.HII_DIM),
         )
         out = {
-            "Ts_box" : Array(shape, dtype=np.float32),
-            "x_e_box" : Array(shape, dtype=np.float32),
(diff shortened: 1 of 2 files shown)


src/py21cmfast/wrapper/outputs.py
```diff
@@ -1286,7 +1286,9 @@ def get_required_input_arrays(self, input_box: OutputStruct) -> list[str]:
         elif isinstance(input_box, PerturbedField):
             required += ["density"]
         elif isinstance(input_box, TsBox):
-            required += ["J_21_LW_box", "x_e_box", "Tk_box"]
+            required += ["Tk_box", "x_e_box"]
+            if self.flag_options.USE_MINI_HALOS:
+                required += ["J_21_LW_box"]
```
daviesje, line 1291: Just check that all of the other structs which require this box have it behind a flag as well

daviesje (approved):

src/py21cmfast/wrapper/outputs.py
jordanflitter: Thanks! I put it behind a flag for `TsBox`, `IonizedBox` and `HaloBox`, I don't think there are more structures that require `J_21_LW_box`.

PR #482, opened 2025-04-02
Throw errmsg if desired quantities don't match inputs

jordanflitter:
This PR follows the previous PRs (#478, #479, #481). If the user asks for a global/lightcone quantity that doesn't match with the input (such that the coeval box doesn't contain the relevant fields), an error is thrown with the appropriate message. This could be implemented more nicely, but it works.


jordanflitter:
> I think this will be better placed as validators in the `Lightconer` (for lightcone_quantities) and `LightCone` (for global_quantities) classes.

Yeah, this is a good idea for global_quantities, however `Lightconer` (unlike `LightCone`) doesn't receive an `InputParameters` object (or its subclasses) as an input.

src/py21cmfast/drivers/lightcone.py
```diff
@@ -321,6 +321,40 @@ def compute_rsds(self, n_subcells: int = 4, fname: str | Path | None = None):
         return tb_with_rsds
 
 
+def check_run_lightcone_inputs(**kwargs):
+    inputs = kwargs["inputs"]
```
steven-murray, line 325: This is fine and will work, but I think it will be less work for us in the future if we do something like the following:

```python
def _check_desired_arrays_exist(desired_arrays: list[str], inputs: InputParameters):
    possible_outputs = [
        InitialConditions.new(inputs),
        PerturbedField.new(inputs, z=0), # these are just dummy/empty outputs, so no need for proper redshift
        TsBox.new(inputs, z=0),
        ...., #etc
    ]
    for name in desired_arrays:
        exists = False
        for output in possible_outputs:
            if name in output.arrays:
                exists = True
                break
        if not exists:
            raise ValueError(f"You asked for {name} but it is not computed for the inputs: {inputs}")
```

steven-murray: Then you would call this function once each for the lightcone quantities and global quantities.

steven-murray (commented):
Thanks @jordanflitter, this is a good addition, but I have a suggestion to make it more generic.

src/py21cmfast/drivers/lightcone.py
```diff
@@ -321,6 +329,27 @@ def compute_rsds(self, n_subcells: int = 4, fname: str | Path | None = None):
         return tb_with_rsds
 
 
+def _check_desired_arrays_exist(desired_arrays: list[str], inputs: InputParameters):
+    possible_outputs = [
+        InitialConditions.new(inputs),
```
steven-murray, line 334: Should `PerturbedHaloField` be in here as well?

steven-murray (commented):
Thanks @jordanflitter. It looks like one of the tests is breaking due to requesting a non-available array, so we should fix that!

src/py21cmfast/drivers/lightcone.py
jordanflitter: I don't think so, I think that fields from halo catalogs cannot be used as global/lightcone quantities, as they are not part of the coeval box fields.

jordanflitter:
> Thanks @jordanflitter. It looks like one of the tests is breaking due to requesting a non-available array, so we should fix that!
Yeah, I see in the logs that "XFAIL tests/test_c_interpolation_tables.py::test_Nion_conditional_tables[FFCOLL-PS..."
I wasn't able to understand from that what specifically caused these tests to fail.

daviesje (approved):

PR #483, opened 2025-04-02
Add benchmarking, fix bug that copied all arrays on compute()

daviesje:
This PR achieves two things:

#### Add Benchmarking:
- places the `test_segfault` calls into `pytest-benchmark` to do automatic benchmarking on a range of lightcone/coeval runs
- Since these tests can be slow, we only do one iteration of each
- (todo) use a github workflow to cache previous results and ensure that we aren't increasing demand too much

#### Fix a numpy casting bug:
- numpy.ndarray.astype copies arrays by default, I have switched this off in `Array.with_value` by setting `copy=False` so that we only copy when the dtypes are different
- I have added a new array evolver, `Array.computed()` for when we want to mark an array as computed, but throw an error instead of silently copying when the types are different, this is now used in the `OutputStruct._compute()` method and in `XraySourceBox`.


steven-murray (approved):
Thanks!

Issue #484, opened 2025-04-02
[Feature Req.] lowres_vcb should exist only if USE_MINI_HALOS=True and FIX_VCB_AVG=False


jordanflitter:
In the spirit of PRs #478, #479, #481, I tried to implement the following logic: `lowres_vcb` exists only if `USE_MINI_HALOS=True` and `FIX_VCB_AVG=False`.

This would not only save memory but also runtime: in `ComputeInitialConditions`, `lowres_vcb` is still computed even if `USE_MINI_HALOS=False` or `FIX_VCB_AVG=True` (though it is not computed if `USE_RELATIVE_VELOCITIES=False`). So even if this (lowres) box is not used in the calculation, 3 (!) highres boxes are computed for no purpose.

Fixing this however requires `ComputeInitialConditions` to receive `flag_options` since this is the structure that holds the said flags. This is where I failed...

jordanflitter:
Issue solved in PR #486 and can therefore be closed.

Note: `USE_RELATIVE_VELOCITIES` should probably be part of `flag_options` than rather `user_params`. The solution to this PR brings us one step closer to this, however `USE_RELATIVE_VELOCITIES` is also used throughout `cosmology.c` where `flag_options` is missing. This will have to wait for a future PR.

PR #486, opened 2025-04-02
lowres_vcb exists only if USE_MINI_HALOS=True and FIX_VCB_AVG=False

jordanflitter:
This PR solves issue #484. It reduces the used memory and improves (slightly) runtime when either `USE_MINI_HALOS=False` or `FIX_VCB_AVG=True`.


steven-murray (commented):
Hmmm, this one I can't approve. Unfortunately there's some subtlety here, because we don't want to make `InitialConditions` dependent on `FlagOptions`. Currently, it is independent of `FlagOptions` so that we can cache a particular `InitialConditions` and re-use it for different flag options (and astro params). Making it dependent would mean invalidating a lot of cache and lead to worse performance in some critical applications (like MCMC).

PR #488, opened 2025-04-03
Add base benchmark

steven-murray:
Having a second go


daviesje (approved):
Just one question: does Bencher automatically store all the result files for us or do we need to have the previous benchmark somewhere?

PR #490, opened 2025-04-03
Fix caching issues

daviesje:
This fixes a couple of issues we had with caching:

- The path templates needed to be able to handle both the string wildcards, as well as possibly numerical values (hex for hashes, .4f for redshift etc), otherwise the `.format()` call would sometimes break
- It turns out that the python builtin `hash()` adds a random seed per session, making in unusable for hashing our input parameters

I have also moved some of the tests which dealt solely with the `OutputCache` into a new class within `test_caching.py` with some extra tests

### DIFF
--- a/src/py21cmfast/io/caching.py
+++ b/src/py21cmfast/io/caching.py
@@ -63,7 +63,7 @@ class OutputCache:
         }
 
     @classmethod
-    def fill_path_template(
+    def _fill_path_template(
         cls,
         *,
         kind: str | None = None,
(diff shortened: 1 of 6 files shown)


src/py21cmfast/io/caching.py
```diff
@@ -39,25 +42,67 @@ class OutputCache:
...
+            ).hexdigest(),
+            "seed": inputs.random_seed,
+            "zgrid": md5(repr(inputs.node_redshifts).encode()).hexdigest(),
+            "astro_flag": md5(
+                (repr(inputs.astro_params) + repr(inputs.user_params)).encode()
+            ).hexdigest(),
         }
 
+    @classmethod
+    def fill_path_template(
```
steven-murray, line 66: I think we should make this `_fill_path_template` because it's not something that the users should be using.

steven-murray (approved):
Wonderful!

PR #491, opened 2025-04-04
changed upper mass integral limit when HALO_STOCHASTICITY=False

jordanflitter:
This change was made in order to calculate more accurately the emissivity fields in `HaloBox` in the (rare) configuration of `FIXED_HALO_GRIDS=False`, `HALO_STOCHASTICITY=False` and `AVG_BELOW_SAMPLER=True`. Now, the upper mass integral limit for the unresolved halos corresponds to the (highres) cell size, in agreement with the smallest scale in the DEXM algorithm in this configuration.

### DIFF
--- a/src/py21cmfast/src/HaloBox.c
+++ b/src/py21cmfast/src/HaloBox.c
@@ -681,7 +681,7 @@ int ComputeHaloBox(double redshift, UserParams *user_params, CosmoParams *cosmo_
         LOG_DEBUG("Gridding %llu halos...",halos->n_halos);
 
         double M_min = minimum_source_mass(redshift,false,astro_params,flag_options);
-        double M_max = M_MAX_INTEGRAL;
+        double M_max;
         double cell_volume = VOLUME/HII_TOT_NUM_PIXELS;
 
         double turnovers[3];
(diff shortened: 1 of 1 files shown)


src/py21cmfast/src/HaloBox.c
```diff
@@ -681,7 +681,7 @@ int ComputeHaloBox(double redshift, UserParams *user_params, CosmoParams *cosmo_
         LOG_DEBUG("Gridding %llu halos...",halos->n_halos);
 
         double M_min = minimum_source_mass(redshift,false,astro_params,flag_options);
-        double M_max = M_MAX_INTEGRAL;
+        double M_max;
```
daviesje, line 684: This should probably be named to something that makes it clear it's the maximum mass of the *integrated* component of the halo boxes, so that nobody can confuse it with the maximum halo mass overall

src/py21cmfast/src/HaloBox.c
```diff
@@ -699,12 +699,19 @@ int ComputeHaloBox(double redshift, UserParams *user_params, CosmoParams *cosmo_
...
         }
         else{
             //set below-resolution properties
             if(user_params->AVG_BELOW_SAMPLER && M_min < user_params->SAMPLER_MIN_MASS){
-                set_fixed_grids(M_min, user_params->SAMPLER_MIN_MASS, ini_boxes,
+                if (flag_options->HALO_STOCHASTICITY){
+                    M_max = user_params->SAMPLER_MIN_MASS;
+                }
+                else {
+                    M_max = RtoM(L_FACTOR*user_params->BOX_LEN/user_params->DIM);
```
daviesje, line 712: There's one edge case that we're still missing here I think, which is when the cell size is so ridiculously small that it goes below the minimum source mass. in which case we can just skip the integrated component entirely. While it *does* get caught in the integrals themselves we probably don't want to go through the whole setup.

daviesje (commented):
good to go after a couple minor comments and tests passing

jordanflitter:
Thanks for the suggestions @daviesje. I renamed `M_max` to `M_max_integral` and I split the `if (AVG_BELOW_SAMPLER && Min < M_max_integral)` into two separate conditions.

daviesje (approved):

PR #492, opened 2025-04-04
Existence of output initial velocities depend on PERTURB_ON_HIGH_RES

jordanflitter:
This change was made to reduce used memory.
If `PERTURB_ON_HIGH_RES = True`, then only hires initial velocities exist.
If `PERTURB_ON_HIGH_RES = False`, then only lowres initial velocities exist. However, note that if `PERTURB_ALGORITHM == "2LPT"`, we must also have 2LPT hires initial velocities (even if we don't use them as an output), as these boxes are always used as an intermediate storage place for the 2LPT calculation, regardless the value of `PERTURB_ON_HIGH_RES`.

NOTE: In the future, we may come up with a similar logic when `PERTURB_ALGORITHM=LINEAR`, in this case the initial velocities are not used at all for the perturbed density and velocity fields. However, these initial velocities can be used in `PerturbHaloField` so they should still exist and be computed if `USE_HALO_FIELD=True` (and `FIXED_HALO_GRIDS=False`). So a future PR should be something like if `PERTURB_ALGORITHM=LINEAR` and (`USE_HALO_FIELD=False` or `FIXED_HALO_GRIDS=True`)->don't bother calculating and saving initial velocities.

### DIFF
--- a/src/py21cmfast/wrapper/outputs.py
+++ b/src/py21cmfast/wrapper/outputs.py
@@ -485,13 +485,13 @@ class InitialConditions(OutputStruct):
     _compat_hash = _HashType.user_cosmo
 
     lowres_density = _arrayfield()
-    lowres_vx = _arrayfield()
-    lowres_vy = _arrayfield()
-    lowres_vz = _arrayfield()
+    lowres_vx = _arrayfield(optional=True)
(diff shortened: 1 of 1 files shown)


daviesje (commented):
This looks right to me but we should get the tests passing

PR #496, opened 2025-04-05
Addition of clang-format to pre-commit hooks, preparation for clang-tidy

daviesje:
This PR will standardise the formatting of the repository, using the Google formatting style with 4-space indents and a column limit of 100.

in-place formatting is done using https://github.com/pocc/pre-commit-hooks. which has several formatters and linters available in case we wish to add these in future.

I also ran clang-tidy on the backend, and there was far too much to address in a reasonable time. I fixed some of the simpler ones but it's not ready to be a part of the pre-commit hooks.

I left the clang-format automatic fixes in one commit (the third one), with no other changes. So when reviewing it's probably best to skip that one.

There are two small issues with the pre-commit I still need to address:
- The github pre-commit doesn't install the dev requirements, so it fails with the clang-format call
- Verify that we actually want the new versions of ruff and pre-commit in the config


steven-murray (approved):
Thanks @daviesje this style looks good!

PR #497, opened 2025-04-07
Parameter reorganisation

daviesje:
This PR re-organises our input parameter structures to have a clearer separation between numerical values vs flags, as well as parameters which only affect the astrophysics and those that affect the matter field.

- `CosmoParams` have been left alone, and the rest of the parameters were split into `MatterParams`, `MatterFlags`, `AstroParams` and `AstroFlags`.
- `-Flags` are always boolean or enums, whereas `-Params` are numerical values
- `Matter-` (and `CosmoParams) affect the matter field, and require a full regeneration when changed, `Astro-` only affect the astrophysics, and so we can use the same ICs, density fields and halo catalogues when they change.
- The broadcasting of the parameter structures has been moved to the frontend to stop the existance of multiple versions of a parameter structure at one time.

I still need to pass over some of the docstrings, however the test currently pass locally.


CHANGELOG.rst
```diff
@@ -98,7 +98,7 @@ Added
   scenarios in line with current observations (#305)
 * Add in an initialisation check for the photon conservation to address some issues
   arising for early EOR histories (#311)
-* Added ``NON_CUBIC_FACTOR`` to ``UserParams`` to allow for non-cubic coeval boxes (#289)
+* Added ``NON_CUBIC_FACTOR`` to ``MatterParams`` to allow for non-cubic coeval boxes (#289)
```
steven-murray, line 101: We should probable leave the changelog unchanged.

README.rst
```diff
@@ -81,7 +81,7 @@ and plotting an image of a slice through it::
     >>> import py21cmfast as p21c
     >>> coeval = p21c.run_coeval(
     >>>     redshift=8.0,
-    >>>     user_params={'HII_DIM': 50, "USE_INTERPOLATION_TABLES": False}
+    >>>     matter_params={'HII_DIM': 50, "USE_INTERPOLATION_TABLES": False}
```
steven-murray, line 84: This seems like it should be `matter_flags` or `matter_options`?

joss-paper/paper.md
```diff
@@ -178,12 +178,12 @@ import py21cmfast as p21c
 lightcone = p21c.run_lightcone(
     redshift=6.0,              # Minimum redshift of lightcone
     max_redshift=30.0,
-    user_params={
+    matter_params={
```
steven-murray, line 181: probably should leave paper.md file unchanged as well

src/py21cmfast/drivers/_param_config.py
```diff
@@ -204,6 +204,9 @@ def _get_inputs(kwargs: dict[str, Any]) -> InputParameters:
         example) and this is fine. So, here we return the inputs of the "most advanced"
         OutputStruct that is given.
         """
+        # TODO: Figure out what happens when the first astro box (HaloBox, IonizedBox or SpinTemp)
+        #   is passed inputs=None, does it pull the astro params from the matter boxes, if so that's bad
+        #   We probably shouldn't ever store/save the astro params in the matter boxes
```
steven-murray, line 209: Let's turn this into an issue and think about solving it, because I think it's an important point.

src/py21cmfast/drivers/single_field.py
```diff
@@ -215,10 +222,13 @@ def compute_halo_grid(
     ----------
     initial_conditions : :class:`~InitialConditions`
         The initial conditions of the run.
-    perturbed_halo_list: :class:`~PerturbHaloField` or None, optional
+    inputs : :class:`~InputParameters`
+        The input parameters specifying the run. Since this will be the first box
+        to use the astro params/flags, if it is computed this is always needed.
```
steven-murray, line 227: This is not how I intended it -- the astro params/flags are gotten from the initial conditions if they're not given explicitly here, because in almost every case you are setting this up via `run_coeval` or `run_lightcone` where the full inputs are defined before anything is run.

src/py21cmfast/io/caching.py
```diff
@@ -575,3 +579,18 @@ def noloop(cls):
             halo_field=False,
             xray_source_box=False,
         )
+
+    @classmethod
+    def last_step_only(cls):
+        """Generate a CacheConfig where only boxes needed from more than one step away are cached."""
```
steven-murray, line 585: can you expand this docstring a bit? I'm not sure I understand the purpose

src/py21cmfast/io/h5.py
```diff
@@ -307,27 +309,35 @@ def read_inputs(
...
+        "flag_options",
+        "global_params",
+    ]
 
-            subgrp = group[kfile]
-            dct = dict(subgrp.attrs)
-            kwargs[k] = kls.from_subdict(dct, safe=safe)
-        else:
-            kwargs[k] = group.attrs[kfile]
+    # Read the input parameter dictionaries from file.
```
steven-murray, line 327: Do we have a test of whether this works? We should definitely have one before we release.

steven-murray (commented):
Thanks @daviesje, this is looking good.

Some broader thoughts:

1. I think `MatterParams` should be `MatterFlags` -- almost all of the "parameters" in there are boolean or integers that aren't physical parameters. 
2. Since there are many more integer-style options in what I'm saying should be `MatterFlags` it might actually be better to call it `MatterOptions`, which is a little more general and sounds like it would include any boolean or integer style options for how to run the simulation. Probably we can rename `AstroFlags` to `AstroOptions` as well to be consistent.
3. It's actually worth considering `CosmoOptions` instead of `MatterOptions` as well, which would be a little more symmetrical (so we'd have CosmoParams/CosmoOptions and AstroParams/AstroOptions)
4. OK as I read further I realized you also have `MatterFlags` defined, so amend points 1-3 accordingly. Unfortunately I have to go now, so will have to think through this more a bit later.
5. We should add an "Upgrading from v3 to v4" doc somewhere that explains the biggest differences from the user-side, and these updates should definitely be in there.

src/py21cmfast/drivers/_param_config.py
daviesje: I forgot that we always require full parameter specification, so this is a minor issue. There are some edge cases since AstroParams are saved to the InitialConditions structs, which means they will just get overwritten with the latest AstroParams (same actual data) and a user might accidentally load in an unexpected parameter set. This should always get caught in the consistency checks though.

src/py21cmfast/drivers/single_field.py
daviesje: Fixed, see above

src/py21cmfast/io/caching.py
daviesje: This is the minimum caching setup which allows purging of objects which are needed at multiple redshifts. PerturbedField and PertrubedHaloField, and HaloBox need to be kept in memory at all redshifts if they are not cached, unlike the rest which are generally only needed at the current and last step.

src/py21cmfast/io/h5.py
daviesje: Due to how much the parameters have changed I would expect this to not produce a complete parameter set most of the time. Parameter names in v3 which are not in v4 will be ignored and parameter names in v4 not in v3 will be set to defaults.

As far as verifying that behaviour, I guess I could hold v3 file somewhere for testing? I'm not sure how to do it otherwise

Issue #472, opened 2025-03-24
[Feature Req.] Changing the flag logic and implementation for RSDs


jordanflitter:
The current logic for accounting for RSDs depends on 2 different flags:
1. `APPLY_RSDS`: whether to apply RSDs in rectilinear lightcones. The RSD calculation (linear and non-linear) is done in C. Cannot be true for angular lightcones.
2. `get_los_velocity` (this is not a standard flag in flag_options, but rather an input to a Lightconer object): whether to apply RSDs in angular lightcones. The RSD calculation (linear and non-linear) is done in python.

A simpler logic would involve only one flag, `APPLY_RSDS`, for both purposes, without the need to have `get_los_velocity`. This can be implemented by moving the `APPLY_RSDS` section (for the linear correction) from `ComputeBrightnessTemp` to the wrapper, as done currently for angular lightcones.

The non-linear correction, `SUBCELL_RSD`, could also be done in the same place (C or python) for both types of lightcones.

steven-murray:
Closed by #498

PR #498, opened 2025-04-09
reorginization of brightness temp and RSD calculations

jordanflitter:
The goal of this PR is to address issue #472. Instead of having two flags `APPLY_RSDS` and `get_los_velocity` for rectilinear and angular lightcones, respectively, now `APPLY_RSDS` controls whether we account for RSDs, regardless the selected lightcone. For both types of lightcones, the RSD calculation (linear and non-linear) is done at the wrapper level via the method `compute_rsds`.

Some general comments about this PR:

- It is important to realize that the effect of RSDs cannot be done now on the coeval boxes, but rather on the lightcone box, as this is a "line-of-sight effect" and requires evaluating the velocity gradient at different redshifts along the line-of-sight. This also means that `global_brightness_temp` (which is evaluated from taking the mean of the coevals) is potentially different than the true global signal, which is evaluated by taking the mean of the lightcone box at each cross-section along the los axis (and includes also the effect of RSDs). 
-  While `brightness_temp` can be written to disk, I think that `brightness_temp_with_rsds` is currently cannot be written to disk. Also, both lightcones do not account for suppression by $\tau_\mathrm{reio}$. This should be fixed in a future PR.

The following modifications were made in order to address issue #472:

- `get_los_velocity` has been removed as a possible input to `Lightconer`.
- An optional box of `tau_21` was added to `BrightnessTemp`. This box exists only if `APPLY_RSDS` and `USE_TS_FLUCT` are True.
- `validate_options` has been modified such that: (1) For `RectilinearLightconer`, if `APPLY_RSDS` then`los_velocity` is added to `self.quantities`. (2) For AngularLightconer, if `APPLY_RSDS` then `los_velocity` is added to `self.quantities`, and  the same error as before is thrown if `KEEP_3D_VELOCITIES` is False. In both cases, if `APPLY_RSDS` and `USE_TS_FLUCT` are True, `tau_21` is also included in `self.quantities`.
- Because of the above bullet, the code lines that have to do with `get_los_velocity` in `setup_lightcone_instance` have been removed.
- In `ComputeBrightnessTemp` (at the C backend) the entire `APPLY_RSDS` section was removed, since it is now implemented in the wrapper, under `compute_rsds` (for both types of lightcones!). This allows to remove the functions and variables that have to do with RSD so now `ComputeBrightnessTemp` is a very simple function. Because of that, subcell_rsds.c and subcell_rsds.h have been removed from the repository (and also from SOURCES.txt for compiling the code. I pushed to modified file though I'm not if it is required). In addition, the output box `brightness_temp` depends only on `USE_TS_FLUCT`. If `USE_TS_FLUCT=False`: it is the brightness temperature under the saturation limit, $T_{21}\propto\tau_{21}$. If `USE_TS_FLUCT=True`: it is the brightness temperature when the saturation limit is relaxed, $T_{21}\propto1-\exp(-\tau_{21})$. If `USE_TS_FLUCT` and `APPLY_RSDS` are True then the values of the optional `tau_21` box are filled.
- In `make_lightcone_slices`, the `get_los_velocity` condition has been replaced with "if `los_velocity` in `self.quantities`" condition. Some extra minor modifications were made in order to support having a `los_velocity` lightcone box for a rectilinear lightcone.
- `compute_rsds` is now a method of all `Lightcone` objects (not necessarily `AngularLightcone` objects as was done previously) and the condition in `_run_lightcone_from_perturbed_fields` has been changed to be a condition on `APPLY_RSDS` (instead of a condition on `AngularLightcone` and `get_los_velocity`).
- In `compute_rsds`, if `USE_TS_FLUCT=True`, the linear RSD is applied on the stored lightcone box of `tau_21` and the brightness temperature is corrected accordingly. In addition, `cosmotile.apply_rsds` is called only if `SUBCELL_RSD` is True. Since this function expects to receive an `AngularLightcone` object, if a rectilinear lightcone is considered, it is transformed to an angular-like lightcone. Once `cosmotile.apply_rsds` returns, if a rectilinear lightcone is considered, the output is transformed again to a rectilinear-like lightcone.
- Lastly, in order to avoid storing lightcone boxes that the user didn’t specify, every lightcone quantity that is not from the original list is popped out of the dictionary. The code here is not written in the most general way since `brightness_temp_with_rsds` is an optional lightcone box that the user doesn’t specify, but we still want it to be as part of the outputs anyway.

Two more "bonuses" were also included in this PR:

- When `_run_lightcone_from_perturbed_fields` is about to finish, at the last redshift, after `compute_rsds` returns, a multiplication by `$exp(-tau_\mathrm{reio})$` is included, to account for suppression in the signal due to Thomson scattering with charged particles.
- The value of `Tcmb` in Constants.h has been updated to 2.7255 K (instead of 2.728 K as before) to match with the known FIRAS measurement and the rest of the literature.


jordanflitter:
I made a comparison between the output of the new branch vs. the main branch. In order to do that, `compute_rsds` can now be used both for lightcone boxes (as an integral part of `run_lightcone` if `APPLY_RSDS` is True) and for coeval boxes (as a post process of the coeval box).

Some discrepancies have been found. I'm listing them down, see also plots in the attached PDF. I also attached here the script I used to make these plots (in the zip file). The parameters used for the comparison are basically the ones found in produce_integration_test_data.py, except for setting `USE_TS_FLUCTS` on True, as well as reducing `ZPRIME_STEP_FACTOR` to 1.02. 
[Untitled.pdf](https://github.com/user-attachments/files/20211945/Untitled.pdf)
[Compare_outputs.zip](https://github.com/user-attachments/files/20211282/Compare_outputs.zip)

1. There is a discrepancy even when `APPLY_RSDS` and `SUBCELL_RSD` are False. This is because the full exponential form of $1-\mathrm{exp}(-\tau_{21})$ is used in the main branch only when `APPLY_RSDS` and `SUBCELL_RSD` are both True, while in other configurations the Taylor expansion $1-\mathrm{exp}(-\tau_{21})\approx\tau_{21}$ is used. This **BUG in the main branch** causes an error in the minimum of the global signal, where the spin temperature is at minimum (and $\tau_{21}$ is at maximum). The Taylor expansion should only be used when `USE_TS_FLUCTS` is False (i.e. we don't have a spin temperature, and assume indeed that $\tau_{21}\ll1$).
2. Another discrepancy arises due to an incorrect minus sign in [here](https://github.com/21cmfast/21cmFAST/blob/main/src/py21cmfast/drivers/lightcone.py#L292). This is **another BUG in the main branch**. Furthermore, the C code in the main branch computes the velocity gradient by computing the spectral derivative (i.e. multiplication by $ik$ in Fourier space) vs. the central derivative that is used by `numpy.gradient` and the python code. This seems not to cause too much of a difference.
3. The C code shifts subcells while assuming periodic boundary conditions, whereas the python code in `compute_rsds` didn't. This is **not** a bug in the main branch, as `compute_rsds` was originally designed to be applied on non-periodic lightcone boxes. Still, in order to compare apples to apples, I enabled an option to assume periodicity in the python code. This feature (periodicity) seems to have non-negligible impact on the power spectrum.
4. Lastly, there is **another BUG in the main branch** due to not clipping the velocity gradient when `SUBCELL_RSDS` is True, as can be seen [here](https://github.com/21cmfast/21cmFAST/blob/main/src/py21cmfast/src/BrightnessTemperatureBox.c#L199). The absence of clipping in this scenario can amplify the power spectrum by an order of magnitude.

When all of the above is fixed, there is a very good agreement between the C code and the python code.

**Note**: In order to make a fair comparison, I disabled the multiplication of the brightness temperature by $\exp(-\tau_\mathrm{reio})$ (though it should still be there, in theory).

jordanflitter:
In order to check the consistency of the RSD codes (C and python) I did the following:

1. I refactored the C code (subcell_rsds.c) so it could be called by a test function.
2. I wrote a script in python that makes some tests, like (1) comparison of the boxes after the RSD has been applied (both in C and in python), (2) makes a sanity check of the sum along the line-of-sight (assuming periodicity, mass should be conserved in the box), (3) makes a sanity check of constant integer displacement (assuming periodicity, output should be a shifted version of the input box).

The necessary files (python and C) are attached in this zip.
[Archive.zip](https://github.com/user-attachments/files/20212204/Archive.zip)

These tests can be included in the future as part of the general acceptance procedure for new PRs. 
**Note**: Since subcell_rsds.c was removed in the new branch, some adjustments are need to be made in order to use the attached files.
**Another note**: I should say that no inconsistencies have been found in the C code of RSDs in subcell_rsds.c.

src/py21cmfast/lightcones.py
```diff
@@ -301,28 +323,127 @@ def construct_los_velocity_lightcone(
...
+                )
+                * self.lc_distances.unit
+            )
+            # Make a new lightconer which is identical to self, but extends to further boundaries!
+            lightconer_rsd = attrs.evolve(self, lc_redshifts=None, lc_distances=lcd_rsd)
+            return lightconer_rsd
+        else:
+            return self
+
+    def find_required_lightcone_limits(
```
steven-murray, line 371: Let's make a specific unit test for this function

steven-murray (commented):
Last thing to do is to write up a short tutorial.

PR #500, opened 2025-04-12
Field renaming

daviesje:
This simply renames output fields to be more intuitive. I also have some fixes to the integration tests which bring us (I think) to the v3 defaults which could be elusive at times.

A summary of the field renaming can be found in the `updates_from_v3.rst` doc


steven-murray (approved):
Thanks @daviesje this is what the world needed

steven-murray (approved):
All looks good to me!

PR #502, opened 2025-04-14
V4

daviesje:
This PR will mark the change to version 4.0.0. It includes a stochastic sampler of the conditional halo mass function allowing us to output lightcones of galaxy properties and self-consistently sampled galaxy catalogues. There are also large overhauls of the input/output structures, caching and refactoring of the C backend.


.qodo/history.sqlite
steven-murray, line 1: Let's add qodo to gitignore

pyproject.toml
```diff
@@ -1,20 +1,97 @@
 [build-system]
-requires = ["setuptools>=42", "wheel", "setuptools_scm[toml]>=3.4"]
+# unpin wheel when https://github.com/pypa/wheel/issues/659 is resolved
```
steven-murray, line 2: This can be resolved now by using higher version of setup tools and removing the wheel dep

pyproject.toml
```diff
@@ -1,20 +1,97 @@
...
-  | \.tox
-  | \.venv
-  | _build
-  | build
-  | dist
-)/
-'''
+[tool.ruff]
+line-length = 88
+target-version = "py310"
```
steven-murray, line 10: Go to 3.11

steven-murray (approved):

PR #512, opened 2025-05-06
Combine integration/segfault tests

daviesje:
We had two test modules which ran lightcones/coevals with a bunch of parameters and checked their outputs. This PR combines them into one and replaces the parameter sets with newer ones which make more sense.

We assert that the run completes and that we get finite brightness temperatures. We also check if anything has changed and print information on runs that are not the same as stored, but we do not assert that they are the same like the previous integration tests.


tests/produce_integration_test_data.py
```diff
@@ -95,138 +81,147 @@
...
             "Z_HEAT_MAX": 25,
             "ZPRIME_STEP_FACTOR": 1.1,
+            "N_THREADS": 4,
+            "USE_RELATIVE_VELOCITIES": True,
+            "POWER_SPECTRUM": "CLASS",
         },
     ],
-    "mdz_and_ts_fluct": [
-        9,
+    "mini_ff": [
```
steven-murray, line 107: Can we use a more descriptive name?

tests/produce_integration_test_data.py
```diff
@@ -357,12 +312,28 @@ def produce_lc_power_spectra(redshift, **kwargs):
...
+
+def produce_lc_power_spectra(redshift, **kwargs):
+    options = get_all_options_struct(redshift, lc=True, **kwargs)
+    print("----- OPTIONS USED -----")
+    print(options)
+    print("------------------------")
+
+    # NOTE: this is here only so that we get the same answer as previous versions,
+    #       which have a bug where the max_redshift gets set higher than it needs to be.
+    node_z = options["inputs"].node_redshifts
```
steven-murray, line 331: Which previous version are we trying to match here? We should just set the standard in this PR.

steven-murray (commented):
Thanks @daviesje, this is definitely an improvement! Other than 2 small comments, I have a few thoughts:

1. Perhaps we should only output plots when there is a "failure", like we only print when there is a failure.
2. Instead of printing the failures, we could raise them as warnings. We can use the `-Werror` switch on the CLI to make them then fail the tests.
3. We should update the plots that are made. That can be done either here or in another PR, but I think it would be very helpful if we plotted a transverse slice (comparing old to new), a lightcone slice (for a lightcone), and then also the power spectra / global quantities. Since these plots will now be a first point of entry into understanding differences.

tests/produce_integration_test_data.py
daviesje: I fixed the behaviour but forgot to delete the comment, it's done now

daviesje:
1 and 2 have been done, but how do we want to handle the plot flag in the workflow and the archiving of the test plots? currently we plot everything and archive everything.

As for 3, if we're only plotting on failed runs I suppose we can make them a lot more detailed, with slices and so on, but I'm not sure how informative they will be since they're all very low-res and high redshift. We could save some cell values and do some histograms and scatter plots to show changes explicitly?

steven-murray:
I think these are good questions, let's maybe have a face-to-face discussion to design the plot? The overall purpose is to draw attention to changes, and to visualize them in a way that most easily shows what might have gone wrong. Of course, if something does go clearly wrong, we can investigate manually by running the case again on main and the new branch to compare in detail.

daviesje:
[tests.test_integration_features.py--test_power_spectra_coeval[inhomo_ts].pdf](https://github.com/user-attachments/files/20100577/tests.test_integration_features.py--test_power_spectra_coeval.inhomo_ts.pdf)
[tests.test_integration_features.py--test_power_spectra_lightcone[inhomo_ts].pdf](https://github.com/user-attachments/files/20100578/tests.test_integration_features.py--test_power_spectra_lightcone.inhomo_ts.pdf)

steven-murray (approved):
This looks good, thanks @daviesje!

PR #513, opened 2025-05-06
update readme, make reference printing function

daviesje:
This updates the readme to include the v4 paper and adds a function which prints out relevant papers depending on input parameters


steven-murray (approved):
Beautiful, thanks!

PR #515, opened 2025-05-13
run benchmarks on main

daviesje:
oops

### DIFF
--- a/.github/workflows/base_benchmarks.yml
+++ b/.github/workflows/base_benchmarks.yml
@@ -1,6 +1,6 @@
 on:
   push:
-    branches: v4-prep
+    branches: main
 
 jobs:
   benchmark_base_branch:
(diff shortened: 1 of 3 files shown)


steven-murray:
I think maybe you want just `on: push` rather than specifying a branch. Specifying `main` as the branch I think will only trigger once the PR is actually merged to main (i.e. when there is a push to the main branch)

daviesje:
Do we really want to run the benchmark on every push to every branch though?

steven-murray:
Yeah, maybe not. On the other hand, as it stands, we won't know about performance problems until after a PR is merged

daviesje:
I've set it up to run on PRs to main instead of pushes. This should be better

daviesje:
Perhaps we should set up a bencher branch protection as well

steven-murray:
Yep, I like this. Is there a way to get Bencher to report directly on a PR? Like a bot that makes a comment or something? Because the action output itself is too verbose.

I don't think we should block on performance -- it should just be there to prompt us to consider it.

Issue #448, opened 2024-12-09
[BUG] "velocity_z" pointer field error in tutorial lightcones.ipynb


raxquax123:
**Describe the bug:**
Attempting to run the lightcones.ipynb code provided in the tutorial Jupyter notebooks (or in the documentation https://21cmfast.readthedocs.io/en/latest/tutorials/lightcones.html) using the Python-wrapped code in the Github master branch returns the following error when calling the run_lightcone function:

> Key velocity_z in PerturbedField(UserParams(BOX_LEN:512, DIM:384, FAST_FCOLL_TABLES:False, HII_DIM:128, HMF:1, KEEP_3D_VELOCITIES:True, MINIMIZE_MEMORY:False, NON_CUBIC_FACTOR:1, NO_RNG:False, N_THREADS:1, PERTURB_ON_HIGH_RES:False, POWER_SPECTRUM:0, USE_2LPT:True, USE_FFTW_WISDOM:False, USE_INTERPOLATION_TABLES:True, USE_RELATIVE_VELOCITIES:False);
> 	CosmoParams(OMb:0.04897, OMm:0.3096, POWER_INDEX:0.9665, SIGMA_8:0.8102, hlittle:0.6766);
> 	random_seed:315828138004;
> 	redshift:12.124841861245521) not a defined pointer field in C.

**To Reproduce:**
Run the Jupyter notebook found at docs/tutorials/lightcones.ipynb using the current Python-wrapped 21cmFAST version from the Github master branch.

**Details:**
- Python version: 3.11.10
- OS: macOS 14.6.1
- Package: I’m using the Python-wrapped code from the master branch of the Github (as the version installed through conda does not include the new lightcone.py file, whose functions are called in the Jupyter notebook / readthedocs page). 

**Additional context**
If I swap out the wrapper.py and outputs.py files for the conda-installed ones, then copy across these new lightcone functions introduced by the Github version into the two conda-based files, run_lightcone will run (although there are key errors later on when introducing RSDs). 
I doubled the number of cells in each dimension compared to the default, but the resolution is the same. Using the default HII_DIM and BOX_LEN produces the same error. 
I’m not sure if it’s an issue on my part with e.g. dependencies, or whether there is a bug!

steven-murray:
[...]

PR #521, opened 2025-06-03
bug fixes to last cell of lightcones.ipynb

raxquax123:


docs/tutorials/lightcones.ipynb
```diff
@@ -1453,7 +1453,7 @@
     "ax.set_rorigin(0)\n",
     "H, D = np.meshgrid(lonh, ang_lcn_horiz.lc_distances.value)\n",
     "img = ax.pcolormesh(\n",
-    "    H, D, lc_horiz.brightness_temp.T, edgecolors=\"face\", cmap=cmap, vmin=vmin, vmax=vmax\n",
+    "    np.transpose(H), np.transpose(D), lc_horiz.lightcones['brightness_temp'][:-1,:-1], edgecolors=\"face\", cmap=cmap, vmin=vmin, vmax=vmax\n",
```
daviesje, line 1456: Both the old and new lines here seem strange to me, I don't think we should need any transposition/reversing if we set up the coordinates correctly. @steven-murray How specifically are the angular lightcone axes and coordinates set up? I believe we could just pass `indexing='ij'` to the `np.meshgrid` call above so it has the correct dimensions in the first place.

daviesje (commented):
Hi @raxquax123, thanks for the PR. In future it would be helpful to include a brief description about what the changes are and what was wrong beforehand. I think some minor changes are needed for clarity but otherwise this fixes some issues with the angular lightcone plotting.

raxquax123:
Hi @daviesje - thank you! Sorry about the lack of description - this was my first PR ever haha - I'll make sure to add a description in the future! This PR was based off of https://github.com/21cmfast/21cmFAST/issues/448, which has a more detailed description of each of the changes - I'll go through each comment now with the reasoning!

docs/tutorials/lightcones.ipynb
```diff
@@ -1390,10 +1390,10 @@
     "        ],\n",
     "    )\n",
     "    try:\n",
-    "        plt.register_cmap(cmap=eor_colour)\n",
+    "        plt.colormaps.register(cmap=eor_colour)\n",
     "    except ValueError:\n",
-    "        cm.unregister_cmap(name)\n",
-    "        plt.register_cmap(cmap=eor_colour)\n",
+    "        plt.colormaps.unregister(name)\n",
+    "        plt.colormaps.register(cmap=eor_colour)\n",
```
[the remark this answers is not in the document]
raxquax123: No worries!

docs/tutorials/lightcones.ipynb
raxquax123: The original line returned this error:

>TypeError Traceback (most recent call last)
>---> [24] img = ax.pcolormesh(H, D, lc_horiz.lightcones['brightness_temp'], edgecolors="face", cmap=cmap, vmin=vmin, vmax=vmax)
>
>TypeError: Dimensions of C (3000, 1573) should be one smaller than X(3000) and Y(1573) while using shading='flat' see help(pcolormesh)

so I removed the last indices of the brightness temperature array (as I was a little nervous to add anything to H and D instead haha). However, this still returned a TypeError, so I checked the shape of H, D, and the brightness temperature, and H and D had transposed dimensions when compared to the brightness temperature, so I transposed them to be the same, and that worked! I definitely should have changed the coordinates at the generation rather than in the plotting, though - sorry about that!

steven-murray: Yeah, I think using `indexing='ij'` should fix this

raxquax123:
No worries at all @steven-murray, and thank you! One thing - I realised that I had made another change that I had forgotten about! It was the USE_TS_FLUCT flag (changed from True to False), as the maximum node redshift recognised was at z=12 (above z_heat_max) and so the cell did not run 'as is', and I changed this flag as a quick 'bandaid' solution! Changing it back, I think I've realised the source of the original issue, which is I think the new redshift range needs to be stated within the evolve_input_structs function, as defining the new redshifts in the lightconer only changes the limits of the displayed lightcone (but doesn't simulate them). Should I submit a separate PR, or include the fix in this one? (Also - apologies for the delay - I'm currently running the notebook to double-check that this fixes things, but I've got four things running on my laptop right now and hence it's rather slow!)

steven-murray:
@raxquax123 I'm not sure I totally understand the problem with the redshifts here. So, indeed, the redshifts in the "Lightconer" merely control which redshifts end up in the lightcone, not which redshifts are actually simulated. To change the redshifts actually simulate, you must change the `node_redshifts` attribute of the `InputParameters`. These `node_redshifts` --- for a simulation with `USE_TS_FLUCT=True` must extend beyond `Z_HEAT_MAX`. 

So, what was the error that you were getting that required you to change the `USE_TS_FLUCT` parameter?

raxquax123:
Hi @steven-murray - no worries at all, I love bug fixing haha! The brightness temperature array _is_ the same size as H and D, but the matplotlib colormesh plotting function requires the coordinate arrays to be one larger than the value array (as I think they are effective 'bin edges')!

steven-murray:
Ahh, I see!

steven-murray (approved):

PR #525, opened 2025-06-12
Bug fixes to mini-halos.ipynb and relative-velocities.ipynb

raxquax123:
Hello! I hope this isn’t presumptuous, but I’ve also done a bit of testing/updating on the other tutorial notebooks!
All ran straight through with no changes apart from mini-halos.ipynb and relative-velocities.ipynb.

It was mostly updating the calling parameters of functions, similarly to lightcones.ipynb, but there were a couple things to note that I’m not sure whether they need to be separately addressed (or not): 
- It looks like L_X_MINI and F_H2_SHIELD are set to their max value already - multiplying them by factors > 1 returns an error.
    - I’ve ‘fixed’(?) this by simply not allowing the multiplying factor to be > 1 for these parameters. Not sure if this is just a bandaid solution, but everything ran smoothly upon doing so!
- Setting inputs_acg = inputs.evolve_input_structs(USE_MINI_HALOS=False) appears to be incompatible with simulating J_21_LW - it returns a ValueError: 
  >ValueError: You asked for J_21_LW but it is not computed for the inputs
- but this doesn’t appear for the fiducial lightcone.
- initial_conditions - I’ve changed to explicitly have write=True, else it doesn’t write to cache (and can’t be accessed later in runcache)
- ‘L_X_MINI' must be > 0 - I’ve changed multiplicative factor in use-mini-halos.ipynb to 0.0000001 (i.e. small but non-zero).

This is less of a bug and just an aesthetics/ease-of-reading question, but in relative-velocities.ipynb, the cell below the markdown title ’21-cm power spectra’ just output the same plots four times whilst not varying any input parameters, so I’ve changed it to only plot each chunk once! Let me know if I’ve done this erroneously :)

I’ve also re-run all of the cells in relative-velocities.ipynb, as a pre-existing keyboard interrupt message meant that a lot of the figures weren’t generated already! (However, mini-halos.ipynb had all figures pre-generated, so I just tested the code with smaller runs in a separate notebook, then copy-pasted any changes over, and left the original figures in.)


steven-murray:
Hi @raxquax123, sorry it has been a while since you made this PR -- we've been doing a lot of background work on the code, and made some reasonably significant API changes in the meantime. 

I appreciate the changes you've made here, but I think actually we probably want to rethink these notebooks entirely. I think documentation notebooks should be runnable in a Hubble time, so we might need to think about how to best present these models. 

@daviesje what do you think?

Issue #495, opened 2025-04-04
[Feature Req.] PerturbedField.c should be reorganized


jordanflitter:
The code in PerturbedField.c is quite old and should be better organized (in the spirit of the v4 changes that were done in SpinTemperatureBox.c and IonisationBox.c). Instead of having a giant `ComputePerturbField`, we should have several sub-functions like `evolve_linearly`, `compute_displacement_field`, `shift_masses`, `cloud_in_cell`, `downsample_field`, `smooth_density_field`. In addition, code lines are duplicated to account for the two optional values of `PERTURB_ON_HIGH_RES`, this could be rewritten in a more elegant way.

All of the above is nice to have when there is only one field that is being evolved (`delta_m`). However, including `delta_b` and exotic species of dark matter can make this code unnecessarily complicated, as was done in 21cmFirstCLASS.

There is also an issue with how the input object of the initial velocity boxes are modified throughout the calculation by multiplying them with the scale-independent growth factor (SIGF), and the input boxes are then restored at the end of the calculation by dividing with the same factor.
In 21cmFirstCLASS, these multiplications and divisions were done in Fourier space as species are evolved with a scale-dependent growth factor (SDGF). Due to this implementation and small numerical errors in FFT, these errors grew larger with more redshift snapshots. While these errors were still less than 1%, they caused an unphysical covariance matrix behavior in Fisher analysis, where a numerical derivative is used with respect to the varied parameters (I believe that the growing errors in the FFT led to significantly greater errors in the evaluation of the numerical derivatives).
Using the input velocity boxes as an temporary storage place was done in order to save memory space, but this implementation ought to be modified if we wish to incorporate a non-linear scale-dependent growth in the code.

Ideally, a similar cleanup should also be done in InitialConditions.c, though changing PerturbedField.c is of higher priority.

PR #530, opened 2025-07-06
Floating-Point Halo Position

daviesje:
Now that we're using halo catalogues as sources, there's no reason to keep the halo positions as integers. This change sets `halo_coords` fields to `float`, from zero to `BOX_LEN` with the following model:

- DexM Halos are placed in the centre of their `DIM` cells
- Initial sampler halos are placed in a random position within their Lagrangian cell
- Progenitors are placed in a random position within the sphere defined by their descendant Lagrangian masses
- `PerturbHaloField` no longer rounds positions onto `HII_DIM`

Currently we are failing some integration tests and I am working out if this is to be expected or if there are some bugs present


src/py21cmfast/src/PerturbHaloField.c
```diff
@@ -58,6 +58,8 @@ int ComputePerturbHaloField(float redshift, InitialConditions *boxes, HaloField
         growth_factor = dicke(redshift);  // normalized to 1 at z=0
         displacement_factor_2LPT = -(3.0 / 7.0) * growth_factor * growth_factor;  // 2LPT eq. D8
 
+        // TODO: combine/match with PerturbField.c
+        //  which uses (D(z) - D(init))/BOXLEN
```
steven-murray, line 62: Marking this as something to do.

src/py21cmfast/src/Stochasticity.c
```diff
@@ -213,22 +213,90 @@ void stoc_set_consts_cond(struct HaloSamplingConstants *const_struct, double con
...
+    // generate a random point in a sphere of given radius and centre
+    double x1, y1, z1, d1, r1;
+    x1 = gsl_rng_uniform(rng);
+    y1 = gsl_rng_uniform(rng);
+    z1 = gsl_rng_uniform(rng);
+    d1 = sqrt(x1 * x1 + y1 * y1 + z1 * z1);
+    r1 = gsl_rng_uniform(rng);
+    point[0] = centre[0] + (radius * r1 * x1 / d1);
+    point[1] = centre[1] + (radius * r1 * y1 / d1);
+    point[2] = centre[2] + (radius * r1 * z1 / d1);
```
steven-murray, line 275: How can this give negative values of a coordinate? `radius` is positive, `r1` is on (0,1), `x1` is on (0,1) and d1 is positive.

steven-murray (commented):
Thanks @daviesje this is looking good so far! I think it's possible that integration tests would fail because of the different RNG state that you'll have in this PR.

src/py21cmfast/src/PerturbHaloField.c
daviesje: I would classify this under #495. Since ADACS did a bit of this reorganisation in their work I'm planning to piggyback off their efforts in the next big PR to make everything uniform there.

steven-murray: Cool, no worries

src/py21cmfast/drivers/coeval.py
```diff
@@ -687,65 +678,51 @@ def _redshift_loop_generator(
...
                 ionized_box=this_ionized_box,
-                brightness_temperature=_bt,
+                brightness_temperature=this_bt,
                 ts_box=this_spin_temp,
                 halobox=this_halobox,
                 photon_nonconservation_data=photon_nonconservation_data,
             )
 
+            # yield before the cleanup, so we can get at the fields before they are purged
+            yield iz, this_coeval
```
steven-murray, line 700: Great idea!

src/py21cmfast/drivers/coeval.py
```diff
@@ -687,65 +678,51 @@ def _redshift_loop_generator(
...
+
+            # We purge previous fields and those we no longer need
+            if prev_coeval is not None:
+                prev_coeval.perturbed_field.purge(force=True)
+                if inputs.matter_options.USE_HALO_FIELD:
+                    prev_coeval.halobox.prepare_for_next_snapshot(force=True)
```
steven-murray, line 706: Weren't we anyway purging the prev_coeval simply because it goes out of scope?

src/py21cmfast/drivers/coeval.py
```diff
@@ -756,11 +733,7 @@ def _setup_ics_and_pfs_for_scrolling(
 
     # We can go ahead and purge some of the stuff in the initial_conditions, but only if
     # it is cached -- otherwise we could be losing information.
-    with contextlib.suppress(OSError):
-        initial_conditions.prepare_for_perturb(
-            astro_options=inputs.astro_options, force=always_purge
-        )
-
+    initial_conditions.prepare_for_perturb(force=True)
```
steven-murray, line 736: Can you explain why we are now always forcing the purge?

src/py21cmfast/drivers/single_field.py
```diff
@@ -371,20 +376,24 @@ def interp_halo_boxes(
...
+            logger.debug(f"field {field}")
+            logger.debug(
+                f"desc z={hbox_desc.redshift}, min/max/mean = ({field_desc.min()},{field_desc.max()},{field_desc.mean()})"
+            )
+            logger.debug(
+                f"prog z={hbox_prog.redshift}, min/max/mean = ({field_prog.min()},{field_prog.max()},{field_prog.mean()})"
+            )
+            logger.debug(
+                f"intp z={hbox_out.redshift}, min/max/mean = ({interp_field.min()},{interp_field.max()},{interp_field.mean()})"
+            )
```
steven-murray, line 396: Is this still required?

src/py21cmfast/src/SpinTemperatureBox.c
```diff
@@ -1746,8 +1747,8 @@ void ts_main(float redshift, float prev_redshift, float perturbed_field_redshift
             Ts_ave += ts_cell.Ts;
             Tk_ave += ts_cell.Tk;
             J_LW_ave += ts_cell.J_21_LW;
-            eps_lya_inj_ave += rad.dstarlya_cont_dt;
-            eps_lya_cont_ave += rad.dstarlya_inj_dt;
+            eps_lya_inj_ave += rad.dstarlya_inj_dt;
+            eps_lya_cont_ave += rad.dstarlya_cont_dt;
```
steven-murray, line 1751: Wow nice catch lol

tests/produce_integration_test_data.py
```diff
@@ -524,42 +532,27 @@ def print_failure_stats(test, truth, inputs, abs_tol, rel_tol, name):
...
-        stacklevel=2,
-    )
-    warnings.warn(
-        f"max abs diff of failures {np.fabs(truth - test)[sel_failed].max():.4e} relative {(np.fabs(truth - test) / truth)[sel_failed].max():.4e}",
-        stacklevel=2,
-    )
+    message = f"{name}: atol {abs_tol} rtol {rel_tol} failed {sel_failed.sum()} of {sel_failed.size} {sel_failed.sum() / sel_failed.size * 100:.4f} %\n"
+    message += f"subcube of failures [min] [max] {[f.min() for f in failed_idx]} {[f.max() for f in failed_idx]}\n"
+    message += f"failure range truth ({truth[sel_failed].min():.3e},{truth[sel_failed].max():.3e}) test ({test[sel_failed].min():.3e},{test[sel_failed].max():.3e})\n"
+    message += f"max abs diff of failures {np.fabs(truth - test)[sel_failed].max():.4e} relative {(np.fabs(truth - test) / truth)[sel_failed].max():.4e}\n"
```
steven-murray, line 538: This actually might be easier to read if formatted as a multiline string, i.e.

```python
message = f"""
{name}: atol {abs_tol} ... etc
subcube of failures [min] [max] .... etc
"""
```

src/py21cmfast/drivers/coeval.py
daviesje: The references are still stored in the `perturbed_field` list, which is loaded at the start of the loop

src/py21cmfast/drivers/coeval.py
daviesje: Good catch, I will change this to depend on the `CacheConfig`

src/py21cmfast/src/SpinTemperatureBox.c
daviesje: Thankfully it only affected the debug printing

tests/produce_integration_test_data.py
daviesje: I cleaned it up a bit using implicit concatenation, but the indentation on the multiline strings makes this a bit annoying.

steven-murray (approved):
Nice! As a side-note, I think it may be helpful to have another object similar to the `CacheConfig` where users can specify which boxes they care about *keeping in memory* (or maybe this can be rolled into the `CacheConfig` itself. Then we can purge individual boxes _even if they're not written to disk_ if the user explicitly says they don't want them at the end.

PR #532, opened 2025-07-11
feat: significantly enhance the CLI interface

steven-murray:
This PR updates and enhances the UX of the CLI.

Things it does:

1. Moves from `click` to `cyclopts` as the CLI framework. This makes the code much easier to read -- it looks much more like regular python.
2. Removes support for most single-field calculations, as I do not see them being very helpful. The exception is `ics` which can be useful to run standalone.
3. Fixes support for `coeval` and `lightcone` commands so they don't break out-of-the-box:
  - Updates the `--config` argument to be either `--template` (a name of a builtin template) or `--param-file` (a custom TOML file made by the user). One of these is required.
  - Better validation of input parameters like `--redshift` for `coeval` and `--redshift-range` for `lightcone`.  
4. Changes the interface to be more heirarchical and intuitive (IMHO). So now you have the following available commands:

```bash
21cmfast template avail       # list available templates
21cmfaste template create  # create a new template
21cmfast run ics
21cmfast run coeval
21cmfast run lightcone
21cmfast run params --help  # list all available simulation parameters, in a nice grouped way
```

5. The `run` subcommands share many arguments, and this is much more easily achieved in a DRY way by `cyclopts`
6. Uses `rich` for rich markup of output, for example:
  - We now have nice colored warnings and success messages
  - We have a banner when starting a `run` command (check it out, it features ducks...)
7. Uses @jordanflitter's nice progressbar for `coeval` and `lightcone`
8.  As mentioned above, all the simulation parameters (i.e. the parameters in `InputParameters`) are now well-documented on the CLI. While it was _possible_ to pass these before (with some hackery of the `click.Context`), they weren't documented (the CLI just accepted arbitrary keyword arguments like `--HII_DIM 50` and tried to coerce them to the correct types). Now, every possible parameter is a first-class CLI parameter that will get coerced to the correct type *according to the InputStruct classes*. Also, the docstrings describing the parameters are automatically read and presented to the user. These do _not_ appear in the `--help` for commands like `run ics` because it's a huge amount of parameters and is very noisy, but the user can run `21cmfast run params --help` to see them all. 

I recommend having a little play with it to try and find edge-cases that aren't smoothed out!


jordanflitter (approved):
By the power that was given to me by Steven, I hereby approve these changes.

src/py21cmfast/cli.py
```diff
@@ -1,827 +1,733 @@
...
-    param_structs = create_params_from_template(config, **ctx)
-    inputs = InputParameters(random_seed=seed, **{**param_structs, **kwargs})
+FORMAT = "%(message)s"
+logging.basicConfig(
+    level="WARNING",
+    format=FORMAT,
+    datefmt="[%X]",
+    handlers=[RichHandler(rich_tracebacks=True, console=cns)],
+)
+logger = logging.getLogger("py21cmfast")
```
daviesje, line 51: I'm guessing we don't want to use `__name__` here since this is a special logger for the CLI?

src/py21cmfast/cli.py
```diff
@@ -1,827 +1,733 @@
...
+    for field in fields(Parameters):
+        this = getattr(params, field.name)
+        kwargs |= {
+            name: val for name, val in attrs.asdict(this).items() if val is not None
+        }
+
+    new = inputs.evolve_input_structs(**kwargs)
+    modified = new != inputs
+    return new, modified
+
```
daviesje, line 211: Just for my understanding, the `_param_cls_factory` returns a new class based on the parameter class that holds all the fields but where they are all optional and default `None`. These are grouped into the `Parameters` class and combined with the other options in `RunOptions`, which contains a template. `Parameters` overwrites all parameters from the template given in `RunOptions`

src/py21cmfast/drivers/coeval.py
```diff
@@ -605,123 +623,121 @@ def _redshift_loop_generator(
...
+                # Only evolve on the node_redshifts, not any redshifts in-between
+                # that the user might care about.
+                prev_coeval = this_coeval
 
-        if z in inputs.node_redshifts:
-            # Only evolve on the node_redshifts, not any redshifts in-between
-            # that the user might care about.
-            prev_coeval = this_coeval
-        yield iz, this_coeval
+            yield iz, this_coeval
```
daviesje, line 740: Is this all just the progressbar change? git has done the diff very strangely

src/py21cmfast/lightconers.py
```diff
@@ -161,7 +160,9 @@ def with_equal_cdist_slices(
         )
 
     def make_lightcone_slices(
-        self, c1: Coeval, c2: Coeval
+        self,
+        c1: Coeval,  # noqa: F821
+        c2: Coeval,  # noqa: F821
```
daviesje, line 165: is it better to have the unused import or the undefined typehints ehre?

src/py21cmfast/src/PerturbField.c
```diff
@@ -204,14 +204,6 @@ int ComputePerturbField(float redshift, InitialConditions *boxes, PerturbedField
...
 
         LOG_DEBUG("Computing Perturbed Field at z=%.3f", redshift);
-        // perform a very rudimentary check to see if we are underresolved and not using the linear
-        // approx
-        if ((simulation_options_global->BOX_LEN > simulation_options_global->DIM) &&
-            matter_options_global->PERTURB_ALGORITHM > 0) {
-            LOG_WARNING(
-                "Resolution is likely too low for accurate evolved density fields\n \
-                It is recommended that you either increase the resolution (DIM/BOX_LEN) or set PERTURB_ALGORITHM to 'LINEAR'\n");
-        }
```
daviesje, line 214: Have we checked what actually happens in this case?

tests/test_cli.py
```diff
@@ -1,198 +1,297 @@
...
+        """Test that providing both --template and --param-file errors."""
+        new = tmp_path / "new.toml"
+        app(f"template create --template simple --out {new}")
+
+        # This should fail
+        with pytest.raises(SystemExit):
+            app(f"template create --param-file {new} --template simple --out here.toml")
+
+    def test_default_minimal(self, tmp_path: Path):
+        """Test that providing neither --template not --param-file errors."""
```
daviesje, line 62: Minor thing, but this docstring seems misleading

tests/test_cli.py
```diff
@@ -1,198 +1,297 @@
...
+        """Test that modifying params without an explicit file creates a random file."""
+        runp = RunParams(
+            param_selection=ParameterSelection(template="simple"), cachedir=self.tmpdir
+        )
+        params = Parameters(simulation_options=cli._SimulationOptions(HII_DIM=37))
+
+        _run_setup(runp, params)
+        out = capsys.readouterr().out
+        assert "Wrote full configuration" in out
+        assert "simple.toml" not in out  # got a random uuid
```
daviesje, line 144: Just so I understand the rationale here: since we already tested the template writing setup earlier it's fine to take the output log lines as evidence that the actual writing was done correctly? I'm fine with this I just wanted to make sure it was intentional

daviesje (approved):
Looks good to me, I had a few minor questions though

src/py21cmfast/cli.py
steven-murray: I think if we use `__name__`, it will affect the logger called `"py21cmfast.cli"` and all its children, but not the loggers from e.g. `py21cmfast.wrapper` etc. So yeah, I wanted to name it with the top-level 21cmFAST package.

src/py21cmfast/cli.py
steven-murray: Yep! Basically, I wanted to make it possible to set any parameter in any of the `InputStruct` classes (`AstroParams` etc), but I didn't want to re-write all the parameters that exist in all of the `InputStruct` subclasses (`AstroParams` etc). At the same time, I wanted the behaviour to be such that when the user explicitly sets such a parameter, it overwrites whatever is given in `--template` or `--param-file`.

So, I couldn't just use the `InputStruct` classes directly, because they all have defaults, and so I couldn't tell if they were "set" or not, so they would always overwrite the template. Thus, I create new dummy classes that just have all the parameters as well as the docstrings, but with default of None.

tests/test_cli.py
```diff
@@ -1,198 +1,297 @@
...
+
+
+_small_box = (
+    "--hii-dim 25 --dim 50 --box-len 50 --zprime-step-factor 1.2 --z-heat-max 20"
 )
```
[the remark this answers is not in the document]
steven-murray: BTW we could consider cascading templates :-D

src/py21cmfast/lightconers.py
steven-murray: I would use a "string" to forward reference the type, but the ruff "upgrade" format rule keeps automatically taking away the string, which then results in F821 firing. This is the best I could manage. See https://github.com/astral-sh/ruff/issues/8700

PR #535, opened 2025-07-20
Add HIRES_TO_LOWRES_FACTOR, LOWRES_CELL_SIZE_MPC and cascading templates

steven-murray:
This adds two major QoL features:

1. Adds the `HIRES_TO_LOWRES_FACTOR` and `LOWRES_CELL_SIZE_MPC` arguments to `SimulationOptions`. These have defaults of 3 an and 1.5 respectively, and control the _default_ value of `DIM` and and `BOX_LEN` based on `HII_DIM`. This means you can easily update `HII_DIM` alone and get reasonable boxes, not worrying about rescaling the DIM and BOX_LEN. The DIM and BOX_LEN can still be set explicitly, and that will override whatever `HIRES_TO_LOWRES_FACTOR` and `LOWRES_CLEL_SIZE_MPC` are set at.
2. Enables cascading templates. In particular, I have added several new "size-" built-in templates that will simply overwrite parameters that affect the "size" of the simulation (including the number of node redshifts), which can now be appended to a given template to change its size. So you can do for example `21cmfast run lightcone --template Park19 --template size-mini`.

It also fixes a bug in `InputParameters.from_template` where it is said in the docstring that the `**kwargs` go to over-riding the parameters, but instead they can only affect the init parameters of the `InputParameters` class (effectively just `node_redshifts` and `random_seed`).


src/py21cmfast/templates/manifest.toml
```diff
@@ -60,14 +61,33 @@ file = "Munoz21.toml"
...
-
 [[templates]]
 name = "fixed-halos"
 file = "fixed-halos.toml"
 aliases = ["fixed-halos"]
 description = "integrals of the CHMF on the Eulerian (evolved) density grid, FFRT-P in Davies+22 or ESF-E in Trac+22"
+
+
+[[templates]]
+name = "size-mini"
```
jordanflitter, line 72: Maybe call it "size-tiny"? Just to reduce confusion with the mini-halos templates

steven-murray: Ah yes of course!

src/py21cmfast/wrapper/inputs.py
```diff
@@ -664,9 +676,45 @@ class SimulationOptions(InputStruct):
     # NOTE (Jdavies): I do not currently have great priors for this value
     CORR_LX: float = field(default=0.2, converter=float)
 
-    @DIM.default
-    def _dim_default(self):
-        return 3 * self.HII_DIM
+    @property
```
jordanflitter, line 679: What happens if the user provides `BOX_LEN`, `HII_DIM` and `LOWRES_CELL_SIZE_MPC` and they are not consistent with each other? (Same question for `DIM`, `HII_DIM` and `HIRES_TO_LOWRES_FACTOR`)
Is there an error we throw in that case?

src/py21cmfast/wrapper/inputs.py
```diff
@@ -1492,6 +1568,13 @@ def asdict(
                 if isinstance(getattr(self, k), InputStruct)
             }
 
+        if only_cstruct_params:
+            for k, v in dct.items():
+                kls = getattr(self, k)
+                if isinstance(kls, InputStruct):
+                    dct[k] = {kk: vv for kk, vv in v.items() if kk in kls.cdict} | {
+                        kk: getattr(kls, kk) for kk in kls.cdict if kk not in dct[k]
+                    }
```
daviesje, line 1577: Two quick things here:

1: maybe I'm wrong about the convention, but `kls` makes me think it's the class itself, when it's an instance.
2: Is the RHS of the dictionary merge for the new non-field properties? Could a user mess these up e.g: `inputs.HII_DIM=20`, `inputs.DIM=40`, `inputs.HIRES_TO_LOWRES_FACTOR=10` and then propogate them with a clone?

daviesje (approved):
This looks amazing! just had one question

src/py21cmfast/templates/manifest.toml
steven-murray: Done

src/py21cmfast/wrapper/inputs.py
steven-murray: No, no error is thrown, but DIM and BOX_LEN take precedence. If they are passed, then they are used (and the `LOWRES_CELL_SIZE_MPC` and `HIRES_TO_LOWRES_FACTOR` are changed under the hood). Unfortunately with these situations where two different parameters affect one required attribute, there is really no single best way to handle it. In the way I've handled it, for example, if you set HII_DIM and DIM (with DIM/HII_DIM=2 say), then later change HII_DIM to be twice as large, it will not think "hey your original factor was 2, so let's scale up DIM as well", it will just keep DIM as 2x the original HII_DIM, because it's set. But I think at least in the way I have it here, the legacy way of setting the params will work without surprises (even after saving/loading the file, or doing an `.evolve()`). And if you just never never set DIM or BOX_LEN, then it should also work without surprises.

jordanflitter: The reason why `DIM` and `BOX_LEN` take precedence is simply because their methods is located higher in inputs.py?
Do we have tests that check this logic and ensure that things behave exactly as planned? (including templates + `.evolve()`)

src/py21cmfast/wrapper/inputs.py
daviesje: What I meant by 3 is similar to Jordan's comment that i don't see a strict requirement for the dimensions and ratio parameters to be consistent. If someone uses a template with `BOX_LEN` and `DIM` and then evolves it with `HIRES_TO_LOWRES_FACTOR` or `LOWRES_CELL_SIZE_MPC`, their changes would be ignored without warning or error. I'm not sure the best way to address this, I do like the precenence of the old parameters, but maybe we should throw an error when evolving one of the dimension ratios *after* the dimension parameters are set?

src/py21cmfast/wrapper/inputs.py
steven-murray: No, the reason they take precedence is that they have a default of None, and if the user explicitly passes them, the class detects this, and uses that value, otherwise the generated value from `HII_DIM * HIRES_TO_LOWRES_FACTOR` is used.

In terms of tests, I am adding some now.

PR #538, opened 2025-07-20
Beta esc

### LINKED ISSUE #537: [Feature Req.] Redshift dependency in parameters of MASS_DEPENDENT_ZETA

As part of discussion with @andreimesinger, I am trying to combine the likelihood of [Qin et al.](https://arxiv.org/abs/2412.00799) @qyx268 with a CMB+BAO likelihood without the Planck tau prior.

I was told that this is implemented in [this branch](https://github.com/21cmfast/21cmFAST/tree/update_Rmax3).
After checking the code, I realised it probably makes little sense for me to continue to modify an old version since the source code has changed significantly. Instead it would be nicer to add this to the current v4.

Redshift dependence in parameters such as BETA_ESC10, BETA_ESC7, BETA_STAR10 that adds redshift dependence to the escape fraction calculation of halos (Eq.2 in 2412.00799)

Maybe this can be done by manually modifying the input parameters at each coeval level for each node redshift, but I think the natural solution should be to add input parameters.

**Additional context**

I have some attempt at doing this, with a very naive verification.
Now it is just for BETA_ESC.
The big problem here I think is that I don't really understand the photon conservation and how it's done at the code level. Some help is appreciated in helping me check this.

zhaotingchen:
Attempt to fix #537

### DIFF
--- a/src/py21cmfast/src/hmf.c
+++ b/src/py21cmfast/src/hmf.c
@@ -69,6 +69,7 @@ struct parameters_gsl_MF_integrals {
     // Nion additions
     double f_esc_norm;
     double alpha_esc;
+    double beta_esc;
     double Mlim_esc;
 
     // Minihalo additions
@@ -358,8 +359,9 @@ double nion_fraction(double lnM, void *param_struct) {
     struct parameters_gsl_MF_integrals p = *(struct parameters_gsl_MF_integrals *)param_struct;
     double Fstar = log_scaling_PL_limit(lnM, p.f_star_norm, p.alpha_star, 10 * LN10, p.Mlim_star);
     double Fesc = log_scaling_PL_limit(lnM, p.f_esc_norm, p.alpha_esc, 10 * LN10, p.Mlim_esc);
+    double Zesc = log(pow((1 + p.redshift) / 8.0, p.beta_esc));
 
-    return exp(Fstar + Fesc - p.Mturn_acg / exp(lnM) + lnM);
+    return exp(Fstar + Fesc + Zesc - p.Mturn_acg / exp(lnM) + lnM);
 }
 
 double nion_fraction_mini(double lnM, void *param_struct) {
@@ -856,7 +858,9 @@ double Nion_General(double z, double lnM_Min, double lnM_Max, double MassTurnove
         .HMF = matter_options_global->HMF,
         .gamma_type = 3,
     };
-    return IntegratedNdM(lnM_Min, lnM_Max, params, &u_nion_integrand, 0);
+    // z factor can be taken out of the integral
+    double Zesc = pow((1 + z) / 8.0, sc->beta_esc);
+    return Zesc * IntegratedNdM(lnM_Min, lnM_Max, params, &u_nion_integrand, 0);
 }
 
 double Nion_General_MINI(double z, double lnM_Min, double lnM_Max, double MassTurnover,
@@ -1008,7 +1012,7 @@ double Nion_ConditionalM(double growthf, double lnM1, double lnM2, double lnM_co
         .delta = delta2,
         .gamma_type = -3,
     };
-
+    double Zesc = pow((1 + sc->redshift) / 8.0, sc->beta_esc);
     if (lnM1 >= lnM_cond) return 0.;
     // return 1 halo at the condition mass if delta is exceeded and the condition is within the
     // integral limits
@@ -1024,7 +1028,8 @@ double Nion_ConditionalM(double growthf, double lnM1, double lnM2, double lnM_co
     // If we don't have a corresponding CMF, use EPS and normalise
     // NOTE: it's possible we may want to use another default
     if (params.HMF != 0 && params.HMF != 1 && params.HMF != 4) params.HMF = 0;
-    return IntegratedNdM(lnM1, lnM2, params, &c_nion_integrand, method);
+    // z factor can be taken out of the integral
+    return Zesc * IntegratedNdM(lnM1, lnM2, params, &c_nion_integrand, method);
 }
 
 double Xray_ConditionalM(double redshift, double growthf, double lnM1, double lnM2, double lnM_cond,
--- a/src/py21cmfast/wrapper/inputs.py
+++ b/src/py21cmfast/wrapper/inputs.py
@@ -971,6 +971,9 @@ class AstroParams(InputStruct):
     ALPHA_ESC : float, optional
         Power-law index of escape fraction as a function of halo mass. See Sec 2.1 of
         Park+2018.
+    BETA_ESC : float, optional
+        Power-law index of escape fraction as a function of redshift. See Eq. 2 of
+        Qin+2025.
     M_TURN : float, optional
         Turnover mass (in log10 solar mass units) for quenching of star formation in
         halos, due to SNe or photo-heating feedback, or inefficient gas accretion. Only
@@ -1066,6 +1069,10 @@ class AstroParams(InputStruct):
     )
     ALPHA_ESC = field(
         default=-0.5,
+        converter=float,
+    )
+    BETA_ESC = field(
+        default=0.0,
         converter=float,
     )
     F_ESC7_MINI = field(
--- a/src/py21cmfast/wrapper/inputs.py
+++ b/src/py21cmfast/wrapper/inputs.py
@@ -731,15 +731,15 @@ class AstroOptions(InputStruct):
         Whether the minimum halo mass (for ionization) is defined by
         mass or virial temperature. Automatically True if `USE_MASS_DEPENDENT_ZETA`
         is True.
-    PHOTON_CONS_TYPE : int, optional
+    PHOTON_CONS_TYPE : str, optional
         Whether to perform a small correction to account for the inherent
         photon non-conservation. This can be one of three types of correction:
 
-        0: No photon cosnervation correction,
-        1: Photon conservation correction by adjusting the redshift of the N_ion source field (Park+22)
-        2: Adjustment to the escape fraction power-law slope, based on fiducial results in Park+22, This runs a
+        'no-photoncons': No photon cosnervation correction,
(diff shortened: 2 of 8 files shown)


zhaotingchen:
A naive test showing that it is affecting the ioniosation efficiency:

```python
import numpy as np
import matplotlib.pyplot as plt
import py21cmfast as p21c
from multiprocessing import cpu_count

max_num_thread = cpu_count()
sim_op = {
    "HII_DIM":32, 
    "BOX_LEN":250,
    "N_THREADS":max_num_thread,
}
matter_op = {
    "PERTURB_ON_HIGH_RES":False, # True for production
    "MINIMIZE_MEMORY":True, # False for production
    "USE_HALO_FIELD":True, # True for production
    "HALO_STOCHASTICITY":False, # True for production
}
astro_op = {
    "USE_TS_FLUCT":False, # True for production
    "SUBCELL_RSD":False, # True for production
    "INHOMO_RECO":False, # True for production
    "USE_MINI_HALOS":False,
    "USE_MASS_DEPENDENT_ZETA":True,
    "PHOTON_CONS_TYPE":"no-photoncons", #v4 # change to something else for production
    #"PHOTON_CONS":True,
}
astro_pars = p21c.AstroParams()
z_min = 5.0
z_max = 10.0
cosmo_pars = p21c.CosmoParams()
cache = p21c.OutputCache(cache_dir)

inputs = p21c.InputParameters(
    random_seed=12345,
    cosmo_params=cosmo_pars,
    matter_options=matter_op,
    astro_options=astro_op,
    astro_params=astro_pars,
    simulation_options=sim_op,
)

coevals = []
for coeval,in_outputs in p21c.generate_coeval(
    inputs=inputs,
    out_redshifts=[8, 9],
    cache=cache,
):
    if in_outputs:
        coevals.append(coeval)

plt.scatter(np.log10(coevals[0].halo_mass).ravel(),np.log10(coevals[0].n_ion.ravel()),)
```
The plot will show a straight line due to the scaling between f_esc and halo_mass.
image

In default, `BETA_ESC` is 0 and this scaling does not change with redshift:
```python
print(coevals[0].redshift,np.polyfit(np.log10(coevals[0].halo_mass).ravel(),np.log10(coevals[0].n_ion.ravel()),1))
print(coevals[1].redshift,np.polyfit(np.log10(coevals[1].halo_mass).ravel(),np.log10(coevals[1].n_ion.ravel()),1))
```
gives
```
9.0 [ 1.66731392 -6.39808893]
8.0 [ 1.66509232 -6.36885117]
```

changing the input to `astro_pars = p21c.AstroParams(BETA_ESC = -1.5)` and repeat gives
```
9.0 [ 1.66731389 -6.54345369]
8.0 [ 1.66509231 -6.44557988]
```
now there is a noticeable change in the intercept. Larger z have smaller n_ion due to beta<0.

`BETA_ESC = 1.5` gives
```
9.0 [ 1.66731392 -6.25272388]
8.0 [ 1.66509231 -6.29212233]
```
Where the relation flips as expected (but the change is too small?)

src/py21cmfast/src/hmf.c
```diff
@@ -358,8 +359,9 @@ double nion_fraction(double lnM, void *param_struct) {
     struct parameters_gsl_MF_integrals p = *(struct parameters_gsl_MF_integrals *)param_struct;
     double Fstar = log_scaling_PL_limit(lnM, p.f_star_norm, p.alpha_star, 10 * LN10, p.Mlim_star);
     double Fesc = log_scaling_PL_limit(lnM, p.f_esc_norm, p.alpha_esc, 10 * LN10, p.Mlim_esc);
+    double Zesc = log(pow((1 + p.redshift) / 8.0, p.beta_esc));
 
-    return exp(Fstar + Fesc - p.Mturn_acg / exp(lnM) + lnM);
+    return exp(Fstar + Fesc + Zesc - p.Mturn_acg / exp(lnM) + lnM);
```
daviesje, line 364: These are very hot functions. So saving log/pow calls here is very important. You can see how the above function `log_scaling_PL_limit` does this for the other power-laws and you could use two less here. However more importantly, this term is only calculated once per redshift, so it can be saved as a constant in the scaling parameters and simply loaded in.

src/py21cmfast/src/hmf.c
```diff
@@ -856,7 +858,9 @@ double Nion_General(double z, double lnM_Min, double lnM_Max, double MassTurnove
         .HMF = matter_options_global->HMF,
         .gamma_type = 3,
     };
-    return IntegratedNdM(lnM_Min, lnM_Max, params, &u_nion_integrand, 0);
+    // z factor can be taken out of the integral
+    double Zesc = pow((1 + z) / 8.0, sc->beta_esc);
+    return Zesc * IntegratedNdM(lnM_Min, lnM_Max, params, &u_nion_integrand, 0);
```
daviesje, line 863: See above comment, redshift-based escape fractions should be pre-computed once per snapshot

src/py21cmfast/src/hmf.c
```diff
@@ -1024,7 +1028,8 @@ double Nion_ConditionalM(double growthf, double lnM1, double lnM2, double lnM_co
     // If we don't have a corresponding CMF, use EPS and normalise
     // NOTE: it's possible we may want to use another default
     if (params.HMF != 0 && params.HMF != 1 && params.HMF != 4) params.HMF = 0;
-    return IntegratedNdM(lnM1, lnM2, params, &c_nion_integrand, method);
+    // z factor can be taken out of the integral
+    return Zesc * IntegratedNdM(lnM1, lnM2, params, &c_nion_integrand, method);
```
daviesje, line 1032: Here as well.

src/py21cmfast/wrapper/inputs.py
```diff
@@ -890,11 +890,11 @@ def _USE_MINI_HALOS_vald(self, att, val):
 
     @PHOTON_CONS_TYPE.validator
     def _PHOTON_CONS_TYPE_vld(self, att, val):
-        """Raise an error if using PHOTON_CONS_TYPE='z_photoncons' and USE_MINI_HALOS is True."""
+        """Raise an error if using PHOTON_CONS_TYPE='z-photoncons' and USE_MINI_HALOS is True."""
         if self.USE_MINI_HALOS and val == "z-photoncons":
             raise ValueError(
                 "USE_MINI_HALOS is not compatible with the redshift-based"
-                " photon conservation corrections (PHOTON_CONS_TYPE=='z_photoncons')! "
+                " photon conservation corrections (PHOTON_CONS_TYPE=='z-photoncons')! "
```
daviesje, line 897: Thanks for fixing this!

src/py21cmfast/wrapper/inputs.py
```diff
@@ -1333,6 +1340,18 @@ def _astro_params_validator(self, att, val):
...
+                and val.BETA_ESC != 0
+                and self.astro_options.PHOTON_CONS_TYPE
+                not in ["no-photoncons", "alpha-photoncons"]
+            ):
+                warnings.warn(
+                    f"You have set BETA_ESC != 0 but PHOTON_CONS_TYPE is {self.astro_options.PHOTON_CONS_TYPE}. "
+                    "This changes the escape fraction so it is not consistent with the manual setting of scaling."
+                    "Set PHOTON_CONS_TYPE to 'no-photoncons' or 'alpha-photoncons' if you want the scaling to be exact.",
+                    stacklevel=2,
+                )
```
daviesje, line 1354: Someone else may have an opinion on this, but if there is a real problem using these two models together then perhaps we should throw an error instead of a warning

daviesje (commented):
I'm assuming from your issue that you have already spoken with @andreimesinger or @steven-murray about the pros/cons of adding new parameters to the escape fraction so I'm just going to review the code here.

As for the "photon-conservation" model, I don't immediately see a problem with running our calibration simulation with a BETA_ESC in any of the cases, but maybe you found something wrong with the `f-photoncons` model here? In any case I wouldn't want these modules getting in the way of new features so in the worst case we can simply throw parameter errors when a user enables both

src/py21cmfast/wrapper/inputs.py
steven-murray: I think an error will be better. Warnings are too easy to ignore.

zhaotingchen: I don't understand is what is happening when we turn on `use_photoncons`, but I see in the code `f_esc10` gets modified per scaling relation if `PHOTON_CONS_TYPE == 3`. I will come back to it later when I understand the code better.

src/py21cmfast/src/hmf.c
daviesje: So I would implement this in the following way:

- Add a new `double` field to the `ScalingConstants` struct (in `scaling_relations.h`) which contains the entire power-law term from `BETA_ESC`
- Add a line to `set_scaling_constants (in `scaling_relations.c`) which initialises this term based on the redshift, this calculates the term once per redshift and stores it for fast multiplication later (no more `pow` calls)
- Multiply by this factor in two places: in `HaloBox.c` where the escape fraction is set in `set_halo_properties`, and in `IonisationBox.c` where the constant `ion_eff_factor_gl` is set in `set_ionbox_constants`.

Regarding the last step, the reason you have to alter two parts is just due to the way the discrete halo / density field models are split. It's not ideal but that's the way it's organised at the moment. Happy to answer any further questions

daviesje (commented):
This is looking a lot better, but I still have two questions:
- I know the Qin+25 model did not include minihalos at all, but would you like to include them? Either with the same scaling or another parameter? At the moment there is no scaling applied to them at all, and that might be confusing. If you wish not to include them, make it clear in the docstring for `BETA_ESC`
- Did you try any runs with `PHOTON_CONS=z-photoncons`?  I can't see anything that could go wrong here, since `BETA_ESC` should be applied to both the calibration and followup simulations. The warning/error check might not be necessary

PR #542, opened 2025-07-24
fix: small bug in reading old HDF5 files with potentially differing type

steven-murray:
This fixes a bug that @DanielaBreitman found when running the new code in main (after the halo coords PR: #530 by @daviesje)

She found that the dtype of the halo coords wasn't being accepted when trying to pass to C, and we realized that it was because she was reading in boxes made by the code before that PR. The problem was that when creating the `Array` objects from an HDF5 file, we use the dtype as saved in the HDF5 file, rather than the dtype required by the Array as specified in the code.

This just updates that logic so that we can cast arrays to their correct dtype as required by the present code.

### DIFF
--- a/src/py21cmfast/io/h5.py
+++ b/src/py21cmfast/io/h5.py
@@ -295,6 +295,11 @@ def read_inputs(
         group = group["InputParameters"]
 
     file_version = group.attrs.get("21cmFAST-version", None)
+    if file_version is None:
+        raise NotImplementedError(
+            f"The file {group.file.filename} is not a valid 21cmFAST v4 file."
+        )
+
     if file_version > __version__:
         warnings.warn(
             f"File created with a newer version {file_version} of 21cmFAST than this {__version__}. "
@@ -344,13 +349,18 @@ def _read_outputs(
 ):
     file_version = group.attrs.get("21cmFAST-version", None)
 
+    if file_version is None:
+        raise NotImplementedError(
+            f"The file {group.file.filename} is not a valid 21cmFAST v4 file."
+        )
+
     if file_version > __version__:
         warnings.warn(
             f"File created with a newer version of 21cmFAST than this. Reading may break. Consider updating 21cmFAST to at least {file_version}",
             stacklevel=2,
         )
-    else:
-        return _read_outputs_v4(group, struct, redshift, inputs)
+
+    return _read_outputs_v4(group, struct, redshift, inputs)
 
 
 def _read_outputs_v4(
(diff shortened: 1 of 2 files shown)


daviesje (approved):
Looks great! Thanks for catching the bug! I outlined some very silly edge cases below but they shouldn't be a problem

src/py21cmfast/io/h5.py
```diff
@@ -339,23 +350,51 @@ def _read_outputs(group: h5py.Group):
...
+            obj,
+            name,
+            attrs.evolve(
+                array,
+                state=ArrayState(on_disk=True),
+                cache_backend=H5Backend(
+                    path=group.file.filename, dataset=group.name + f"/{name}"
+                ),
+            ),
+        )
```
daviesje, line 398: (comment missed in review) I spent a while trying to figure out whether the given file could be accidentally overwritten by the `Array` module here on a compute. Since the caching explicity passes a new `CacheBackend` to `Array.written_to_disk` the only way this could happen is if a user calls `Array.written_to_disk`, after explicitly calling `.compute(allow_already_computed=True) with `write=False`, which I don't think will ever happen.

src/py21cmfast/io/h5.py
```diff
@@ -339,23 +350,51 @@ def _read_outputs(group: h5py.Group):
             stacklevel=2,
         )
     else:
-        return _read_outputs_v4(group)
+        return _read_outputs_v4(group, struct, redshift, inputs)
```
daviesje, line 353: At the moment, if there's a version mismatch, we don't even attempt the read, is this correct?

src/py21cmfast/io/h5.py
```diff
@@ -339,23 +350,51 @@ def _read_outputs(group: h5py.Group):
...
 
 
-def _read_outputs_v4(group: h5py.Group):
-    arrays = {
-        name: Array(
-            dtype=box.dtype,
```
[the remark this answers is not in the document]
steven-murray: This is a good idea, although I think it would take some time to get it right, so I think I'll do this on another PR (since it wouldn't be API breaking, we could do it after v4 is in). I'll make an issue in the meantime.

src/py21cmfast/io/h5.py
daviesje: I think in this case we should warn and attempt to read. Especially since this is for *newer* lightcones, not older. The warning into possible error is probably better than the warning into nothing

steven-murray: Yes, I just noticed what you meant properly as I was adding coverage for missed lines -- the `else:` shouldn't be here! I think that was just a bug. I've removed it now. I also explicitly error on "older" boxes without a 21cmFAST-version attribute at all (this can be lifted once #543 is in).

PR #548, opened 2025-07-29
fix: small bug in caching

DanielaBreitman:
`get_output_struct_at_z` currently uses the index `closest` instead of the actual redshift value `zs_of_kind[closest]`.
This leads to incorrect behaviour when the user passes a given `z` to this method.


steven-murray (approved):
Wow, good catch!

PR #550, opened 2025-07-30
maint: changed halo field buffer

DanielaBreitman:
Do not raise an error if we have less than 1e6 halos. Use the 1.2 buffer factor only above 1e6 halos.
This is because the simulation keeps crashing with e.g. `More than 111505 halos (expected 9.3e+04) with buffer size factor 1.2` where the number of halos is pretty close to the expected number with the buffer size.

### DIFF
--- a/src/py21cmfast/src/Stochasticity.c
+++ b/src/py21cmfast/src/Stochasticity.c
@@ -859,7 +859,7 @@ int sample_halo_grids(gsl_rng **rng_arr, double redshift, float *dens_field,
                         // we do not want to save these
                         if (hm_buf[i] < simulation_options_global->SAMPLER_MIN_MASS) continue;
 
-                        if (count >= arraysize_local) {
+                        if (count >= 1e6 && count >= arraysize_local) {
                             LOG_ERROR(
                                 "More than %llu halos (expected %.1e) with buffer size factor %.1f",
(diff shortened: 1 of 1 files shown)


daviesje (commented):
Doing this without changing the actual buffer will cause many segfaults.

Instead of changing the error , you should change the minimum buffer size here: https://github.com/21cmfast/21cmFAST/blob/4813fca7236559cd3ddd2439c808a7e1d219d633/src/py21cmfast/wrapper/outputs.py#L772

DanielaBreitman:
Good idea @daviesje ! Though looking quickly at the code, it seems like[ the default min](https://github.com/21cmfast/21cmFAST/blob/main/src/py21cmfast/wrapper/cfuncs.py#L127) size is already 1e6?

steven-murray:
@daviesje something is fishy here -- the minimum buffer size should be 1e6, as pointed out by @DanielaBreitman, but she's getting segfaults for nhalos ~ 1e5.

DanielaBreitman:
I added some print statements and turns out the buffer size is indeed not set and then gets correctly set to some value over 1e6 [here](https://github.com/21cmfast/21cmFAST/blob/main/src/py21cmfast/wrapper/outputs.py#L772)... so I am not sure why I am getting those errors...

Issue #551, opened 2025-08-01
[BUG] error when running coeval boxes with out_redshifts


jordanflitter:
The new purge lines that were recently added cause a problem when running a coeval with `out_redshifts`. The problem is [in this line](https://github.com/21cmfast/21cmFAST/blob/main/src/py21cmfast/drivers/coeval.py#L817). The condition should be `iz + 1 < len(inputs.node_redshifts)` (thanks @steven-murray for quickly suggesting a solution). If `out_redshifts` is not empty, `all_redshifts` contains more entries than `inputs.node_redshifts` and at the last iterations an error is raised as we passed amount of entries in `inputs.node_redshifts`.

I fixed that on a local branch on my machine and it seems to work fine now. I'll make a PR shortly just for that. @steven-murray Let me know if you want me to include new tests for `run_coeval` with `out_redshifts` (and tell me where to make these tests).

jordanflitter:
This issue was fixed in #554.

PR #554, opened 2025-08-03
Fixed bug and automatic warning when halos are purged

jordanflitter:
This short PR fixes the bug reported in #551 . In addition, I force purging the `XraySourceBox` 4D arrays in order to avoid the automatic warning that is raised when they are not saved on disk, as I can't believe someone will ever use them for post-processing (I believe that the vast majority of 21cmFAST users are even unaware to their existence).


src/py21cmfast/drivers/coeval.py
```diff
@@ -814,7 +814,7 @@ def _redshift_loop_generator(
                 if (
                     inputs.matter_options.USE_HALO_FIELD
                     and write.halobox
-                    and iz + 1 < len(all_redshifts)
+                    and iz + 1 < len(inputs.node_redshifts)
```
steven-murray, line 817: Eh, sorry but I think it's a little more complicated than this. I think we need another counter variable that tracks which node redshift we're up to. `iz` is the the number of redshifts we've completed including the out redshifts, but we have no tracker of the node redshift count. That's what should be used here.

src/py21cmfast/drivers/coeval.py
```diff
@@ -814,11 +814,11 @@ def _redshift_loop_generator(
                 if (
                     inputs.matter_options.USE_HALO_FIELD
                     and write.halobox
-                    and iz + 1 < len(all_redshifts)
+                    and iz_node + 1 < len(inputs.node_redshifts)
                 ):
                     for hbox in hbox_arr:
                         hbox.prepare_for_next_snapshot(
-                            next_z=inputs.node_redshifts[iz + 1]
+                            next_z=inputs.node_redshifts[iz_node + 1]
```
daviesje, line 821: I think we should just be using `all_redshifts` here. This may purge too early if the next step is not a node

daviesje: This also makes it simpler since we don't actually need to iz_node tracker

jordanflitter: @steven-murray what do you think?

steven-murray: Having a closer look, I think actually that the `hbox_arr` is incorrect. It appends halo boxes at `out_redshift` redshifts as well as `node_redshift` redshifts, which is going to end up meaning that the `xray_source_field` depends on the `out_redshift` array, which is not properly cached. 

All the computations should _only_ depend on the `node_redshifts`. @daviesje do you agree?

PR #557, opened 2025-08-04
Error outside pragma loop

jordanflitter:
The goal of this PR is to place error messages outside of pragma loops (if error is being raised during a parallel session, it could not print the details of the error and just show "segmentation fault", or "kernel crashed" in Jupyter notebooks). We do it now by initializing a shared flag to be `false` and "continuing" if the criteria for the error has met. If the flag is `true` at the end of the pragma loop we throw an error. @daviesje let me know if there are other places in the code where we could use the same technique (I currently made modifications only in  Stochasticity.c).


src/py21cmfast/src/Stochasticity.c
```diff
@@ -860,15 +860,8 @@ int sample_halo_grids(gsl_rng **rng_arr, double redshift, float *dens_field,
...
-                                "More than %llu halos (expected %.1e) with buffer size factor %.1f",
-                                arraysize_local,
-                                arraysize_local / config_settings.HALO_CATALOG_MEM_FACTOR,
-                                config_settings.HALO_CATALOG_MEM_FACTOR);
-                            LOG_ERROR(
-                                "If you expected to have an above average halo number try raising "
-                                "config_settings.HALO_CATALOG_MEM_FACTOR");
-                            Throw(ValueError);
+                            OUT_OF_BUFFER = true;
+                            continue;
```
daviesje, line 864: Since `count` is a private variable, the other threads will run through entirely. To avoid this you can set an `if(OUT_OF_BUFFER) continue;` at the start of the loop

src/py21cmfast/src/Stochasticity.c
```diff
@@ -789,7 +789,7 @@ int sample_halo_grids(gsl_rng **rng_arr, double redshift, float *dens_field,
     double total_volume_excluded = 0.;
     double total_volume_dexm = 0.;
     double cell_volume = VOLUME / pow((double)simulation_options_global->HII_DIM, 3);
-
+    bool OUT_OF_BUFFER = false;
```
daviesje, line 792: I would prefer lower-case for non defined/macro variables. I know this convention is not strictly kept throughout the code but let's try to move in that direction

src/py21cmfast/src/Stochasticity.c
```diff
@@ -907,6 +900,15 @@ int sample_halo_grids(gsl_rng **rng_arr, double redshift, float *dens_field,
...
     }
+    if (OUT_OF_BUFFER) {
+        LOG_ERROR("More than %llu halos (expected %.1e) with buffer size factor %.1f",
+                  arraysize_local, arraysize_local / config_settings.HALO_CATALOG_MEM_FACTOR,
+                  config_settings.HALO_CATALOG_MEM_FACTOR);
+        LOG_ERROR(
+            "If you expected to have an above average halo number try raising "
+            "config['HALO_CATALOG_MEM_FACTOR']");
+        Throw(ValueError);
+    }
```
daviesje, line 911: In order to avoid some confusion (and help #550), it might be a good idea to output counts in each thread in the error message, or at least refer to which thread had the overflow

daviesje (commented):
Unfortunately there are a lot of places where `Throw` is called from code which gets called in threaded sections and I can't list them all but if you search for `Throw` in the C directory you'll see what I mean. If you just want to fix this one for now that's fine, since there won't be a one-size-fits-all solution:

- If the Throw is in the same function as the `pragma`, you can handle it like you did here.
- If it's in a `void` function called from a parallel section, you can change it to return an integer error code and handle the error in the parent function
- If it's in a non-void function you could return `NAN` or try to refactor it to return error codes

daviesje (approved):

PR #561, opened 2025-08-21
ESF-L Implementation

daviesje:
This is the second step towards the Lagrangian source model.

This PR implements the ESF-L Source model from https://arxiv.org/pdf/2109.10375, both for the `AVG_BELOW_SAMPLER` or `FIXED_HALO_GRIDS` contributions when `USE_HALO_FIELD=True`. Here halo properties are integrated over the HMF using the low resolution initial conditions density field, and then those contributions are perturbed using 2LPT to their real locations.

A series of comparison plots can be found at https://drive.google.com/drive/folders/1Q-YBXlkA0kStx5cD76Bs6vX34-raSQ84?usp=drive_link.


tests/test_perturb.py
```diff
@@ -106,43 +106,24 @@ def get_fake_ics(self, inputs, test_pt_z):
...
+            else inputs.simulation_options.INITIAL_REDSHIFT
+        )
+        roll_var = {
+            "LINEAR": (0, 0, 0),
+            "ZELDOVICH": (0, 1, 0),
+            "2LPT": (0, 1, -1),
+        }[inputs.matter_options.PERTURB_ALGORITHM]
+        d_z = cf.get_growth_factor(inputs=inputs, redshift=z_d)
+
+        expected_dens = np.roll(ics.get("lowres_density"), roll_var, (0, 1, 2))
```
steven-murray, line 126: Can you add a brief comment above this... it's not obvious what the rolling is supposed to do

steven-murray (approved):
Thanks again @daviesje -- this all looks good to me. I have made one suggested comment, but feel free to merge once you've add the explanatory comment

PR #563, opened 2025-09-02
Onthefly haloperturb

daviesje:
Next step in the new source model: The 2LPT for the halo fields is now done on-the-fly within `Halobox.c`. We use a Cloud-in-Cell both to read the turnover mass grids and to sum halo properties onto the `HII_DIM` grid.

This has effects on memory and time that are not straightforward: We should be using approximately half the memory in the halo loop from the fields since we no longer have the `PerturbHaloField` in the main run. however we need to hang onto the IC velocity fields for the entire run.

If a user wants the real location of halos, they can call `perturb_halo_field` outside of a full simulation. Since there is no RNG there this should be fine.

TODOs for this PR:
- Make this work properly with the `PERTURB_ON_HIGH_RES` flag
- Perhaps make `PerturbHaloField` return real halo properties instead of the RNG fields, since it's a post-processing field now
- fix any failing tests
- examine the new power-spectra, since we expect a reduction in shot noise from the CIC

### DIFF
--- a/src/py21cmfast/drivers/single_field.py
+++ b/src/py21cmfast/drivers/single_field.py
@@ -206,7 +206,7 @@ def compute_halo_grid(
     redshift: float,
     initial_conditions: InitialConditions,
     inputs: InputParameters | None = None,
-    halo_list: HaloField | None = None,
+    halo_field: HaloField | None = None,
     previous_spin_temp: TsBox | None = None,
     previous_ionize_box: IonizedBox | None = None,
 ) -> HaloBox:
@@ -245,11 +245,11 @@ def compute_halo_grid(
     """
     box = HaloBox.new(redshift=redshift, inputs=inputs)
 
-    if halo_list is None:
+    if halo_field is None:
         if not inputs.matter_options.FIXED_HALO_GRIDS:
-            raise ValueError("You must provide halo_list if FIXED_HALO_GRIDS is False")
+            raise ValueError("You must provide halo_field if FIXED_HALO_GRIDS is False")
         else:
-            halo_list = HaloField.dummy()
+            halo_field = HaloField.dummy()
 
     # NOTE: due to the order, we use the previous spin temp here, like spin_temperature,
     #       but UNLIKE ionize_box, which uses the current box
@@ -279,7 +279,7 @@ def compute_halo_grid(
 
     return box.compute(
         initial_conditions=initial_conditions,
-        halo_field=halo_list,
+        halo_field=halo_field,
         previous_ionize_box=previous_ionize_box,
         previous_spin_temp=previous_spin_temp,
     )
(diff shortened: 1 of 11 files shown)


src/py21cmfast/io/caching.py
```diff
@@ -291,7 +295,7 @@ class RunCache:
     IonizedBox: dict[float, Path] = _dict_of_paths_field()
     BrightnessTemp: dict[float, Path] = _dict_of_paths_field()
     HaloBox: dict[float, Path] | None = _dict_of_paths_field()
-    PerturbHaloField: dict[float, Path] | None = _dict_of_paths_field()
+    HaloField: dict[float, Path] | None = _dict_of_paths_field()
```
steven-murray, line 298: I don't know if this is something we should change at this point, but I don't like having `HaloBox` and `HaloField`... just from the name it's not easy to tell what the difference between them is.

jordanflitter: How about `HaloCatalog` or `HaloList` instead of `HaloField`?

src/py21cmfast/drivers/single_field.py
```diff
@@ -158,8 +158,11 @@ def determine_halo_list(
 def perturb_halo_list(
     *,
     initial_conditions: InitialConditions,
-    halo_field: HaloField,
-) -> PerturbHaloField:
+    inputs: InputParameters | None = None,
+    previous_spin_temp: TsBox | None = None,
+    previous_ionize_box: IonizedBox | None = None,
+    halo_field: HaloCatalog,
```
steven-murray: this should be `halo_catalog`

src/py21cmfast/drivers/single_field.py
```diff
@@ -168,12 +171,12 @@ def perturb_halo_list(
     initial_conditions : :class:`~InitialConditions`
         The initial conditions of the run. The user and cosmo params
         as well as the random seed will be set from this object.
-    halo_field: :class: `~HaloField`
+    halo_field: :class: `~HaloCatalog`
```
steven-murray: and here

src/py21cmfast/drivers/single_field.py
```diff
@@ -245,13 +273,11 @@ def compute_halo_grid(
     """
     box = HaloBox.new(redshift=redshift, inputs=inputs)
 
-    if perturbed_halo_list is None:
+    if halo_field is None:
         if not inputs.matter_options.FIXED_HALO_GRIDS:
-            raise ValueError(
-                "You must provide the perturbed halo list if FIXED_HALO_GRIDS is False"
-            )
+            raise ValueError("You must provide halo_field if FIXED_HALO_GRIDS is False")
```
steven-murray: and here

steven-murray (commented):
OK, just a few minor naming things -- everything else looks great!

steven-murray (approved):

PR #569, opened 2025-10-10
Constant cleanup

daviesje:
The `Constants.h` file currently contains a bunch of macros for physical constants, array sizes and so on.

However, putting macros in header files means that those macros are applied to *any* file after it in the translation unit, which can include standard libraries, as well as many other places we don't want to mess with. I ran into errors with these macros when trying to implement some matrix math into the halo property sampling (with macros such as `C` and `D` which appear in BLAS libraries a lot).

This PR replaces many of the macros in `Constants.h` in a few ways:
- Where there is a standard library with the constant (i.e PI) we simply use that
- Where the macro was only used in one file, we move it to the `.c` file where it's needed
- We create a singleton struct (backend only) with most of the physical constants used across the backend. Some of these constants have been changed slightly (third or greater decimal place) to reflect Astropy constants. I don't expect performance drops using a struct instead of a macro, since most compilers should optimize them away. But I will keep an eye on it since they do appear in very hot functions.


src/py21cmfast/src/Constants.h
```diff
@@ -8,220 +8,107 @@
...
+} PhysicalConstants;
+
+extern const PhysicalConstants physconst;
+
+// Below are a few leftover macros, where they were used across multiple files or
+// depended on input parameters.
+
+// BEWARE: Since these macros are defined in a header, they *can* be applied to
+// any code included by 21cmFAST, e.g. if file A includes Constants.h, then includes
+// file B which includes fftw.h, the macros do find/replaces on fftw.h during compilation.
```
steven-murray: Thanks for adding the explanation. I wonder if we shouldn't 'namespace' our own macros... like give each a prefix `_21CMFAST_`?

src/py21cmfast/src/HaloField.c
```diff
@@ -84,15 +84,17 @@ int ComputeHaloField(float redshift_desc, float redshift, InitialConditions *box
         // store highly used parameters
         int grid_dim = simulation_options_global->DIM;
         int z_dim = D_PARA;
+        int dim_arr[3] = {grid_dim, grid_dim, z_dim};
```
steven-murray: sorry to be persnickety, but could we call this variable the same as it is in other c modules? I think there it's called `box_dim`

PR #571, opened 2025-10-29
add approximate mean adjustment

daviesje:
Setting the flag `HALO_SCALING_RELATIONS_MEDIAN` treats the scaling relations provided in the parameters as medians (or log-means) of their log-normal distributions, rather than means. This means that scatter imparts a `exp(sigma^2/2)` factor on the mean of each distribution.

There was a bug in the integrated components of our source fields, which did not account for this as they lack any stochasticity.

This change adds a shift to the mean in the `HaloBox` integrated components when the flag is on, in a similar manner to the way they are subtracted from the stochastic components when the flag is off.

One important caveat is that we cannot perfectly reproduce the effect of the SFR scatter, which evolves with stellar mass, without significant changes. As a result, a lower limit on its effect is implemented here, using the scatter limit at high stellar mass.

### DIFF
--- a/.github/workflows/base_benchmarks.yml
+++ b/.github/workflows/base_benchmarks.yml
@@ -34,10 +34,11 @@ jobs:
           pip install ".[dev]"
 
       - name: Track base branch benchmarks with Bencher
+        env:
+          BENCHER_API_TOKEN: ${{ secrets.BENCHER_API_TOKEN }}
         run: |
           bencher run \
           --project 21cmfast \
-          --token '${{ secrets.BENCHER_API_TOKEN }}' \
           --branch main \
           --testbed ubuntu-latest \
           --threshold-measure latency \
--- a/.github/workflows/pr_benchmarks.yml
+++ b/.github/workflows/pr_benchmarks.yml
@@ -36,13 +36,13 @@ jobs:
           pip install ".[dev]"
 
       - name: Track base branch benchmarks with Bencher
+        env:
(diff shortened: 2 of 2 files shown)


src/py21cmfast/src/HaloBox.c
```diff
@@ -288,6 +288,20 @@ int set_fixed_grids(double M_min, double M_max, InitialConditions *ini_boxes, fl
     set_integral_constants(&integral_cond, consts->redshift, M_min, M_max, M_cell);
     double growthf = dicke(consts->redshift);
 
+    // If our scaling relations define a median, the scatter will will increase the mean value
+    // due to the asymmetry of the lognormal distribution, we mimic this in the
+    // sub-sampler component. NOTE: this will also occur if FIXED_HALO_GRIDS is true, and it's
+    // not obvious whether that's what a user would want.
+    // TODO: redesign the USE_HALO_FIELD, FIXED_HALO_GRIDS, SCALING_RELATIONS_MEDIAN logic
```
steven-murray: This seems kind of critical -- if this is something we want to do, then we probably should do it before v4 or else we'll have major API problems...

steven-murray (approved):
Thanks @daviesje, this all looks good to me. I agree that at the moment the whole scaling median/mean thing (and halo field/fixed grids) is all a bit tricky. It would be nice to consider a refactor of this to be simpler, but as I said in my comment that would be API breaking. If we want to do this in any v4.x version we should think about it now.

PR #575, opened 2025-11-06
docs: update readme badges and main image

steven-murray:
This updates some outdated badges in the readme, and changes the hero image on the readme to the (much more striking, if I may say so) angular lightcone.


daviesje (approved):
nice!

PR #578, opened 2025-11-21
Get transfer functions from CLASS (for real this time)

jordanflitter:
The goal of this PR is to pass the transfer functions from `CLASS` to the C-backend when `POWER_SPECTRUM="CLASS"`. This is done with a new `CosmoTables` class, which contains the following fields:
- `transfer_density` (`Table1D`)
       - `size` (`int`)
       - `x_values` (`nd.array`)
       - `y_values` (`nd.array`)
- `transfer_vcb` (`Table1D`)
       - `size` (`int`)
       - `x_values` (`nd.array`)
       - `y_values` (`nd.array`)

There is also a new field to `SimulationOptions`, called `K_MAX_FOR_CLASS`, which controls the highest wavenumber in the `CLASS` calculation. For best precision, we need quite large values to compute the CHMF for the smallest halos (especially when `USE_MINI_HALOS=True`), but this requires a longer time for `CLASS` to finish (about 30 seconds). Impatient users can set `K_MAX_FOR_CLASS` to lower values to run `CLASS` faster, making the actual power spectrum that is used in the backend to be less precise.

`CosmoTables` is a new field in `InputParameters` (but it is not a sub-class of `InputStruct`). This was made in order to leave the current internal and external API intact (for example `evolve_input_stucts` automatically updates the `cosmo_tables` field, if necessary). So similarly to other fields of `InputParameters`, `cosmo_tables` is also broadcasted globally to the C-backend. The difference however is that the broadcasted interpolation tables are required to be allocated and ultimately freed when `21cmFAST` returns (this includes macro functions like `run_coeval` and `run_lightcone`, or any other function from `single_field.py` or `cfuncs.py`).

Sadly, at the moment, there appears to be a segfault issue that could be thrown the second time `21cmFAST` runs. It happens during `Broadcast_struct_global_all`, but only after `Free_cosmo_tables_global` has returned. The line that causes the segfault is the `memcpy` command, which makes me suspect that the memory to `cosmo_tables` is somehow garbage collected by `numpy` when `Free_cosmo_tables_global` returns. This will have to be fixed.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -91,6 +91,7 @@ setup(
         "classy",
         "cyclopts",
         "tomlkit",
+        "hmf",
     ],
     extras_require={"tests": test_req, "docs": doc_req, "dev": test_req + doc_req},
     setup_requires=["cffi>=1.0", "setuptools_scm"],


src/py21cmfast/drivers/coeval.py
```diff
@@ -635,11 +624,12 @@ def _obtain_starting_point_for_scrolling(
     inputs: InputParameters,
     initial_conditions: InitialConditions,
     photon_nonconservation_data: dict,
-    cache: OutputCache,
-    regenerate: bool,
     minimum_node: int | None = None,
+    **iokw,
```
steven-murray: Again, I prefer explicit arguments here too. The `**kw` arguments are useful when either:

1. You're immediately passing on a large subset of arguments to another function without much (preferably without any) amendment
2. The function parameters change a lot over (development) time, and there are many different viable combinations of parameters. In that case its sometimes easier to have a `**kw` and deal with the combinations via logic in the function. But this usually means you messed up the architecture earlier on.

src/py21cmfast/drivers/lightcone.py
```diff
@@ -403,18 +403,15 @@ def _run_lightcone_from_perturbed_fields(
     apply_rsds: bool,
     n_rsd_subcells: int,
     halofield_list: list[HaloCatalog],
-    regenerate: bool | None = None,
-    cache: OutputCache = _ocache,
     cleanup: bool = True,
     write: CacheConfig = _cache,
     progressbar: bool = False,
     lightcone_filename: str | Path | None = None,
+    **iokw,
```
steven-murray: Here it's fine :-D

src/py21cmfast/src/cosmology.c
```diff
@@ -134,67 +128,44 @@ double transfer_function_White(double k) {
...
-            }
-        }
-        fclose(F);
-
-        LOG_SUPER_DEBUG("Read CLASS Transfer file");
+        kclass = cosmo_tables_global->transfer_density->x_values;
+        Tmclass = cosmo_tables_global->transfer_density->y_values;
+        Tvclass_vcb = cosmo_tables_global->transfer_vcb->y_values;
+        size_density = cosmo_tables_global->transfer_density->size;
+        size_vcb = cosmo_tables_global->transfer_vcb->size;
```
steven-murray: can `size_density` and `size_vcb` ever be different?

src/py21cmfast/wrapper/classy_interface.py
```diff
@@ -210,7 +285,145 @@ def compute_rms(
...
+        self.theta_cmb = self.T_cmb / 2.7
+
+        self.f_nu = max(self.Omega_nu / self.Omega_m, 1e-10)
+        self.f_baryon = max(self.Omega_b / self.Omega_m, 1e-10)
+
+        self._set_parameters()
+
+    def _set_parameters(self):
+        """Set constants used in the EH transfer function (matches TFset_parameters in cosmology.c)."""
+        f_nu = self.f_nu
```
steven-murray: I wonder actually if it's not just better to depend on `hmf`... it already has the EH transfer function in it, and also a CAMB function, and BBKS, and...

src/py21cmfast/wrapper/inputs.py
```diff
@@ -290,6 +300,102 @@ def from_subdict(cls, dct, safe=True):
...
+        elif (self is not None and other is None) or (
+            self is None and other is not None
+        ):
+            return False
+        else:
+            return (
+                self.size == other.size
+                and np.all(self.x_values == other.x_values)
+                and np.all(self.y_values == other.y_values)
+            )
```
steven-murray: This is given to you for free by `attrs`, so you don't need to define it. You might have to pass `eq=attrs.cmp_using(eq=np.array_equal)` to the `field()` calls for the numpy arrays. See https://www.attrs.org/en/stable/comparison.html

src/py21cmfast/wrapper/inputs.py
```diff
@@ -290,6 +300,102 @@ def from_subdict(cls, dct, safe=True):
...
+
+    @cached_property
+    def ctab_pointer(self):
+        """Cached pointer to the memory of the object in C."""
+        ctab = ffi.new("Table1D *")
+        ctab.size = self.size
+        ctab.x_values = ffi.cast("double *", ffi.from_buffer(self.x_values))
+        ctab.y_values = ffi.cast("double *", ffi.from_buffer(self.y_values))
+        return ctab
```
steven-murray: In other classes like this, I've called this attribute the `cstruct`. Can we do this here as well?

steven-murray (commented):
Thanks very much @jordanflitter this is looking great! I have a few comments (see below) most of which I think you can easily handle.

src/py21cmfast/drivers/coeval.py
```diff
@@ -391,10 +391,8 @@ def evolve_halos(
     inputs: InputParameters,
     all_redshifts: list[float],
     write: CacheConfig,
-    initial_conditions: InitialConditions,
-    cache: OutputCache,
-    regenerate: bool,
     progressbar: bool = False,
+    **kw,
```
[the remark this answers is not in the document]
jordanflitter: Alright. The reason why I did this change is because `iokw` now has a new entry (`free_cosmo_tables`), and if the function doesn't have `**kw` argument and 'free_cosmo_tables' is not explicitly listed as a possible argument, an error is raised when the function is called. I changed it anyway according to your comment, the way I did is the following:
1. `_obtain_starting_point_for_scrolling` has explicit inputs `cache` and `regenerate`, but no 'free_cosmo_tables'. This means that we can't use `**iokw` as before in `lightcone.py`.
2. `evolve_halos` has explicit inputs `cache`, `regenerate` and  `free_cosmo_tables`. This enables to use `**kw` when this function is called.

src/py21cmfast/src/cosmology.c
jordanflitter: Right now they have the same size because they are of the size of `k_transfer` from `classy_interface.py`. I made the interpolation tables to be similar to what we had in `Transfer_z0.dat` where both transfer functions were of the same size. I think it's okay to leave it like this, if in the future we decide to have different sizes for the transfer functions (we are no longer limited that they must be of the same size, like we had with `Transfer_z0.dat`)

src/py21cmfast/wrapper/inputs.py
jordanflitter: Done!

steven-murray (approved):

PR #584, opened 2026-01-02
ci: update bencher to cope with forks

steven-murray:
This uses the docs [here](https://bencher.dev/docs/how-to/github-actions/#pull-requests-from-forks) to implement benchmarking with forks (so things like dependabot work...)


jordanflitter (approved):
Thanks @steven-murray, looks good to me, I have nothing to comment about.

PR #589, opened 2026-01-21
Add logo files and ReadMe update

nikos-triantafyllou:


steven-murray (commented):
Thanks @nikos-triantafyllou! I'm good with the workaround for now, though yeah we should definitely look at improving it soon. 

Quick couple of comments: I think we should remove all the experimental logos, so as not to bloat the repo. Keep the transparent background png, and the square logo, preferably in transparent as well. I think that's all we need. I will say that keeping the black-background  ones around will be useful (perhaps we can attach them as media on the website we make eventually). But I'd remove all but the transparent horizontal + square logos for now.

steven-murray (approved):

PR #604, opened 2026-02-13
Optional quantities

steven-murray:
This PR adds the general ability for us to have "optional quantities" that can be output for the user, but otherwise are not required. For starters, I've added the three easiest of these (mean free path, kinetic_temp_neutral and kinetic_temperature). 

This adds a new InputStruct, but since it is a simple one, old boxes and templates should all read in just fine.

### DIFF
--- a/src/py21cmfast/src/IonisationBox.c
+++ b/src/py21cmfast/src/IonisationBox.c
@@ -538,7 +538,9 @@ double set_fully_neutral_box(IonizedBox *box, TsBox *spin_temp, PerturbedField *
                 box->neutral_fraction[ct] =
                     1. - spin_temp->xray_ionised_fraction[ct];  // convert from x_e to xH
                 global_xH += box->neutral_fraction[ct];
-                box->kinetic_temperature[ct] = spin_temp->kinetic_temp_neutral[ct];
+                if (optional_quantities_global->kinetic_temperature) {
+                    box->kinetic_temperature[ct] = spin_temp->kinetic_temp_neutral[ct];
+                }
             }
         }
         global_xH /= (double)HII_TOT_NUM_PIXELS;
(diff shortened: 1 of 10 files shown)


src/py21cmfast/src/IonisationBox.c
```diff
@@ -1180,48 +1191,55 @@ void set_ionized_temperatures(IonizedBox *box, PerturbedField *perturbed_field,
                       HII_D_PARA};
 
     unsigned long long int idx;
+    if (optional_quantities_global->kinetic_temperature) {
```
jordanflitter: Notice that the sole purpose of `set_ionized_temperatures` is to determine the `kinetic_temperature`, so it is best to put this condition when `set_ionized_temperatures` is called (from `ComputeIonizedBox`) rather than to place this condition in `set_ionized_temperatures`.

src/py21cmfast/wrapper/inputs.py
```diff
@@ -1460,6 +1460,47 @@ def _node_redshifts_converter(value) -> tuple[float] | None:
     return (float(value),)
 
 
+@attrs.define(kw_only=True, frozen=True)
+class OptionalQuantities(InputStruct):
```
jordanflitter: Does it really need to be a sub-class of `InputStruct`?

jordanflitter (commented):
Thanks @steven-murray ! I made some minor comments/questions. My major comments/questions are the following:

1. I understand the overall logic, but I wish, if it’s possible, to include the optional quantities in a more elegant way. For example, in the (very near) future, I’d like to include MANY optional quantities from `SpinTemperatureBox.c` (e.g. `x_alpha`, Lya flux, `tau_GP`, heating/cooling rates, essentially all the physical quantities that the code computes), as it will be very illuminating to see how all of these IGM’s properties evolve over time (with `run_global_evolution`). If we keep specifying manually each possible optional quantities, we will end up with a struct that contains A LOT of fields, where most of them will be discarded (unless we `run_global_evolution`). So what I have in my mind is that `run_coeval` will have a `quantities` argument (similarly as `run_lightcone`). Then, based on the user’s specified quantities + required quantities from each output struct, we will have a list of all the quantitates that we do want to keep. There’s a different question now of how to pass that list elegantly to the backend. One idea that I have from the top of my had is to send the backend a long string that contains all the desired fields, separated by a dash (or any other separator you have in mind), e.g. “neutral_fraction-brightness_temp”. Then the C code could easily understand if a given quantity should be stored as an output or not. Similarly, the wrapper will understand if it needs to allocate memory for that quantity. What I’m suggesting here might be an overkill for this preliminary PR, especially when considering its urgency, but maybe something between the current implementation and this idea could be done.
2. I don’t understand why `kinetic_temp_neutral` from `TsBox` is an optional quantity. If `USE_TS_FLUCT=True`, that quantity is absolutely necessary for the evaluation of the spin temperature. Otherwise, if `USE_TS_FLUCT=False`, no `TsBoxes` are made, and the C code in `IonisationBox.c` already knows what to do when `kinetic_temp_neutral` is not available.

PR #605, opened 2026-02-13
perf: don't allocation unnormalised_nion if unnecessary

steven-murray:
Removes allocation of `unnormalized_nion` when using Lagrangian source models, which is totally legit according to Mr. Claude.

### DIFF
--- a/src/py21cmfast/wrapper/outputs.py
+++ b/src/py21cmfast/wrapper/outputs.py
@@ -1483,14 +1483,12 @@ class IonizedBox(OutputStructZ):
             if (
                 self.matter_options.mass_dependent_zeta
                 and self.astro_options.USE_MINI_HALOS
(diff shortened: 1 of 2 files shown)


src/py21cmfast/wrapper/outputs.py
```diff
@@ -1481,14 +1483,12 @@ def get_required_input_arrays(self, input_box: OutputStruct) -> list[str]:
             if (
                 self.matter_options.mass_dependent_zeta
```
jordanflitter: I think the `mass_dependent_zeta` condition can be removed, since if `USE_MINI_HALOS=True` then by definition `mass_dependent_zeta=True`

jordanflitter (approved):
Thanks @steven-murray (and Mr. Claude!). I made two minor comments, but overall looks good to me

PR #607, opened 2026-02-16
ci: update pypi publish branch

steven-murray:
Updates the PyPI release action to not be master which is deprecated


jordanflitter (approved):
It's the best!

PR #612, opened 2026-02-16
ci: only create sdists

steven-murray:
I think this is the one. The problem was that the build creates both a wheel (binary format) and a source distribution. Since this code is not cross-platform (i.e. it has C code) we can only create the source dist, not the wheel.


jordanflitter (approved):
I am not sure then why it worked until now.

Also, I have one comment, it's better to change `name` to `Build source distribution` :)

PR #613, opened 2026-02-16
Ability to inject initial conditions (only density field)

jordanflitter:
This PR implements two new more features.

The first one is the ability to inject `hires_density_array` to `p21c.compute_initial_conditions`. If a non-zero array is injected, then the usual Gaussian sampling in Fourier space is skipped, and then the rest of the computation continues. As before, this modified `initial_conditions` can then be used as an input to `run_coeval` (or `run_lightcone`) in order to make a "roundtrip test", something that could be useful for @nikos-triantafyllou's project.

The second new feature is the ability to `run_global_evolution` with `delta_z0` which is not necessarily 0. This could be helpful in making (speedy) linear perturbation theory* with `21cmFAST`.

*Note that this is a scale-independent perturbation theory, since we currently assume that the density field evolves in a scale-independent manner (as we simply multiply it by the scale-independent growth factor). This is actually not true in the dark ages, where the baryon density field evolves in a scale-dependent manner (unlike CDM, which roughly evolves in a scale-dependent manner), but this will wait for another PR


src/py21cmfast/drivers/coeval.py
```diff
@@ -840,11 +840,15 @@ def _setup_ics_and_pfs_for_scrolling(
...
     progressbar: bool,
+    delta_z0: float | None = None,
     **iokw,
 ) -> tuple[InitialConditions, list[PerturbedField], list[HaloCatalog], dict]:
     if initial_conditions is None:
         initial_conditions = sf.compute_initial_conditions(
-            inputs=inputs, write=write.initial_conditions, **iokw
+            inputs=inputs,
+            write=write.initial_conditions,
+            hires_density_array=delta_z0,
```
steven-murray: This doesn't seem right... why are we passing a float to the hires_density_array??

steven-murray (commented):
Thanks @jordanflitter this is almost there, we discussed just now the following:

* Changing the name of the `hires_density_array` parameter
* Changing the `with_value_not_computed` to be private
* Updating `compute_xray_source_field` to use your approach
* Adding your paper (and others) to `show_references()`
* Adding a small section to a new tutorial that promotes your new feature.

steven-murray (approved):

PR #616, opened 2026-02-19
docs: improve input parameter docs

steven-murray:
Updates to the documentation:

* Improves the new `models.rst` documentation to be more "human friendly"
* Fixes the main index page of the documentation so that the API appears in the left-hand scrollbar
* Updates the input parameter docstrings so that inner lists appear correctly
* Makes some attributes of the `InputStruct` private because they were never meant to be user-facing and I realized that they clutter the documentation with methods that shouldn't be exposed.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -84,7 +84,7 @@ autoapi_options=[
     'undoc-members', 
     'show-inheritance', 
     'show-module-summary', 
-    'special-members', 
+    #'special-members', 
     'imported-members',
     'inherited-members',
 ]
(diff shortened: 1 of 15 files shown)


jordanflitter (approved):
Thanks @steven-murray for all these changes! There were MANY changes, mostly in the docs and attribute naming, and since all the tests passed, I admit I didn't bother going on each every change (I also trust you knew what you were doing). I have one comment regarding a change in `updates_from_v3.rst`, but other than that I think this could be merged.

Issue #364, opened 2024-03-28
Move Photon Conservation Model to Python


daviesje:
The current implementation of the redshift-dependent photon conservation approximation spans the Python Wrapper and C backend in a very awkward way, relying on many globally defined arrays which have to be passed back and forth. This makes changes to the model very difficult.

Since the model has to run a calibration simulation from the wrapper, and otherwise consists of relatively light array operations, I see no reason not to migrate the entire model to the wrapper. In this way we can simply pass a structure containing all the photon conservation info into `ionize_box`.

steven-murray:
We'll do this later.

PR #617, opened 2026-02-20
fix: segfault when using photoncons

### LINKED ISSUE #593: [BUG] Segfault when using photon conservation correction
The photon conservation correction doesn't work now. I get a segfault while running the following script:

```python
import py21cmfast as p21c
from tempfile import mkdtemp

inputs = p21c.InputParameters.from_template(
    ["latest", "tiny"], random_seed=1234
).evolve_input_structs(PHOTON_CONS_TYPE="z-photoncons")

p21c.run_coeval(
    inputs = inputs,
    cache = p21c.OutputCache(mkdtemp()),
    write = p21c.CacheConfig.last_step_only(),
    progressbar = True
)
```

Notes that if `PHOTON_CONS_TYPE` is
1. `"no-photoncons"`: no segfault (everything works fine).
2. `"z-photoncons"`: segfault at the "Evolving Astrophysics" progress bar (i.e. the photoncons initialization succeeded).
3. `"alpha-photoncons"`: segfault during photoncons initialization.
4. `"f-photoncons"`: segfault during photoncons initialization.

steven-murray:
Fixes #593


jordanflitter (commented):
Thanks @steven-murray for looking into this! I don't understand enough the details of how the photon non-conservation correction works to appreciate this PR :)
(I'll have to study more what's going on there one day....)

Can we really tell for sure that these changes fix #311, or even #593? Besides testing this feature in `test_integration_features.py` (at redshift 18! Way before reionization starts and this feature becomes meaningful!), we don't really have a robust test module, e.g. `test_photoncons.py`.

In my opinion, the entire photon non-conservation calculations in `21cmFAST` should go under a major revision, making the code clearer and up to the standards we normally apply in other parts of the code. I don't make an issue on that since there is something similar already in #364. I believe that the robust `test_photoncons.py` shall be included in such a project.

Still, I think the very least we can do at the moment is to simply check that there are no segfaults when running a small coeval box with this feature on (for the three variations that the code probably supports), and finishing the simulation after reionization has ended. I am not sure that `test_integration_features.py` is the best place for this test. I think that maybe it's a good opportunity for us to have a first `test_photoncons.py`, with only this test in the meantime.

docs/conf.py
```diff
@@ -11,46 +11,6 @@
 # stack trace, whereas if it breaks inside the autodoc step, it's harder to debug.
 import py21cmfast
 
-class ClassDecoratedDocumenter(Documenter):
```
jordanflitter: What is the reason for the removal of this module?

jordanflitter (approved):
Thanks @steven-murray I'm glad we finally have a test that checks that this feature actually works! I made some really minor comments, but the PR looks good to me and is ready to be merged

PR #618, opened 2026-02-20
feat: allow random_seed and node_redshifts to be set in TOML profiles, with CLI support

### LINKED ISSUE #559: [Feature Req.] Ability to store node_redshifts in the TOML spec

It's annoying to have to read a parameter TOML file, then read in a corresponding ICs box to get the appropriate node redshifts, because they don't exist in the TOML.

Ability to save the node redshift info in the TOML. I think we could support either, or both of, the following specs:

```
[node_redshifts]
func: get_logspaced_redshifts
zmin:
zmax
zstep:
```

```
[node_redshifts]
[5,6,7,8,9,10,...]
```

It may be easier long-term to only store the explicit node_redshifts instead of allowing functions/parameters. Not sure.

steven-murray:
This updates the reading/writing of TOML "template" files to allow `random_seed` and `node_redshifts` to be written there.

In particular, this allows for _much_ easier loading of cached coeval boxes after running a simulation. E.g. in the run script you might have...

```python
import py21cmfast as p21c

inputs = p21c.InputParameters.from_template("EOS21", random_seed=1, node_redshifts=[5,6,7,8,9,10])
inputs.write_template("my_parameters.toml")
p21c.run_lightcone(inputs=inputs, write=True)
```

Now, after running that very large simulation, you would like to read some boxes with the nifty `RunCache` tool in a separate script. Before this PR you must do:

```python
import py21cmfast as p21c

inputs = p21c.InputParameters.from_template("my_parameters.toml", random_seed=1, node_redshfits=[5,6,7,8,9,10])
....
```

And hope that you kept the random seed and node redshifts in sync between the two scripts.

Now you can just do:

```python
import py21cmfast as p21c
inputs = p21c.InputParameters.from_template("my_parameters.toml")
...
```

By default if `mode='full'` in `write_template()` the full information is written (including random seed and node redshifts). However, if `mode="minimal"` then these are _not_ written. A minimal template is meant to be more like a "physical model" (like our built in templates) which doesn't try to specify what the exact random seed is.

## CLI: Embedding Node Redshifts and Random Seed in a Template

The `21cmfast template create` command can now embed `node_redshifts` and `random_seed` directly into the template file by specifying one or more of the following new options:

| Option | Default | Description |
|---|---|---|
| `--random-seed` / `--seed` | — | Random seed to embed in the template |
| `--zmin` | 5.5 | Minimum redshift of the node grid |
| `--zmax` | `Z_HEAT_MAX` from model | Maximum redshift of the node grid |
| `--zstep` | `ZPRIME_STEP_FACTOR` from model | Step factor (logspace) or step size (linear) |
| `--nz` | — | Fixed number of nodes; overrides `--zstep` |
| `--znode-func` | `logspace` | Spacing function: `logspace` or `linear` |

Example usage:

```bash
# Fully specified template with seed and log-spaced grid
21cmfast template create --template simple --out my.toml --seed 42 --zmin 5.0 --zmax 35.0 --zstep 1.01

# Log-spaced grid with fixed number of nodes
21cmfast template create --template simple --out my.toml --zmin 5.0 --zmax 30.0 --nz 30

# Linearly-spaced grid
21cmfast template create --template simple --out my.toml --zmin 5.0 --zmax 35.0 --zstep 1.5 --znode-func linear

# Linearly-spaced grid with fixed number of nodes
21cmfast template create --template simple --out my.toml --zmin 5.0 --zmax 35.0 --nz 30 --znode-func linear
```

The `--znode-func logspace` option calls `InputParameters.with_logspaced_redshifts` (which also gains a new `nz` parameter). The `--znode-func linear` option calls the new `InputParameters.with_linear_redshifts` method.

Documentation for the new options has been added to `docs/tutorials/cli_usage.rst`.


steven-murray:
@copilot please add the ability for the CLI command `21cmfast template create` to add node redshifts to a custom template (in the same format as this PR has already defined). Let the following command be possible:

```
21cmfast template create --zmin 5.0 --zmax 35.0 --zstep 1.01
```

but where `--zmax` has default of `Z_HEAT_MAX` from the inputs and `--zstep` has default of `Z_STEP_FACTOR` from the `inputs` (and `--zmin` has the standard static default defined in `inputs.py`). 

Furthermore, let there be another possible argument, `--nz` which replaces `--zstep` if given, and defines the number of redshifts in the array. 

Finally, let there be an argument `--zscroll-func` which defines the function used to specify the node redshifts. By default this is "logspace" (which ends up calling `InputParameters.with_logspaced_redshifts`). However, let it also be "linear" which should call a new method `with_linear_redshifts`. 

Add the new functionality to the CLI documentation in the `docs/tutorials/cli_usage.rst`.

src/py21cmfast/cli.py
```diff
@@ -308,15 +304,44 @@ def template_create(
     param_selection: ParameterSelection = ParameterSelection(),
     user_params: Parameters = Parameters(),
     mode: TOMLMode = "full",
+    zmin: Annotated[float | None, Parameter(name="--zmin")] = None,
+    zmax: Annotated[float | None, Parameter(name="--zmax")] = None,
+    zstep: Annotated[float | None, Parameter(name="--zstep")] = None,
+    nz: Annotated[int | None, Parameter(name="--nz")] = None,
+    zscroll_func: Annotated[
```
steven-murray: Let's actually change the name of this parameter to `znode_func` to be more consistent with naming throughout the code.

steven-murray (commented):
Thank you @copilot for the new feature. I have made one comment for modification, and there are two other things to accomplish also:

1. Please also add a `random_seed` (aliased to both `--random-seed` and `--seed`) argument to the `21cmfast template create` command. This way the user can get a fully specified template from this command.
2. Once the new feature and modification are implemented, ensure that the CI checks pass (including the pre-commit checks)

steven-murray:
@copilot please merge the main branch into this one, and ensure that all pre-commit checks pass.

steven-murray:
@jordanflitter I believe this one is ready for review again. Unfortunately after merging the current `main` branch, it looks like this PR udpates 33 files -- but these are just changes that happened in main that I hadn't yet merged into the v4.2 release branch. I've updated v4.2 now to be up to date with `main` but for some reason this PR still shows a diff for all these files that it doesn't really touch. The important differences are in the CLI and cli_usage.rst (and test_cli). 

I think the best test will be a quick play with it.

jordanflitter (changes requested):
Thank you very much @steven-murray, this is a very nice feature!

Before approving, there is one serious issue that needs to be addressed. The `cli_usage.rst` says that 
```
21cmfast template create --template simple --out my-simple.toml --zmin 5.0 --zmax 35.0 --zstep 1.5 --znode-func linear
```
should work, but it doesn’t! I could only make it work with
```
21cmfast template create --template simple --out my-simple.toml --nodez.min 5.0 --nodez.max 35.0 --nodez.step 1.5 --nodez.spacing linear
```
So this have to be fixed.

Besides that, I have some more minor comments.

1. The doc-string of `from_template` still says that `random_seed` "must be specified manually”. This needs to be changed.
2. `with_logspaced_redshifts` calls `get_logspaced_redshifts` while `with_linear_redshifts` doesn’t call a similar `get_linspaced_redshifts`, but rather `np.linspace` or `np.arange`. Is there a reason why we don’t use `np.logspace`? Note that `with_logspaced_redshifts` returns an array with a maximal value that is slightly higher than `zmax`, while `np.linspace` doesn’t. I think we should be consistent about this when constructing the `node_redshifts` array, whether we use linear or logarithmic spacing.
2. In `get_logspaced_redshifts`, `zstep_factor` is automatically determined from `ZPRIME_STEP_FACTOR`, but the logic in `with_linear_redshifts` is different (error is raised if both `nz` and `zstep` are not provided). Again, I think it would be better to have the same behavior in both functions. Also, note that we aim to remove `ZPRIME_STEP_FACTOR` from `simulation_options` (see #664). Maybe this PR is a good opportunity to do that.
3. I have some suggestions about the names of the arguments. Can we rename `zstep` in `with_linear_redshifts` to be `dz`? Also, I think it would be better if in the cli there won't be a general `zstep` argument, but rather something that matches the names in the python interface (`zstep_factor` and `dz`, one of them is ignored based on the state of `znode-func`). I think that way we can reduce confusion. For example, in the current naming, if the user doesn’t specify `--znode-func`, then in that case the user could get confused if their `zstep` refers to `zstep_factor` (in log spacing) or `dz` (in linear spacing). Forcing them to type the “correct” name for `zstep` could reduce their confusion.

jordanflitter:
Talking with @steven-murray, we agreed that:
1. Okay
2. Let's change implementation of `get_logspaced_redshifts`.
3. We'll leave things as they are, and just remove `ZPRIME_STEP_FACTOR` from the code in #664.
4. I'll let you think about it.

Also, we won't change the syntax of the cli, but the tutorial in `cli_usage.rst`

steven-murray:
OK, done!

I opted to change all step-parameters to be called "step". 

I also updated the CLI so that the full list of all model parameters are not printed out when using `--help`. However, they can be shown with `21cmfast params --help`.

jordanflitter (approved):
Perfect! Thank you very much! Green light from me :)

PR #621, opened 2026-02-21
feat: ability to run global-evolution through CLI

### LINKED ISSUE #614: run_global_evolution from cli
Currently the cli only allows to `run_coeval` and `run_lightcone`. It would be great if we could also `run_global_evolution` from the cli.

steven-murray:
Fixes #614 

@jordanflitter I suggest taking this for a spin with some examples to make sure everything acts as you think it should. 

I'm proposing we merge this into a "v4.2" branch because it's technically a new feature and so should bump minor version, but I think maybe we want to include some other features in v4.2 as well, and don't want to stop bug fixes from going into v4.1.

It is a very small feature though, so we could pass it off as a patch version if you think that's OK.

### DIFF
--- a/.github/workflows/fork_pr_benchmarks_run.yml
+++ b/.github/workflows/fork_pr_benchmarks_run.yml
@@ -40,12 +40,12 @@ jobs:
           pytest --benchmark-json results.json tests/test_integration_features.py
 
       - name: Upload Benchmark Results
-        uses: actions/upload-artifact@v6
+        uses: actions/upload-artifact@v7
         with:
           name: benchmark_results.json
           path: ./results.json
       - name: Upload GitHub Pull Request Event
-        uses: actions/upload-artifact@v6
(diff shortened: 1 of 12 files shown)


jordanflitter (commented):
Thanks @steven-murray for adding this feature! It is probably better to add it to v4.2 since it is a new (minor) feature, but I guess it's also okay to include it a new patch for v4.1, so I leave the decision to you :)

I played a bit with this new feature with the cli and overall things seem to work smoothly. Besides the comment/question that I had about `redshift-range`, my other comment is that we should probably spend some words on this new feature in `cli_usage.rst`. Also, I'm surprised that you didn't include this feature in a new test in `test_cli.py` (I don't understand how codecov didn't fail 😅)

jordanflitter (approved):
Thanks! Looks good to me (though the tests need to pass before merging..)

I guess we will modify the cli's `node_redshifts` input once we deal with #618.

PR #624, opened 2026-02-26
Initialize only once + tracking initializations state

### LINKED ISSUE #591: Move interpolation initializers to python
There are four types of interpolation tables used in the code:
1. Fixed tables that are found in the external files in the `_data` folder. They are initialized by calling `init_heat` (in `heating_helper_progs.c`).
2. Power spectrum initialization, this is done by calling `init_ps` (in `cosmology.c`). An interpolation table is used only if `POWER_SPECTRUM="CLASS"`, but other variables must be initialized in other scenarios.
3. `Sigma(M)` (and its derivative) initialization, this is done by calling `initialiseSigmaMInterpTable` (in `interp_tables.c`).
4. HMF integrals (e.g. SFRD), they are initialized by calling functions in `interp_tables.c`.

All of these initializations (and their deallocation) are done in C. Moreover, these initializations are not done efficiently, since they are done at each redshift snapshot and sometimes even more than once per snapshot (for example, `init_ps` is called in almost every C module which is accessible via the wrapper), whereas most of these tables are not redshift-dependent! As a consequence, the initializations are currently the bottleneck in `run_global_evolution`.

This kind of implementation was necessary for historical reasons, when the code was purely written in C. However, now that we have the python wrapper, we should manage the initializations more cleverly. We should initialize all the redshift-independent tables only one time at the beginning of the calculation, and in order to do that we could use public packages like `hmf`, `classy` or `camb`. In addition, we should store the tables in `CosmoTables` and use a quick linear interpolation (rather than `gsl`) in order to do the evaluation in the C code.

The only tables that should probably remain to be handled by the C code are the tables of the HMF integrals, since they are redshift-dependent, model-dependent (e.g. SFRD) and the grid of the interpolation depends on the box (e.g. the maximum/minimum over-density in the box is used for determining the grid for the conditional SFRD). I am open however to discussions on how to handle these tables at the python level.

Note that there is some extra discussion on this topic in #362.

jordanflitter:
Fix #591.

This PR deletes all the calls of `init_ps`, `initialiseSigmaMInterpTable` and `init_heat` from the C backend. The initializations (as well as the broadcast to C) are done in the wrapper only ONCE, regardless if we run a high level function, a single field function, or something from `cfuncs.py`.

I moved the nested wrapper in `cfuncs.py` to `_param_config.py`. The update wrapper (whose every line was written with blood, sweat and mostly tears) contains a dynamic `_InitManager` instance that is passed across different levels of functions via their `kwargs`, allowing the functions to inspect the state of the initializations and decide whether a given initialization must be performed or not. The decorators above the functions also give an idea for what must be initialized in order for the execution to complete.

Besides the more elegant design (according to me!), this architecture reduces the runtime of `run_global_evolution` by 50%! (because the sigma interpolation tables are not initialized every time a single field function is called)
Moreover, it cleans some memory leak bugs (`init_ps` was very often not accompanied with `free_ps` at the end, sometimes there were even calls to `init_ps` without actually using the power spectrum).

Other ideas that I had but still haven't implemented them:

- Update `performance.rst`.
- Use the data in `_InitManager` to raise an error (if something was not initialized, to prevent a segfault) or a warning if the computation was complete but something was left unfreed. While this is a nice feature in theory, the architecture of the updated wrapper will make it very difficult to test these scenarios (since the whole idea is that the wrapper uses `_InitManager` to decide what to initialize and what not.


steven-murray:
Thanks for the your comments @jordanflitter. Yes, it might be easiest to discuss this next week directly. You may be right that there is no real discernible difference between our methods, but I certainly _felt_ mine was much simpler as I was writing it haha.

jordanflitter:
Main improvements that were made with respect to the last pythonic implementation:
1. `init_manager` now contains `inputs`, and is set to `_InitManager(inputs=inputs)` (namely, all the initialization states are set to False) if the current `inputs` doesn't match `init_manager.inputs`.
2. Use `inspect.isgeneratorfunction(func)` to determine whether the called function is a generator or not.
3. `init_manager` also tracks the initialization of the recombination table, removing the call to `init_MHR` from the C code. Also, `free_MHR` is called, something that never happened before (bug?)
4. Use of a single decorator called `c_state_initializer` with multiple arguments, rather than multiple decorators with no arguments like before.
5. Updated performance plots (note: it seems like performance plots were slightly degraded after the merger with main, I'm not sure why, and I don't think we should spend time on figuring this out because it adds only up to 2 seconds in the total runtime of `run_global_evolution`).

jordanflitter:
The macos runs fail because in `test_halo_sampler.py::test_perturb_halos`, there is a mismatch of a single element out of 883 in total! (0.113%)
The mismatch happens because the relative tolerance in the test is 1e-5, while the relative tolerance in the faulty element is 4e-5.

This happens because in this PR we initialize the sigma interpolation tables only once, with `M_min=5e2` and `M_max=1e20`, instead of the tailor-made limits that were determined in the C code before this PR.

My suggestion is to increase the tolerance in the test from 1e-5 to 5e-5.

steven-murray (approved):

PR #628, opened 2026-03-06
fix: bug in which cstruct attrs weren't being set for CosmoTables

steven-murray:
@jordanflitter the problem here was that when you read floats etc. from an HDF5 file, the type that you get is `np.float64` rather than simply `float`, so we were missing setting these values on the backend  cstruct properly.


jordanflitter (approved):
Thanks @steven-murray!

PR #632, opened 2026-03-09
Reducing more memory with MINIMIZE_MEMORY

jordanflitter:
`mean_free_path` and `kinetic_temperature` are allocated and computed only if `MINIMIZE_MEMORY=False` (these fields are unnecessary for the evaluation of the brightness temperature). This could be useful to @DanielaBreitman.


steven-murray (commented):
I think this works, but could be made a little faster by moving some of the conditionals outside loops. See #604 for how I did this (but don't take all the logic from that PR)

steven-murray (approved):
I stand corrected -- you have the best implementation here.

PR #648, opened 2026-03-23
Fixed halo_sfr unit typo in tutorial

### LINKED ISSUE #634: [Question]Unit of lightcone output fields
What is the unit of the output HALO_SFR field? As it is cunstructed I enterpret it to be a density field, even tho the plots in the tutorials says the unit is Msolar/yr, but it should be Msolar/yr/Mpc^3 ?

Also, for z=7 I get rought mean SFR: 4.906106076951744e-10  Msolar/yr/Mpc^3 from halo_sfr but I expect the value to be roughly 10^-1 to 10^-3 Msolar/yr/Mpc^3 , so is the unit of HALO_SFR really  Msolar/s/Mpc^3?

Would be great with some clarification! 🚀

jordanflitter:
Fix #634.


steven-murray (approved):
LGTM

PR #650, opened 2026-04-01
ux: better error message for unknown array

steven-murray:
Very simply updates the error message for an unknown lightcone quantity to print out the possible quantities.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -1,8 +1,6 @@
 """Configuration options for the docs."""
 
 import os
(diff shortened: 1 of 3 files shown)


jordanflitter (approved):
Thanks @steven-murray, I'm not sure where the motivation for this PR came from (and why `docs/conf.py` was modified), but overall this looks good to me.

I think updating the tests could pass the coverage barrier (and also convince us that the new error message logic works as expected).

PR #671, opened 2026-04-22
Fixed some bugs in `perturb_halo_catalog`

### LINKED ISSUE #647: Bugs in perturb_halo_catalog
I used for the first time `perturb_halo_catalog`. I had to make many changes in `outputs.py` in order to be able to run this function.

1. In this [line](https://github.com/21cmfast/21cmFAST/blob/main/src/py21cmfast/wrapper/outputs.py#L898) I had to put `buffer_size: float | None = None`, similarly like we have [here](https://github.com/21cmfast/21cmFAST/blob/main/src/py21cmfast/wrapper/outputs.py#L784). The reason is because the wrapper (in `_config_params.py`) checks if this object had been computed and stored in the cache, so it calls `new()` without specifying `buffer`.
2. Then, I had to add [these lines](https://github.com/21cmfast/21cmFAST/blob/main/src/py21cmfast/wrapper/outputs.py#L806) below `PerturbedHaloCatalog.new()` such that `buffer` will not be `None` when the wrapper checks if the object is stored in the cache.
3. Lastly, I had to add [here](https://github.com/21cmfast/21cmFAST/blob/main/src/py21cmfast/wrapper/outputs.py#L962) that `star_rng`, `sfr_rng` and `xray_rng` are all required arrays, otherwise I got segfault!

I suppose that our tests with `perturb_halo_catalog` pass because we might not be using the cache in those tests (I haven’t looked yet into that). I have no idea why we don’t get segfault in the tests though.

jordanflitter:
Fixes #647.

### DIFF
--- a/.github/workflows/test_build_macos.yaml
+++ b/.github/workflows/test_build_macos.yaml
@@ -2,22 +2,19 @@ name: Test MacOS Build
 
 # Test on all pushes, except when the push is literally just a tag (because we
 # tag automatically via CI, and therefore there's no extra code in that push).
-# Also, only test on pull requests into main.
 on:
   push:
     tags-ignore:
       - 'v*'
   pull_request:
     branches:
-      - 'main'
+      - '*'
 
 
 jobs:
   tests:
     if: "!contains(github.event.pull_request.labels.*.name, 'auto-pr')"
-    env:
-      PYTHON: ${{ matrix.python-version }}
-      CC: gcc
+
     name: Test MacOS Build
     runs-on: ${{ matrix.os }}
     strategy:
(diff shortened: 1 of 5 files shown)


steven-murray (commented):
@jordanflitter the fixes look correct to me. However, it would be good also to have a test that would have triggered the segfault!

jordanflitter:
Thanks @steven-murray, I now simply added `regenerate=False` in `test_using_cached_halo_catalog`, it makes sense since the whole point of this test is to use the cache (we will fix the default logic of `regenerate` in #677). Had `regenerate` been False before this PR, we would have caught the segfault.

steven-murray (approved):

Issue #516, opened 2025-05-14
[BUG] Non-monotonic reionization when node redshifts are irregular


samgagnon:
**Describe the bug:**

**To Reproduce:**
1. Run a lightcone with node_redshifts defined as follows
    `node_redshifts = p21c.get_logspaced_redshifts(min_redshift  = min_redshift,
                             max_redshift  = max_redshift,
                             z_step_factor = 1.02)`
    `node_redshifts = np.concatenate((node_redshifts, np.array([4.9, 5.7, 6.0, 6.6, 7.0, 7.3])))`
    and the toml template provided in the attached file `sgh25.txt` (change file extension to .toml)
3. Read the average neutral fraction as a function of node redshift using `lc.global_quantities['neutral_fraction'][::-1]`
4. The neutral fraction does not increase monotonically with redshift. It exhibits discontinuities coinciding with the redshifts appended to the node redshift array.

**Expected behavior:**
Neutral fraction should increase monotonically with redshift.

**Details:**
 - 21cmFAST main branch synced to commit f6bf7ede844d48c8deb317618e8d1910ae45de91

[sgh25.txt](https://github.com/user-attachments/files/20209629/sgh25.txt)

steven-murray:
Ah, this is a strange usage. Probably we're just not forcing the `node_redshifts` to be sorted in `InputParameters` when we should.

steven-murray:
MWE from @samgagnon:

```python
import numpy as np
import py21cmfast as p21c

import logging, os

#    cache = p21c.OutputCache(args.cache)
cacheconfig = p21c.CacheConfig.off()
cache.list_datasets(
    kind=None,
    inputs=None,
    redshift=None,
    all_seeds=True,
)
p21c.config['HALO_CATALOG_MEM_FACTOR'] = 2.0

min_redshift=5.0
max_redshift=35.0

node_redshifts = p21c.get_logspaced_redshifts(
  min_redshift  = min_redshift,
  max_redshift  = max_redshift,
  z_step_factor = 1.02
)
  
node_redshifts = np.concatenate((node_redshifts, np.array([4.9, 5.7, 6.0, 6.6, 7.0, 7.3])))

inputs = p21c.InputParameters.from_template('./sgh25.toml',\
            random_seed=1,\
            node_redshifts=node_redshifts).evolve_input_structs(N_THREADS=4)

lcn = p21c.RectilinearLightconer.with_equal_cdist_slices(
[...]

Issue #600, opened 2026-02-11
[Feature Req.] Allow having USE_MINI_HALOS=True even if USE_TS_FLUCT or INHOMO_RECO are False


jordanflitter:
Currently we throw an error if `USE_MINI_HALOS=True`  and either `USE_TS_FLUCT` or `INHOMO_RECO` are `False`.

In our default settings, `USE_TS_FLUCT=False` and `INHOMO_RECO=False`, meaning we somewhat encourage the users to use this configuration, mostly for quickly generating coeval boxes at the end of reionization, while considering the ionizing radiation that comes from **only popII stars**. It makes sense to generalize this configuration and allow a quick generation of coeval boxes at the end of reionization, while considering the ionizing radiation that comes from **both popII and popIII stars**.

The reason why we don't allow this configuration is because the `M_TURN` of mini-halos depends on two quantities:
1. It depends on the LW flux, which is computed only when we compute the spin temperature (`USE_TS_FLUCT=True`).
2. It also depends on the ionization rate, `ionisation_rate_G12`, (through the "reionization feedback" on `M_TURN`), which is only computed when `INHOMO_RECO=True`, probably due to historical reasons (as the `INHOMO_RECO=True` feature was added before `USE_MINI_HALOS=True`).

We should allow the code to support having `USE_MINI_HALOS=True` even if `USE_TS_FLUCT=False` and `INHOMO_RECO=False`:
1. We can estimate the global LW flux by calling `run_global_evolution` (while setting `USE_TS_FLUCT=True`), and then use that value for computing the inhomogeneous `M_TURN`. This is of course not precise, so a warning can be raised in that scenario.
2. We should compute `ionisation_rate_G12` not only when `INHOMO_RECO=True`, but also if `USE_MINI_HALOS=True`.

PR #681, opened 2026-04-30
Homogeneous recombination

jordanflitter:
## Summary

This PR implements a comprehensive refactoring of recombination model handling in 21cmFAST, replacing the boolean `INHOMO_RECO` flag with a new enumerated `RECOMB_MODEL` variable and introducing optimized support for homogeneous recombination.

#### What Changed
- Replaced `INHOMO_RECO` (boolean) with `RECOMB_MODEL` enum with three options:
  - `"none"`: No recombination
  - `"homogeneous"`: Spatially uniform recombination (new diagnostic mode)
  - `"inhomogeneous"`: Spatially-varying recombination (original physics model)

#### Backward Compatibility
To maintain API compatibility, the old `INHOMO_RECO` boolean still works:
- **Automatic conversion**: If user sets `INHOMO_RECO=True`, the code automatically sets `RECOMB_MODEL="inhomogeneous"` and shows a deprecation warning
- **Validation**: Conflicts between old and new parameters raise clear error messages
- **Property accessor**: `INHOMO_RECO` property still available and correctly infers from `RECOMB_MODEL`

**Example:**
```python
# Old way (deprecated but still works)
opts = AstroOptions(INHOMO_RECO=True)  # Shows warning, sets RECOMB_MODEL="inhomogeneous"

# New way (recommended)
opts = AstroOptions(RECOMB_MODEL="inhomogeneous")  # No warning
```

#### Motivation
For diagnostic testing of recombination physics, we need a spatially uniform recombination field. This new mode enables:
- Testing whether temporal implementation of `cumulative_recombinations` can causes kinks in the global `neutral_fraction` (see #516).
- Isolating recombination physics from density perturbations
- Includes runtime and memory reductions when `RECOMB_MODEL="homogeneous"`

#### Key Optimizations

**1. Single-value storage**
- `cumulative_recombinations` allocated as (1,1,1) instead of (HII_DIM, HII_DIM, HII_D_PARA)
- Memory savings for diagnostic tests

**2. Compute once, not per-cell**
```c
if (astro_options_global->RECOMB_MODEL == 1) {
    // Compute dNrec ONCE using global averages
    double dNrec = splined_recombination_rate(consts->stored_redshift, global_photionization_rate) * 
                   consts->fabs_dtdz * consts->dz * (1. - global_xHI);
    box->cumulative_recombinations[0] = previous_ionize_box->cumulative_recombinations[0] + dNrec;
} else {
    // Original: loop over all cells for inhomogeneous recombination
    // ...
}
```
- Eliminates `O(HII_DIM³)` loop for homogeneous case

**3. No filtering needed**
- Enforces `CELL_RECOMB=True` for homogeneous mode (validation added)
- No need for FFT filtering on uniform field
- Skips power spectrum computation for `cumulative_recombinations` in test suite

#### Implementation Details
- Modified `src/py21cmfast/src/IonisationBox.c`: Separate code paths for homogeneous vs. inhomogeneous
- Modified `src/py21cmfast/wrapper/outputs.py`: Conditional allocation based on `RECOMB_MODEL`
- Added validation in `src/py21cmfast/wrapper/inputs.py`: Prevent `CELL_RECOMB=False` with homogeneous
- Updated test infrastructure to handle field selection per recombination model

### Testing

✅ All existing tests pass (backward compatibility verified)
✅ New tests for `RECOMB_MODEL` validation and property inference
✅ Integration tests for `RECOMB_MODEL="homogeneous"` mode
✅ Test data properly handles field selection for power spectrum comparison

### Files Modified
- `src/py21cmfast/wrapper/inputs.py` - New enum and validation logic
- `src/py21cmfast/src/IonisationBox.c` - Separated homogeneous/inhomogeneous code paths
- `src/py21cmfast/wrapper/outputs.py` - Conditional allocation for cumulative_recombinations
- `tests/test_input_structs.py` - New validation tests
- `tests/produce_integration_test_data.py` - Field selection logic + new configuration for the new feature ("homo")
- Template files - Updated to use `RECOMB_MODEL` instead of `INHOMO_RECO`

### Benefits
- 🚀 **Performance**: Homogeneous recombination runs should be faster (no looping, no FFT filtering)
- 💾 **Memory**: Reduces `cumulative_recombinations` allocation from full box to single cell
- 🧪 **Testing**: Enables diagnostic tests to isolate recombination physics
- 📦 **API**: Backward compatible—existing code continues to work
- 🎯 **Clarity**: Explicit enum is more maintainable than boolean flag

## Closing statement
We still have `CELL_RECOMB` as an independent boolean flag. In my opinion, `CELL_RECOMB=False` (now possible only when `RECOMB_MODEL="inhomogeneous"`) should become the 4th option of `RECOMB_MODEL`, something like `RECOMB_MODEL="inhomogeneous-filtered"`. I have not made this change as part of this PR, but I think we should consider this design for the long term.


joss-paper/paper.md
steven-murray, line 1: We shouldn't change this file because it's a historical artifact.

src/py21cmfast/drivers/coeval.py
```diff
@@ -922,7 +922,10 @@ def _get_required_redshifts_coeval(
     # Turn into a set so that exact matching user-set redshift
     # don't double-up with scrolling ones.
     if (
-        (inputs.astro_options.USE_TS_FLUCT or inputs.astro_options.INHOMO_RECO)
+        (
+            inputs.astro_options.USE_TS_FLUCT
+            or inputs.astro_options.RECOMB_MODEL != "none"
+        )
```
steven-murray: Actually I think this logic should be replaced by `inputs.evolution_required`

src/py21cmfast/wrapper/__init__.py
```diff
@@ -75,8 +75,8 @@
 >>> all_boxes = p21.run_coeval(
 >>>                 redshift=[7,8,9],
 >>>                 simulation_options=p21.SimulationOptions(HII_DIM=100),
->>>                 astro_options=p21.AstroOptions(INHOMO_RECO=True),
->>>                 do_spin_temp=True
+>>>                 astro_options=p21.AstroOptions(RECOMB_MODEL="inhomogeneous"),
+>>>                 USE_TS_FLUCT=True
```
steven-murray: This is actually just wrong -- this doesn't exist as a parameter any more.

src/py21cmfast/wrapper/inputs.py
```diff
@@ -1200,20 +1211,63 @@ class AstroOptions(InputStruct):
...
+            warnings.warn(
+                "INHOMO_RECO is deprecated and will be removed in a future version. "
+                "Please use RECOMB_MODEL directly instead.",
+                stacklevel=2,
+            )
+            return "inhomogeneous"
+        elif self._INHOMO_RECO is False:
+            return "none"
+        else:  # _INHOMO_RECO is None (not provided)
+            return "none"
```
steven-murray:
```suggestion
        if self._INHOMO_RECO is None:
            return "none"
        
        warnings.warn(
              "INHOMO_RECO is deprecated and will be removed in a future version. "
              "Please use RECOMB_MODEL directly instead.",
              stacklevel=2,
        )
        
        return "inhomogeneous" if self._INHOMO_RECO else "none"
```

src/py21cmfast/wrapper/photoncons.py
```diff
@@ -280,8 +280,8 @@ def calibrate_photon_cons(
     from ..drivers.single_field import compute_ionization_field, perturb_field
 
     # Create a new astro_params and astro_options just for the photon_cons correction
-    # NOTE: Since the calibration cannot do INHOMO_RECO, we set the R_BUBBLE_MAX
-    #   to the default w/o recombinations ONLY when the original box has INHOMO_RECO enabled.
+    # NOTE: Since the calibration cannot do RECOMB_MODEL != "none", we set the R_BUBBLE_MAX
+    #   to the default w/o recombinations ONLY when the original box has RECOMB_MODEL != "none".
```
steven-murray:
```suggestion
    # NOTE: Since the calibration cannot do recombinations, we set the R_BUBBLE_MAX
    #   to the default w/o recombinations ONLY when the original box has recombinations.
```

tests/test_input_structs.py
```diff
@@ -234,6 +236,52 @@ def test_bad_inputs(self):
...
             AstroOptions(USE_EXP_FILTER=True, HII_FILTER="sharp-k")
 
+    @pytest.mark.parametrize("recomb_model", ["none", "homogeneous", "inhomogeneous"])
+    def test_recomb_model_basic(self, recomb_model):
+        """Test basic RECOMB_MODEL usage without INHOMO_RECO."""
+        opts_none = AstroOptions(RECOMB_MODEL=recomb_model)
+        assert recomb_model == opts_none.RECOMB_MODEL
+        assert opts_none.INHOMO_RECO is False if recomb_model == "none" else True
+
+    def test_inhomo_reco_deprecated_warning(self):
```
steven-murray, line 247: Could consider using this library so we remember to remove the deprecated code: https://pypi.org/project/deprecation/

steven-murray (commented):
This all looks very good, thank you @jordanflitter. Only a few small comments.

joss-paper/paper.md
jordanflitter: Oh I'm sorry I didn't know I was messing with an historical artifact! 😱

I fixed it back now.

src/py21cmfast/drivers/coeval.py
jordanflitter: Done!

Note that `inputs.evolution_required` is not exactly the same as above because it also contains `or inputs.astro_options.USE_MINI_HALOS`. This is actually quite redundant because we force the other two conditions when `USE_MINI_HALOS=True` (see #600), I guess we will fix it when we deal with that issue.

src/py21cmfast/wrapper/__init__.py
jordanflitter: Right! I modified that file to match better with our current API (including the changes you made in `with_logspaced_redshifts` in #618).

I wonder why this file and its content are necessary, considering we already have comprehensive docs and tutorials.

src/py21cmfast/wrapper/inputs.py
jordanflitter: Thanks! I like this suggestion! Done!

src/py21cmfast/wrapper/photoncons.py
jordanflitter: Done!

tests/test_input_structs.py
jordanflitter: Done! I also added `deprecation` in `install_requires` in `setup.py`.

jordanflitter:
Thanks @steven-murray for your great comments! I fixed the branch accordingly!

steven-murray (approved):
Nice, thanks!

PR #689, opened 2026-05-12
Haloprop Correlation

daviesje:
_Note: I've written this description in a hurry, I intend to add to it after digging through my notes more, I'm also happy to answer questions_
This PR serves as documentation and status for where I left off with the source model upgrade project. This is still unfinished in a few aspects but hopefully it's mostly just bugfixes, performance and cleanup.

# The Idea: Self-Consistent Star-formation histories.

Determining SFR from stellar mass results in an inconsistency where integrating the SFR over time does not necessarily result in the stellar mass of the halo. this is exacerbated by the random sampling introduced in v4.

So instead of sampling the stellar mass, sfr and x-ray luminosity semi-independently, we sample the following three quantities:
- The SFR in the last 10 Myr
- The SFR in the last 100 Myr
- The SFR in the last snapshot

All three of these quantities are sampled from a single spectral-density, (Carvajal-Bohorquez et al. 2025). Covariance matrices are computed analytically at the start of each snapshot and then sampled for each halo (see https://arxiv.org/html/2507.14746v1#S2). We get a 6x6 full covariance matrix since we condition each of the three quantities on their descendant.

After sampling the quantities, the 10Myr and 100Myr SFR can be used to compute emissivities correlated with shorter and longer timescales. At the moment 10Myr is used for X-ray, and 100Myr is used for Gamma 12. The snapshot-SFRs are added together, resulting in the stellar mass.

# Status
- All the matrix math in `correlated_sfh.c` is complete and tested:
- Below is an exaggerated example (high sigma setting) of a halo SFH history, imagine an RNG value of +1 as 1 sigma above mean etc:
image
- Results are reasonable, where 10Myr and snapshot (~30Myr max) are very similar, 100Myr is much more slowly varying, and closer to the mean.
- The implementation into the rest of the code (HaloBox and afterwards) is unfinished. While all the components are there, it has not been run through an entire simulation, and likely contains many bugs
- There is test code still present, much of the old source model (including parameters) has not been deleted
- Since I needed access to multiple redshifts in the C code, I've created a structure similar to the parameters which holds the node redshifts in the backend. We don't need to pass the redshift directly to the backend anymore

# TODOS
- Get the model running through HaloBox, use the PerturbHalofield to check halo emissivity histories.
- Check the inputs to Spintemp and reionisation algorithms, these should not have changed much.
- rewrite all the halo sampler and property tests to accomodate the new model
- clean up all the old code and parameters
- We need to consider performance more thoroughly, speed seems similar at the moment, we need to hang on to the previous halo fields a little longer though. There is one large allocation in HaloBox.c summing progenitor properties which I was meaning to find a way around.

# Known Issues
- I believe there is a bug in the PSD timescale `SFH_TAU`, When I change it the SFHs don't respond the way I expect
- The first (dummy) halofield does not have `sfh_computed` set properly, causing a crash at the first call of `ComputeHaloBox`

## Summary by Sourcery

Introduce a self-consistent, temporally correlated star-formation history model and propagate it through halo sampling, halo gridding, and radiative transfer pipelines, using globally broadcast snapshot metadata instead of per-call redshift plumbing.

New Features:
- Add a correlated SFH model based on a PSD parametrization, including covariance construction, sampling utilities, and backend test hooks.
- Extend astro- and input-parameter handling to support SFH PSD parameters and node redshift metadata, exposing a Python test helper for SFH correlations.

Bug Fixes:
- Ensure halo-box and global-mean calculations use consistent Hubble time and star-formation timescales derived from cosmology and astro-parameters.
- Fix several redshift-handling inconsistencies by deriving current, previous, and descendant redshifts from a shared snapshot index instead of function arguments.

Enhancements:
- Refactor halo property definitions to work with multi-timescale SFR (10 Myr, 100 Myr, snapshot-integrated) and track progenitor contributions for self-consistent stellar mass growth.
- Replace legacy stochastic RNG-based halo property assignment with SFH-based correlated sampling throughout the halo catalog, progenitor sampler, and mapping to grids.
- Unify and simplify C API signatures for core compute kernels (perturbed fields, halo catalogs, spin temperature, ionization, brightness temperature, halo boxes) to use internal redshift state instead of explicit arguments.
- Improve logging and debug output for halo and box averages, including new SFH-related quantities.
- Extend struct and FFI layers (C and Python) to support new halo catalog fields, 64-bit indices, and SFH bookkeeping, while cleaning up obsolete scatter/median-handling paths.

Documentation:
- Update Python-level driver and wrapper expectations around required halo catalog inputs, especially previous catalogs for discrete halo models and SFH-enabled runs.

Tests:
- Add backend SFH correlation tests that validate PSD-based correlation tables and covariance construction against analytic expectations.

Chores:
- Remove unused or redundant scaling-relation scatter plumbing and related helper functions now superseded by the SFH model.


steven-murray:
@daviesje thanks for this! 

One question -- you mention that the reionization module should stay pretty similar. However, as you may see in #516, there is currently a problem with how we calculate reionization, due to the way cumulative recombinations are handled on short time-steps. We thought that perhaps your new source model might allow computing this based on the differential ionisation rate, dnion/dt rather than the cumulative nion. Do you have any thoughts on this?

daviesje:
> @daviesje thanks for this!
> 
> One question -- you mention that the reionization module should stay pretty similar. However, as you may see in #516, there is currently a problem with how we calculate reionization, due to the way cumulative recombinations are handled on short time-steps. We thought that perhaps your new source model might allow computing this based on the differential ionisation rate, dnion/dt rather than the cumulative nion. Do you have any thoughts on this?

I don't think this model specifically "allows" that change (none of the techincal details really overlap), but having more consistent SF histories might avoid some pitfalls that could occur imnplementing it

jordanflitter:
Thanks @daviesje, this sounds really cool!

Could you share with us your notes (with equations)? Having them would make it easier to understand better the idea of this PR and review it. Plus, they could be the basis of any future paper that will come out of this work.

daviesje:
> Thanks @daviesje, this sounds really cool!
> 
> Could you share with us your notes (with equations)? Having them would make it easier to understand better the idea of this PR and review it. Plus, they could be the basis of any future paper that will come out of this work.

https://www.wolframcloud.com/obj/2d233119-1277-40e1-a3b2-a8e561a58851

steven-murray:
Hi @daviesje -- thanks for this work again. I just wanted to make one general comment: you've added a new global node redshift array, but we're actively trying to diminish the number of globals we are dealing with. Passing in the full array of node redshifts rather than just a single redshift is great (and we were planning on doing this already!), but can we make it so that we pass these in as a parameter, rather than a global?

daviesje:
@steven-murray of course! Effectively I've added another parameter structure, minus some of the class logic. I guess when you're looking to reduce globals it should follow the whatever pattern you implement for the rest of the `InputParameters` classes.

PR #699, opened 2026-05-22
Halo overflow

jordanflitter:
# Halo Buffer Overflow Handling & Array Trimming

## Overview
This PR improves the handling of halo catalog buffer overflows and adds array trimming to eliminate wasted memory. The changes enhance error messaging to guide users toward solutions, fix M_MIN calculation issues, and ensure thread-safe buffer management.

### 🔴 Error Detection & Messaging (HaloCatalog.c & Stochasticity.c)

- **Restored buffer overflow detection** in `ComputeHaloCatalog()` that checks if total halos exceed allocated buffer
- **Enhanced error messages** that calculate and suggest the exact `HALO_CATALOG_MEM_FACTOR` value needed to resolve the overflow
- **Thread-safe overflow handling** in `sample_halo_grids()` and `sample_halo_progenitors()` using:
  - `any_thread_overflowed` flag with `#pragma omp critical` for safe thread coordination
  - Proper array assignment guarding with `if (!out_of_buffer)` checks
- **Restored `expected_nhalo()` logic** that correctly differentiates between `SOURCE_MODEL == 3` (DEXM-ESF) and `SOURCE_MODEL == 4` (CHMF-SAMPLER)

### 🔧 M_MIN Calculation Fix (HaloCatalog.c)

- **Removed `fmax()` wrapper** from M_MIN assignment
- **Why**: The previous code used `fmax(M_MIN, calculated_value)` with uninitialized `M_MIN` value. The uninitialized `M_MIN` value contains garbage, and finding the maximum between it and the calculated value could lead to unexpected results
- **Fix**: Revert to direct assignment without the call to `fmax()`

### 📦 Array Trimming (outputs.py)

- **Added `trim_to_n_halos()` method** to `HaloCatalog` class that:
  - Creates a new halo catalog with `buffer_size = n_halos` (actual count)
  - Dynamically copies only the used portion of each array (handles both 1D and 2D arrays)
  - Purges the original oversized catalog to free memory
  - Can reduce storage footprint by ~25-30% (in typical cases)

### 🔗 Integration (single_field.py)

- `determine_halo_catalog()` now calls `trim_to_n_halos()` before returning
- Ensures output catalogs have no wasted buffer space

### ✅ Validation (cfuncs.py)

- Restored validation in `get_expected_nhalo()` that prevents calls with non-discrete halo models
- Provides clear error message directing users to use `SOURCE_MODEL='DEXM-ESF'` or `'CHMF-SAMPLER'`

### 🧪 Comprehensive Testing

- **Added `test_halo_buffer_overflow_error_message()`** that:
  - Triggers a controlled buffer overflow with `HALO_CATALOG_MEM_FACTOR=1.0`
  - Verifies error message format and content
  - Validates per-thread halo counts are reported
  - Confirms suggested factor calculation matches error output
  
- **Added `test_using_cached_halo_catalog()` assertions** that verify:
  - `buffer_size == n_halos` (no wasted slots)
  - All array shapes match the actual halo count
  - Handles both 1D arrays (masses, RNG) and 2D arrays (coordinates)

- **Restored `test_bad_input_for_expected_nhalo()`** to catch invalid SOURCE_MODEL usage

## Testing Status
✅ All existing tests pass  
✅ New comprehensive buffer overflow test passes  
✅ New trimming assertions pass  
✅ No memory leaks or thread safety issues

## Impact
- **User experience**: Clear, actionable error messages when buffer overflows occur
- **Performance**: ~25-30% reduction in cached halo catalog size
- **Code quality**: Better thread safety, proper M_MIN handling, comprehensive validation, cleaner error handling

## Summary by Sourcery

Improve halo catalog buffer overflow handling, memory usage, and validation around discrete halo models.

Bug Fixes:
- Correct M_MIN initialization in halo catalog computation to avoid using uninitialized values.
- Restore correct expected_nhalo behavior and validation to only allow discrete halo models.
- Fix and harden halo buffer overflow detection in halo catalog generation, including thread-safe handling in halo sampling routines.

Enhancements:
- Add informative error messages on halo buffer overflow with per-thread diagnostics and a calculated suggestion for HALO_CATALOG_MEM_FACTOR.
- Trim halo catalogs to the actual number of halos returned from C, eliminating unused buffer space in cached outputs.
- Ensure single-field halo catalog generation returns a trimmed catalog and update tests to reflect discrete-halo SOURCE_MODEL usage.

Tests:
- Add regression tests covering halo buffer overflow error messaging, halo catalog caching behavior, and invalid usage of expected_nhalo.


src/py21cmfast/wrapper/outputs.py
```diff
@@ -824,6 +824,36 @@ def new(
...
+            shape = (n_halos,) if name != "halo_coords" else (n_halos, 3)
+            val = (
+                array.value[:n_halos]
+                if name != "halo_coords"
+                else array.value[:n_halos, :]
+            )
+            setattr(
+                halo_catalog_trimmed,
+                name,
+                Array(shape=shape, dtype=np.float32).initialize().with_value(val=val),
```
steven-murray: The downside here is that we'll temporarily have two copies of the array, so peak memory might actually go up, instead of down. 

I wonder if there's a way to do this without creating a temporary copy that you transfer the data into?

steven-murray (commented):
Thanks @jordanflitter, this is looking very nice indeed. I have only one comment, though sourcery seems to have a few more that you should consider :-)

src/py21cmfast/wrapper/outputs.py
jordanflitter: Oh yeah, that's a great comment! (better than the comments of sourcery imo ;)

There is no way to simply trim a computed array from C (since its values are frozen, also its shape). Because of this limitation, I think we ought to copy the array's content, trim it, and then remove the original array from the original catalog. This would be smarter than what I currently do.
Alternatively (and this is something that you probably won't like), we could remove the "frozen" limitation just for arrays from `HaloCatalog`.
If you have any other idea I'd love to hear it! (I'll ask in the meantime my buddy Claude for other ideas)

jordanflitter:
Thanks @steven-murray, I added a `MutableArray` class for the arrays in `HaloCatalog`. The arrays are then become frozen right before `trim_to_n_halos` returns. I think that implementation can't increase the peak memory because there is no allocation of new arrays. I also addressed sourcery's comments.

src/py21cmfast/wrapper/arrays.py
```diff
@@ -208,3 +208,19 @@ def loaded_from_disk(self, backend: CacheBackend | None = None) -> Self:
...
+    def trim_and_freeze(self, trimmed_shape, trimmed_value) -> Array:
+        """Convert to immutable Array after trimming."""
+        return Array(
+            shape=trimmed_shape,
+            value=trimmed_value,
+            dtype=self.dtype,
+            state=self.state,
+            initfunc=self.initfunc,
+            cache_backend=self.cache_backend,
+        )
```
steven-murray: This can actually be much simpler, now that I think about it. We don't need a MutableArray at all, just a `trimmed()` method on `Array`:

```python
def trimmed(self, trimmed_shape: tuple[int]) -> Self:
    slc = tuple(slice(0, n) for n in trimmed_shape)
    view = self.value[slc]
    return attrs.evolve(self, shape=trimmed_shape, value=view)
```

src/py21cmfast/wrapper/outputs.py
```diff
@@ -814,45 +814,35 @@ def new(
...
                 else array.value[:n_halos, :]
             )
-            setattr(
-                halo_catalog_trimmed,
-                name,
-                Array(shape=shape, dtype=np.float32).initialize().with_value(val=val),
+            frozen_array = array.trim_and_freeze(
+                trimmed_shape=trimmed_shape, trimmed_value=trimmed_value
             )
+            setattr(self, name, frozen_array)
```
steven-murray: Then if you take my first suggestion, this would just be

```python
new_shape = (n_halos,) + array.shape[1:]
new_array = array.trimmed(new_shape)
setattr(self, name, new_array)
```

src/py21cmfast/wrapper/outputs.py
```diff
@@ -814,45 +814,35 @@ def new(
...
 
-        # Purge the original halo catalog because it was exposed to C
-        self.purge(force=True)
-
-        return halo_catalog_trimmed
+        return self
```
steven-murray: I wouldn't return it, since it's an in-place operation,

steven-murray (commented):
THanks @jordanflitter  -- I think I found a simplification :-)

jordanflitter:
Yes, great suggestion @steven-murray! I implemented it in the latest commit!

steven-murray (approved):
Awesome, thanks!

Issue #701, opened 2026-05-25
[BUG] Inconsistent results with halo sampling when changing HII_DIM


ececcotti:
**Describe the bug:**
With the stochastic halo sampler, the reionization state of the box changes drastically, flipping between fully ionized and fully neutral, when the box size (`HII_DIM`) or the low-res cell size (`LOWRES_CELL_SIZE_MPC`) is changed.

**To Reproduce:**
This is the code I used. Between runs, I varied only `HII_DIM` and `LOWRES_CELL_SIZE_MPC`:
```python
import py21cmfast as p21c
p21c.config['EXTRA_HALOBOX_FIELDS'] = True
p21c.config['HALO_CATALOG_MEM_FACTOR'] = 2
inputs = p21c.InputParameters.from_template(
    'latest-dhalos', 
    random_seed=1420, 
    ZPRIME_STEP_FACTOR=1.2
).evolve_input_structs(
    N_THREADS = 32, 
    SAMPLER_MIN_MASS = 1e8, 
    HII_DIM = 100, 
    HIRES_TO_LOWRES_FACTOR = 4, 
    LOWRES_CELL_SIZE_MPC = 4, 
    USE_TS_FLUCT = False, 
    INHOMO_RECO = False, 
    HALOMASS_CORRECTION = 1., 
    USE_EXP_FILTER = False, 
    USE_UPPER_STELLAR_TURNOVER = False, 
    CELL_RECOMB = False, 
    R_BUBBLE_MAX = 15
)
ics = p21c.compute_initial_conditions(inputs = inputs)
perturbed_field = p21c.perturb_field(redshift = 6.95, initial_conditions = ics)
halo_catalog = p21c.determine_halo_catalog(redshift = 6.95, inputs = inputs, initial_conditions = ics)
halo_box = p21c.compute_halo_grid(redshift = 6.95, initial_conditions = ics, halo_catalog = halo_catalog)
ionized_field = p21c.compute_ionization_field(initial_conditions = ics, perturbed_field = perturbed_field, halobox = halo_box)
xHI = ionized_field.get('neutral_fraction')[..., 0]
``` 
I plotted `xHI` for different box sizes / low-res resolutions:
- Case 1 (correct): `HII_DIM = 100`, `LOWRES_CELL_SIZE_MPC = 4`
    <img width="500" alt="Image" src="https://github.com/user-attachments/assets/fef51ba6-2fbf-42d5-b4ee-c023e70d8a77" />
[...]

PR #703, opened 2026-05-27
Fix large-grid density normalization overflow

guotsuan:
Hi,

We encountered anomalous results when running 21cmFAST with a large grid size, e.g. DIM = 1300. In particular, the density field was found to be -1 across all pixels. We attempted to identify and resolve this issue. Please see the attached figure: the left panel shows the result before applying the fix, while the right panel shows the result after applying the fix.

Weixin Image_20260527092014_29_772

Simulation codes:
```
N_THREADS = 24
HII_DIM = 650
BOX_LEN = 1000
DIM = 1300
KEEP_3D_VELOCITIES = True

inputs = p21c.InputParameters.from_template(
    ['latest', 'gpc'],
    random_seed=1234,
    node_redshifts=p21c.wrapper.inputs.get_logspaced_redshifts(
        min_redshift=13,
        z_step_factor=1.1,
        max_redshift=36
    ),
    INHOMO_RECO=True,
    USE_MINI_HALOS=False,
    USE_RELATIVE_VELOCITIES=True,
    MINIMIZE_MEMORY=False
).evolve_input_structs(
    BOX_LEN=BOX_LEN,
    HIRES_TO_LOWRES_FACTOR=2,
    HII_DIM=HII_DIM,
    N_THREADS=N_THREADS,
    KEEP_3D_VELOCITIES=KEEP_3D_VELOCITIES
)
```
  Description

  This PR fixes integer overflow issues that can occur when running large-grid simulations.

  The main fix updates density normalization to use the existing pixel-count constants directly, avoiding intermediate int multiplications that can overflow for large grids. It also
  updates several loop counters and counters from int to unsigned long long int where they iterate over HII_TOT_NUM_PIXELS, and fixes one calloc call that was incorrectly casting the
  allocation size through float.

  Changes

  - Avoid overflow in normalise_delta_grid by computing mass_factor from HII_TOT_NUM_PIXELS / TOT_NUM_PIXELS.
  - Prevent overflow in volume-ratio calculations by casting dimensions before multiplication.
  - Use unsigned long long int loop counters for large pixel-count loops.
  - Use an unsigned counter and matching %llu format for large bad_count values.
  - Remove unsafe float cast from a calloc(HII_TOT_NUM_PIXELS, sizeof(float)) allocation.
  
  Contributors: 

Yepeng Yan, Le Zhang, Quan Guo, Yun Yu, Xiaolong Yang, Yiming Wang, Peng Wang, Hongmin Cao

## Summary by Sourcery

[...]

PR #704, opened 2026-05-27
Fixed huge integers into `unsigned long long int`

jordanflitter:
This PR follows the work of #703.

I took the liberty to make another commit on top of their work. I used AI (mainly Claude) to audit all the C files in order to find other lines where potentially huge integers are defined in the code with `int`. According to a mapping I made, there are 10 types of integers that appear throughout the code. 
 
1.  Number of cells in each dimension (e.g. `HII_DIM`, `DIM`, `D_PARA`, etc).
2. Any dummy index that runs along a single dimension of the box (e.g. `i`, `j`, `k`, or `x`, `y`, `z`).
3. A multiplication of the integers above with themselves, to denote the total number of cells in the box (e.g. `TOT_NUM_PIXELS`, `HII_TOT_NUM_PIXELS`, etc).
4. Any dummy index that runs through all the cells in the box.
5. Size of halo catalog. This corresponds to either the buffer size of the halo catalog, or the actual number of halos that were found in the simulation.
6. Any dummy index that runs through the halo catalog. Again, this could mean running along the buffer size, or along the actual number of halos that were found in the simulation.
7. Random seed.
8. Integers that are set by the user, e.g. `N_THREADS`, `N_STEP_TS`, `N_COND_INTERP` and `N_PROB_INTERP`, as well as the corresponding running dummy indices in the code.
9. There are also fixed sizes of interpolation tables used in the code, and running indices along their grid.
10. Finally, there are also enums in the code, like `HMF`, `POWER_SPECTRUM`, `PERTURB_ALGORITHM`, and more.

Only integers of type 3-7 have the potential to be very large, above the upper limit of 2.1e9 that is supported by 32-bit `int`. (Note: apparently, we also allow users to run the code with very large random seeds!)

I asked the AI to scan other places where integers of type 3-7 are not defined as `unsigned long long int`. Apart of the work that was done in #703, the AI managed to find 4 more places where `int` was wrongly used (most importantly, in `map_mass.c` and `HaloCatalog.c`). I also made some stylistic changes, e.g. `unsigned long long -> unsigned long long int`.

## Summary by Sourcery

Update handling of potentially very large loop indices, grid indices, halo counts, and random seeds to use 64‑bit unsigned integers where needed to avoid overflow.

Bug Fixes:
- Use unsigned long long int for grid, velocity, and halo catalog indices that can exceed 32‑bit int limits, including in map_mass, halo catalog construction, perturbed field normalization, and filtering.
- Change counters and halo totals that may grow beyond 2.1e9 to unsigned long long int and fix corresponding logging format specifiers.
- Ensure allocation sizes and loop variables that depend on total pixel counts or halo numbers use 64‑bit unsigned integers to prevent overflow or truncation.

Enhancements:
- Align type usage and style across the codebase by consistently using unsigned long long int for large indices and random seeds and simplifying volume and mass factor computations.

### DIFF
--- a/src/py21cmfast/src/BrightnessTemperatureBox.c
+++ b/src/py21cmfast/src/BrightnessTemperatureBox.c
@@ -60,7 +60,7 @@ int ComputeBrightnessTemp(float redshift, TsBox *spin_temp, IonizedBox *ionized_
                                 T_rad) private(i, j, k, pixel_deltax, pixel_x_HI)               \
     num_threads(simulation_options_global -> N_THREADS)
         {
-            index_t index;
+            index_huge index;
 #pragma omp for reduction(+ : ave)
             for (i = 0; i < simulation_options_global->HII_DIM; i++) {
                 for (j = 0; j < simulation_options_global->HII_DIM; j++) {
--- a/src/py21cmfast/src/BrightnessTemperatureBox.c
+++ b/src/py21cmfast/src/BrightnessTemperatureBox.c
@@ -60,7 +60,7 @@ int ComputeBrightnessTemp(float redshift, TsBox *spin_temp, IonizedBox *ionized_
                                 T_rad) private(i, j, k, pixel_deltax, pixel_x_HI)               \
     num_threads(simulation_options_global -> N_THREADS)
         {
-            unsigned long long int index;
+            index_t index;
 #pragma omp for reduction(+ : ave)
             for (i = 0; i < simulation_options_global->HII_DIM; i++) {
                 for (j = 0; j < simulation_options_global->HII_DIM; j++) {
--- a/src/py21cmfast/src/HaloBox.c
+++ b/src/py21cmfast/src/HaloBox.c
@@ -167,7 +167,7 @@ HaloProperties get_halobox_averages(HaloBox *grids) {
 
 #pragma omp parallel for reduction(+ : mean_count, mean_mass, mean_stars, mean_stars_mini, \
                                        mean_sfr, mean_sfr_mini, mean_n_ion, mean_xray, mean_wsfr)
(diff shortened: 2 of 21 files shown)


steven-murray (commented):
Thanks @jordanflitter this all looks good to me. I do think that sourcery's suggestions are a good idea here though -- so look to those first! We also need to ensure that the authors of the original PR get credit here!

Have you confirmed that these udpates fix the issue in #701 ?

jordanflitter:
I have now addressed sourcery's comments. We now have `index_t` that replaces `unsigned long long int` throughout the code. I haven't checked yet if this PR also fixes #701.

guotsuan:
Glad to see this bug is being fixed. If needed, we are happy to help the test too.

Cheers,

steven-murray (commented):
Hi @jordanflitter sorry for the delay on the review!

I think that the current implementation could be improved actually. You have created a new `index_t` type which is `unsigned long long int` and then replaced _every_ instance of `unsigned long long int` throughout the code with this type. However, the type is meant to be _semantic_ -- it should represent _indices_ not just any unsigned long long. 

You could either roll back the usage to only overwrite actual indices, or you could create some new types: `size_t` to indicate box sizes, `seed_t` for random seeds, etc. 

The current code will _work_ but I think is misleading.

jordanflitter:
Thanks @steven-murray for your comment! After the last commit, we now have `index_huge`, `size_huge` and `random_huge`. By scanning the code to make these changes, I think that we now use these macros ONLY when they are actually needed.
In addition, I made some stylistic changes in `indexing.h`, I removed `0llu`, `1llu` and `2llu` in favor of casting indices into `index_huge`. I think it looks cleaner that way. While I have confirmed with Claude that I didn't introduce any bug with these stylistic changes, it would be great if you could triple-check that.

steven-murray (approved):

PR #708, opened 2026-06-01
Multiple scattering small updates after publication

jordanflitter:
Small adjustments, after multiple scattering paper was published :)

## Summary by Sourcery

Update multiple-scattering references from the arXiv preprint to the published Physical Review D article across code, utilities, and tests.

Enhancements:
- Refresh inline comments in the filtering implementation to point to the correct equation labels in the published paper rather than the preprint.

Tests:
- Adjust reference-printing tests to expect the DOI and citation details of the published Lyman-alpha multiple-scattering paper.

### DIFF
--- a/src/py21cmfast/src/filtering.c
+++ b/src/py21cmfast/src/filtering.c
@@ -190,7 +190,7 @@ double asymptotic_2F3(double kR, double alpha, double beta) {
     // as given in
     // https://functions.wolfram.com/HypergeometricFunctions/Hypergeometric2F3/06/02/03/01/0003/. In
     // our case, z=-kR^2/4 and the parameters of the hypergeometric function are given below
-    // (see also Eq. D8 in arxiv: 2601.14360)
+    // (see also Eq. E8 in arxiv: 2601.14360)
     double a1 = (2. + alpha) / 2.;
     double a2 = (3. + alpha) / 2.;
     double b1 = 5. / 2.;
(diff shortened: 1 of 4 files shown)


src/py21cmfast/src/filtering.c
```diff
@@ -190,7 +190,7 @@ double asymptotic_2F3(double kR, double alpha, double beta) {
     // as given in
     // https://functions.wolfram.com/HypergeometricFunctions/Hypergeometric2F3/06/02/03/01/0003/. In
     // our case, z=-kR^2/4 and the parameters of the hypergeometric function are given below
-    // (see also Eq. D8 in arxiv: 2601.14360)
+    // (see also Eq. E8 in arxiv: 2601.14360)
```
steven-murray, line 193: This is not correct! Should either be updated to the paper reference or keep the D8 reference, or update the arxiv pdf ;-)

src/py21cmfast/src/filtering.c
```diff
@@ -256,7 +256,7 @@ double hyper_2F3(double kR, double alpha, double beta) {
         return 3.0 / (pow(kR, 3)) * (sin(kR) - cos(kR) * kR);
     }
     // For a small argument, we compute the hypergeometric function through power-law expansion
-    // This is Eq. (D7) in arxiv: 2601.14360
+    // This is Eq. (E7) in arxiv: 2601.14360
```
steven-murray, line 259: This too

src/py21cmfast/src/filtering.c
jordanflitter: If you checked the paper on arxiv now, then you are indeed correct. But wait for tomorrow :)

steven-murray: Aha!

steven-murray (approved):

PR #717, opened 2026-06-03
added an update method to CacheConfig

### LINKED ISSUE #673: Make a convenient method to `evolve` an existing `CacheConfig` instance
It would be nice to have a convenient `evolve` method for `CacheConfig`, something like `p21c.CacheConfig.on(xray_source_box=False)`, or `p21c.CacheConfig.on().update(xray_source_box=False)`.

jordanflitter:
Fix #673

## Summary by Sourcery

Add an update method to CacheConfig and allow predefined cache configurations to be customized via keyword overrides.

New Features:
- Introduce a CacheConfig.update method to create modified cache configurations based on existing ones.

Enhancements:
- Allow CacheConfig.on, .off, .noloop, and .last_step_only to accept keyword overrides when constructing cache configurations.

Tests:
- Add unit tests for CacheConfig factory methods and the new update behavior to ensure configurations are correctly overridden.

### DIFF
--- a/src/py21cmfast/io/caching.py
+++ b/src/py21cmfast/io/caching.py
@@ -9,7 +9,7 @@ import logging
 import re
 from hashlib import md5
 from pathlib import Path
-from typing import ClassVar, Self
+from typing import ClassVar, Self, TypedDict, Unpack
 
 import attrs
 import numpy as np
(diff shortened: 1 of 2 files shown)


steven-murray (approved):
I agree with sourcery's suggestions, but otherwise definitely approve.

PR #718, opened 2026-06-05
Recombination model enum in C

jordanflitter:
In #681 we added a homogeneous recombination model, so we switched from the boolean flag `INHOMO_RECO` (which will become deprecated) to an enum-type variable called `RECOMB_MODEL`.

Meanwhile, in #698 we have introduced enums in the C code. In addition, in #624 we incorporated the `init_once` logic in the main branch (v4.1) which depends partially on the archaic flag `INHOMO_RECO`.

This PR does the following:
1. It replaces the remaining appearances of `INHOMO_RECO` in favor of `RECOMB_MODEL`.
2. It uses enums in the C code for `RECOMB_MODEL`.

## Summary by Sourcery

Switch recombination logic in the C and Python code to use the new RECOMB_MODEL enum and helper, removing remaining dependencies on the deprecated INHOMO_RECO flag.

Enhancements:
- Use RECOMB_MODEL enum constants and a uses_recombination helper throughout C recombination, ionization, halo, and mapping routines.
- Align global initialization and reference-printing utilities with the new RECOMB_MODEL API, including updated tests and comments.

### DIFF
--- a/src/py21cmfast/drivers/_global_initialization.py
+++ b/src/py21cmfast/drivers/_global_initialization.py
@@ -151,7 +151,7 @@ class GlobalInitializationManager:
         # NOTE: run_global_evolution at the moment does not do recombination calculations, so we only initialize if HII_DIM > 1.
         # If this is changed in the future, it will be necessary to remove the HII_DIM > 1 condition below, since it would cause a segfault!
         if (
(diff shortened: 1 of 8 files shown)


steven-murray (approved):
Looks great to me! Not sure why Python 3.14 is failing :/

Issue #688, opened 2026-05-06
[BUG] Munoz21 Template SFRD Doesn't Match Munoz+21 Paper


steven-murray:
**Describe the bug:**

[...]

PR #721, opened 2026-06-08
Turnover mass are computed or extracted automatically in `cfuncs.py`

### LINKED ISSUE #115: ability to pass mturn to compute_luminosity_function

Previously without minihalos, the turnover mass is a free parameter. Therefore, we can calculate LFs without actually running 21cmFAST -- we can determine LF with the astro parameters. However, with minihalos, the turnover mass is determined by feedback and we need to run 21cmFAST to get that. I’ve modified the function `compute_luminosity_function` to take turnover masses as an argument. Users need to pass the turnover masses to this function if USE_MINI_HALOS. (this will not affect if not USE_MINI_HALOS)

Right now, IonizedBox has two properties (i.e. `log10_Mturnover_ave` and `log10_Mturnover_MINI_ave`) which give us the average turnover masses for ACGs and MCGs in logarithm. I need to find a way to pass them to LC so that users can access them.

Also, it's better to provide a function to do interpolation of `log10_Mturnover_ave` and `log10_Mturnover_MINI_ave`, so the user can get their values at the LF redshifts. something like:
`lightcone.log10_Mturnover_ave(redshift)` and `lightcone.log10_Mturnover_MINI_ave(redshift)`

Potentially, we can also pass the LC to `compute_luminosity_function`. 

**Additional context**

This is particularly important if people are interested in using LFs, e.g. 21CMMC

jordanflitter:
Fixes #115 (and also somewhat of #688).

 This PR contains the following:

1. New python and C functions with the same name were made, `compute_mturns`.
2. Using `compute_mturns`, the fields `log10_Mturnover_ave` and `log10_Mturnover_MINI_ave` from `IonisationBox` are computed as part of `run_global_evolution`.
3. In `compute_luminosity_function`, a `TypeError` with a message is raised if either `mturnovers` or `mturnovers_mini` is not `None`.
4. If both `mturnovers` and `mturnovers_mini` are `None`, the code checks if the possible argument `lightcone` is provided. If it is provided, the global turnover masses are extracted from the lightcone, otherwise the code performs a quick `run_global_evolution` simulation and extracts the global turnover masses from its output.
5. A similar logic has been implemented also for `evaluate_SFRD_z`, `evaluate_Nion_z`, `evaluate_SFRD_cond`, `evaluate_Nion_cond`, and `evaluate_Xray_cond`. Moreover, these functions now return a physically meaningful output (before this change, these functions basically returned weighted integrals over the HMF, without the multiplicative factors that make them interesting physical quantities). The doctorings of all these functions have been improved a lot.
6. In addition, since the global turnover masses are extracted/computed automatically, they mostly depend on their `redshifts`, not `densities`. This has been reflected in both C and python.
7. Three new tests were added in `test_cfuncs.py`: these are `test_functions_with_and_without_lightcone`, `test_removed_log10mturns_argument`, and `test_removed_arguments_are_cleaned_up_in_v5`.

## Summary by Sourcery

Automate computation and propagation of global turnover masses for ACGs and MCGs, and update related high-level integrals to use physically meaningful, self-consistent quantities derived from global evolution or lightcone data.

New Features:
- Add Python and C helper functions to compute halo turnover masses for atomic and molecular cooling galaxies from thermochemical quantities.
- Automatically compute and store global turnover masses in global evolution and lightcone outputs, and expose them through new IonisationBox and global_evolution quantities.
- Allow luminosity function and global/conditional SFRD, Nion, and X-ray evaluators to infer turnover masses from either a provided lightcone or an internal global evolution run instead of requiring them as inputs.

Bug Fixes:
- Ensure SFRD, Nion, and X-ray evaluators now include the correct physical prefactors so their outputs are in meaningful physical units.
- Fix mini-halo handling in conditional and global emissivity/SFRD calculations so that mini-halo contributions are consistently gated by USE_MINI_HALOS and source model choices.

Enhancements:
- Simplify APIs for turnover-mass-dependent routines by deprecating explicit mturn/log10mturn arguments in favor of automatic estimation, with clear TypeError messages for legacy usage.
- Refine integral wrappers to work with scalar global turnover masses (especially for mini-halos) and apply consistent efficiency and normalization factors in C.
- Extend global evolution to track and interpolate global turnover masses across redshift and feed them into downstream calculations.
- Improve docstrings, tests, and interpolation-table validations around global and conditional SFRD, Nion, and X-ray quantities, including support for running with and without lightcones and mini-halos.

Tests:
- Add end-to-end tests for turnover-mass-dependent functions with and without a lightcone, and with/without mini-halos enabled.
- Add regression tests ensuring removed mturn/log10mturn arguments now raise informative TypeErrors and are cleaned up in v5.
- Update interpolation-table tests for SFRD, Nion, and conditional integrals to the new turnover-mass API and physical normalization assumptions.

### DIFF
--- a/src/py21cmfast/src/integral_wrappers.c
+++ b/src/py21cmfast/src/integral_wrappers.c
@@ -323,10 +323,13 @@ void get_conditional_Xray(double redshift, double R, int n_densities, double *de
         if (dens > max_dens) max_dens = dens;
     }
 
+    double X_RAY_FACTOR = RHOcrit * cosmo_params_global->OMm * 1e38;
+
     if (uses_hmf_interpolation(matter_options_global->USE_INTERPOLATION_TABLES)) {
         initialise_Xray_Conditional_table(redshift, min_dens, max_dens, M_min, M_cond, M_cond, &sc);
     }
     for (i = 0; i < n_densities; i++)
-        out_xray[i] = EvaluateXray_Conditional(densities[i], log10_mturn_mini, redshift, growthf,
-                                               M_min, M_cond, M_cond, sigma_cond, &sc);
+        out_xray[i] = X_RAY_FACTOR * EvaluateXray_Conditional(densities[i], log10_mturn_mini,
+                                                              redshift, growthf, M_min, M_cond,
+                                                              M_cond, sigma_cond, &sc);
 }
--- a/src/py21cmfast/src/integral_wrappers.c
+++ b/src/py21cmfast/src/integral_wrappers.c
@@ -212,7 +212,7 @@ void get_conditional_SFRD(double redshift, double R, int n_densities, double *de
 }
 
 void get_conditional_Nion(double redshift, double R, int n_densities, double *densities,
-                          double *log10_mturns_acg, double *log10_mturns_mcg, double *out_nion,
+                          double log10_mturn_acg, double log10_mturn_mcg, double *out_nion,
                           double *out_nion_mini) {
     double M_min = minimum_source_mass(redshift, true);
     double M_cond = RtoM(R);
@@ -237,8 +237,8 @@ void get_conditional_Nion(double redshift, double R, int n_densities, double *de
     double dens, l10mturn_a, l10mturn_m;
     for (i = 0; i < n_densities; i++) {
         dens = densities[i];
-        l10mturn_a = log10_mturns_acg[i];
-        l10mturn_m = log10_mturns_mcg[i];
+        l10mturn_a = log10_mturn_acg;
+        l10mturn_m = log10_mturn_mcg;
         if (dens < min_dens) min_dens = dens;
         if (dens > max_dens) max_dens = dens;
         if (l10mturn_a < min_l10mturn_acg) min_l10mturn_acg = l10mturn_a;
@@ -253,13 +253,13 @@ void get_conditional_Nion(double redshift, double R, int n_densities, double *de
                                            max_l10mturn_mcg, &sc, false);
     }
     for (i = 0; i < n_densities; i++)
-        out_nion[i] = EvaluateNion_Conditional(densities[i], log10_mturns_acg[i], growthf, M_min,
+        out_nion[i] = EvaluateNion_Conditional(densities[i], log10_mturn_acg, growthf, M_min,
                                                M_cond, M_cond, sigma_cond, &sc, false);
     if (astro_options_global->USE_MINI_HALOS) {
         for (i = 0; i < n_densities; i++)
             out_nion_mini[i] =
-                EvaluateNion_Conditional_MINI(densities[i], log10_mturns_mcg[i], growthf, M_min,
-                                              M_cond, M_cond, sigma_cond, &sc, false);
+                EvaluateNion_Conditional_MINI(densities[i], log10_mturn_mcg, growthf, M_min, M_cond,
+                                              M_cond, sigma_cond, &sc, false);
     }
 }
 
--- a/src/py21cmfast/src/integral_wrappers.c
+++ b/src/py21cmfast/src/integral_wrappers.c
@@ -230,22 +230,18 @@ void get_conditional_Nion(double redshift, double R, int n_densities, double *de
     int i;
     double min_dens = -1;
     double max_dens = 10;
-    double min_l10mturn_acg = 10.;
-    double min_l10mturn_mcg = 10.;
-    double max_l10mturn_acg = 5.;
-    double max_l10mturn_mcg = 5.;
     double dens, l10mturn_a, l10mturn_m;
     for (i = 0; i < n_densities; i++) {
         dens = densities[i];
-        l10mturn_a = log10_mturn_acg;
-        l10mturn_m = log10_mturn_mcg;
         if (dens < min_dens) min_dens = dens;
         if (dens > max_dens) max_dens = dens;
-        if (l10mturn_a < min_l10mturn_acg) min_l10mturn_acg = l10mturn_a;
-        if (l10mturn_a > max_l10mturn_acg) max_l10mturn_acg = l10mturn_a;
-        if (l10mturn_m < min_l10mturn_mcg) min_l10mturn_mcg = l10mturn_m;
-        if (l10mturn_m > max_l10mturn_mcg) max_l10mturn_mcg = l10mturn_m;
     }
+
+    double eps = 0.01;
+    double min_l10mturn_acg = log10_mturn_acg - eps;
+    double max_l10mturn_acg = log10_mturn_acg + eps;
+    double min_l10mturn_mcg = log10_mturn_mcg - eps;
+    double max_l10mturn_mcg = log10_mturn_mcg + eps;
 
     if (uses_hmf_interpolation(matter_options_global->USE_INTERPOLATION_TABLES)) {
         initialise_Nion_Conditional_spline(redshift, min_dens, max_dens, M_min, M_cond, M_cond,
--- a/src/py21cmfast/src/integral_wrappers.c
+++ b/src/py21cmfast/src/integral_wrappers.c
@@ -174,7 +174,7 @@ void get_conditional_FgtrM(double redshift, double R, int n_densities, double *d
 }
 
 void get_conditional_SFRD(double redshift, double R, int n_densities, double *densities,
-                          double *log10_mturns, double *out_sfrd, double *out_sfrd_mini) {
+                          double log10_mturns_mini, double *out_sfrd, double *out_sfrd_mini) {
     double M_min = minimum_source_mass(redshift, true);
     double M_cond = RtoM(R);
     double sigma_cond = EvaluateSigma(log(M_cond));
@@ -207,7 +207,7 @@ void get_conditional_SFRD(double redshift, double R, int n_densities, double *de
     if (astro_options_global->USE_MINI_HALOS) {
         for (i = 0; i < n_densities; i++)
             out_sfrd_mini[i] = EvaluateSFRD_Conditional_MINI(
-                densities[i], log10_mturns[i], growthf, M_min, M_cond, M_cond, sigma_cond, &sc);
+                densities[i], log10_mturns_mini, growthf, M_min, M_cond, M_cond, sigma_cond, &sc);
     }
 }
 
--- a/src/py21cmfast/src/integral_wrappers.c
+++ b/src/py21cmfast/src/integral_wrappers.c
@@ -104,6 +104,8 @@ void get_global_SFRD_z(int n_redshift, double *redshifts, double *log10_turnover
     int i;
     double z_min = simulation_options_global->Z_HEAT_MAX;
     double z_max = 0.;
+    double SFRD_FACTOR, SFRD_FACTOR_MINI;
+
     for (i = 0; i < n_redshift; i++) {
         if (redshifts[i] < z_min) z_min = redshifts[i];
         if (redshifts[i] > z_max) z_max = redshifts[i];
@@ -114,9 +116,15 @@ void get_global_SFRD_z(int n_redshift, double *redshifts, double *log10_turnover
     }
 
     for (i = 0; i < n_redshift; i++) {
-        out_sfrd[i] = EvaluateSFRD(redshifts[i], &sc);
-        if (astro_options_global->USE_MINI_HALOS)
-            out_sfrd_mini[i] = EvaluateSFRD_MINI(redshifts[i], log10_turnovers_mcg[i], &sc);
+        SFRD_FACTOR = astro_params_global->F_STAR10 * cosmo_params_global->OMb * RHOcrit /
+                      astro_params_global->t_STAR / t_hubble(redshifts[i]);
+        out_sfrd[i] = SFRD_FACTOR * EvaluateSFRD(redshifts[i], &sc);
+        if (astro_options_global->USE_MINI_HALOS) {
+            SFRD_FACTOR_MINI = astro_params_global->F_STAR7_MINI * cosmo_params_global->OMb *
+                               RHOcrit / astro_params_global->t_STAR / t_hubble(redshifts[i]);
+            out_sfrd_mini[i] =
+                SFRD_FACTOR_MINI * EvaluateSFRD_MINI(redshifts[i], log10_turnovers_mcg[i], &sc);
+        }
     }
 }
 
@@ -204,6 +212,11 @@ void get_conditional_SFRD(double redshift, double R, int n_densities, double *de
     double min_dens = -1;
     double max_dens = 10;
     double dens;
+    double SFRD_FACTOR = astro_params_global->F_STAR10 * cosmo_params_global->OMb * RHOcrit /
+                         astro_params_global->t_STAR / t_hubble(redshift);
+    double SFRD_FACTOR_MINI = astro_params_global->F_STAR7_MINI * cosmo_params_global->OMb *
+                              RHOcrit / astro_params_global->t_STAR / t_hubble(redshift);
+
     for (i = 0; i < n_densities; i++) {
         dens = densities[i];
         if (dens < min_dens) min_dens = dens;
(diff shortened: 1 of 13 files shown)


src/py21cmfast/src/thermochem.c
```diff
@@ -305,3 +305,13 @@ double reionization_feedback(float z, float Gamma_halo_HII, float z_IN) {
            pow((1. + z) / 10, REION_SM13_B) *
            pow(1 - pow((1. + z) / (1. + z_IN), REION_SM13_C), REION_SM13_D);
 }
+
+void compute_mturns(float z, float J_21_LW, float vcb, float Gamma12, float z_reion,
+                    double *M_turn_a, double *M_turn_m) {
+    double M_turn_r = reionization_feedback(z, Gamma12, z_reion);
+    *M_turn_a = atomic_cooling_threshold(z);
+    *M_turn_m = lyman_werner_threshold(z, J_21_LW, vcb);
+    *M_turn_a = fmax(*M_turn_a, fmax(M_turn_r, astro_params_global->M_TURN));
```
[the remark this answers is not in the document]
jordanflitter: Are you sure? Look at how `get_log10_turnovers` in `HaloBox.c` is written, these are essentially the same code lines.

src/py21cmfast/src/integral_wrappers.c
```diff
@@ -133,13 +141,25 @@ void get_global_Nion_z(int n_redshift, double *redshifts, double *log10_turnover
...
     }
 
+    double ION_EFF_FACTOR, ION_EFF_FACTOR_MINI;
+    if (source_model_is_mass_dependent(matter_options_global->SOURCE_MODEL)) {
+        ION_EFF_FACTOR = astro_params_global->F_STAR10 * astro_params_global->F_ESC10 *
+                         astro_params_global->POP2_ION;
+        ION_EFF_FACTOR_MINI = astro_params_global->F_STAR7_MINI * astro_params_global->F_ESC7_MINI *
+                              astro_params_global->POP3_ION;
+    } else {
+        // no mini-halos when SOURCE_MODE is mass independent (constant ionization efficiency)
```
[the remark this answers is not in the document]
jordanflitter: Mini-halos cannot be used together with a mass-independent source model, we have protections against that in `inputs.py`.

src/py21cmfast/wrapper/cfuncs.py
```diff
@@ -170,42 +240,42 @@ def compute_luminosity_function(
...
+        global_evolution = run_global_evolution(inputs=inputs)
+        mturnovers_global = pow(10.0, global_evolution.quantities["log10_mturn_acg"])
+        mturnovers_mini_global = pow(
+            10.0, global_evolution.quantities["log10_mturn_mcg"]
         )
-        component = "acg"
+
+    # Interpolate the mturnover arrays at the input redshifts
+    mturnovers = np.interp(redshifts, inputs.node_redshifts, mturnovers_global)
+    mturnovers_mini = np.interp(
```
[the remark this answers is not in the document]
jordanflitter: You are right! Thanks for catching this bug! I have fixed that (I reversed the order of the arrays used for the interpolation grid).

tests/test_cfuncs.py
```diff
@@ -528,3 +523,134 @@ def make_matterfield_comparison_plot(
...
+            log10mturns=np.linspace(8.0, 9.0, num=800),
+        )
+
+
+def test_removed_arguments_are_cleaned_up_in_v5():
+    """Reminder to remove the TypeError checks for log10mturns etc. in v5."""
+    version = tuple(int(x) for x in p21c.__version__.split(".")[:2])
+    if version >= (5, 0):
+        pytest.fail(
+            "Version is now >= 5.0 — please remove the deprecated `mturnovers`, "
```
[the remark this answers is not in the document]
jordanflitter: Thanks for the suggestion. I have included a new test, called `test_roundtrip_mturns`, where in this test the inputs are arrays. Every time the tests call `run_global_evolution` `compute_mturns` is tested with scalar inputs, so this branch is covered by other tests. I also test now the expected `ValueError` in `compute_mturns` (in `test_bad_integral_inputs`).

src/py21cmfast/src/integral_wrappers.c
```diff
@@ -174,7 +194,7 @@ void get_conditional_FgtrM(double redshift, double R, int n_densities, double *d
 }
 
 void get_conditional_SFRD(double redshift, double R, int n_densities, double *densities,
-                          double *log10_mturns, double *out_sfrd, double *out_sfrd_mini) {
```
steven-murray, line 177: Can you justify the choice to make `mturn` not depend on density here, given that this is for a conditional SFRD?

src/py21cmfast/wrapper/cfuncs.py
```diff
@@ -76,6 +78,75 @@ def get_halo_catalog_buffer_size(
...
+@init_c_state(broadcast_inputs=True)
+def compute_mturns(
+    *,
+    inputs: InputParameters,
+    redshifts: float | Sequence[float],
+    J_LW_21: float | Sequence[float],
+    v_cb: float | Sequence[float],
+    ionisation_rate_G12: float | Sequence[float],
+    z_reion: float | Sequence[float],
+):
```
steven-murray:
```suggestion
) -> tuple[float, float]:
```

src/py21cmfast/wrapper/cfuncs.py
```diff
@@ -170,42 +240,44 @@ def compute_luminosity_function(
...
-        if component in ["both", "mcg"]:
-            if mturnovers_mini is None:
-                raise ValueError(
-                    "calculating MCG LFs with mini-halo feature requires users to "
-                    "specify mturnovers_MINI!"
-                )
+    if (mturnovers is not None) or (mturnovers_mini is not None):
+        raise TypeError(
+            "`mturnovers` and `mturnovers_mini` have been removed. "
+            "Please use the `lightcone` argument instead, or don't use it to estimate the turnover masses."
```
steven-murray:
```suggestion
            "Please use the `lightcone` argument instead, or leave unspecified and they will be estimated automatically."
```

src/py21cmfast/wrapper/cfuncs.py
```diff
@@ -170,42 +240,44 @@ def compute_luminosity_function(
...
-        mturnovers = (
-            np.zeros(len(redshifts), dtype=np.float32) + 10**astro_params.M_TURN
+        from ..drivers.global_evolution import run_global_evolution
+
+        # If lightcone is not provided, we estimate the turnover masses from the global evolution
+        global_evolution = run_global_evolution(inputs=inputs)
+        mturnovers_global = pow(10.0, global_evolution.quantities["log10_mturn_acg"])
+        mturnovers_mini_global = pow(
+            10.0, global_evolution.quantities["log10_mturn_mcg"]
         )
```
steven-murray, line 278: I recommend extracting this logic out into a (private) helper function.

src/py21cmfast/wrapper/cfuncs.py
```diff
@@ -549,24 +621,61 @@ def evaluate_SFRD_z(
...
+    -------
+    sfrd : np.ndarray
+        The global star formation rate density at the given redshifts for ACGs.
+    sfrd_mini : np.ndarray
+        The global star formation rate density at the given redshifts for MCGs.
+    """
+    if log10mturns is not None:
+        raise TypeError(
+            "`log10mturns` has been removed. "
+            "Please use the `lightcone` argument instead, or don't use it to estimate the turnover masses."
```
steven-murray: same suggestion as above.

src/py21cmfast/wrapper/cfuncs.py
```diff
@@ -579,24 +688,61 @@ def evaluate_Nion_z(
...
+    -------
+    nion : np.ndarray
+        The global number of ionising photons per baryon at the given redshifts for ACGs.
+    nion_mini : np.ndarray
+        The global number of ionising photons per baryon at the given redshifts for MCGs.
+    """
+    if log10mturns is not None:
+        raise TypeError(
+            "`log10mturns` has been removed. "
+            "Please use the `lightcone` argument instead, or don't use it to estimate the turnover masses."
```
steven-murray: same suggestion again (OK, stopping commenting on this now)

tests/test_c_interpolation_tables.py
```diff
@@ -362,41 +356,37 @@ def test_FgtrM_conditional_tables(R, delta_range, plt):
...
 @pytest.mark.parametrize("name", options_hmf)
-def test_SFRD_z_tables(name, z_range, log10_mturn_range, plt):
+def test_SFRD_z_tables(name, z_range, plt):
     redshift, kwargs = OPTIONS_HMF[name]
     inputs = get_all_options_struct(
         redshift,
         USE_MINI_HALOS=True,
         RECOMB_MODEL="inhomogeneous",
         USE_TS_FLUCT=True,
+        ZPRIME_STEP_FACTOR=1.2,  # speed it up a bit because the functions below call run_global_evolution
```
steven-murray: Fine, but also if you add the ability to pass in global evolution directly that will also make this more efficient.

steven-murray (commented):
Thanks @jordanflitter, only a few more small comments in addition to what we talked about earlier.

src/py21cmfast/src/integral_wrappers.c
jordanflitter: That's a very good question. The turnover masses are determined directly from the physical fields `J_21_LW`, `vcb`, `Gamma12`, and `z_reion` (and also on redshift), as can be seen from the new C-function `compute_mturns`. The dependence of the turnover mass on the density field is thus implicit, as all the input physical fields depend on the current local density (but they also depend on past non-local values of the density field! For example the LW flux at a point depends on the history of the density/emissivity field around that point).

However, in this PR the turnover masses that are passed to the C code are the **global** turnover masses, as extracted from `lightcone` or `global_evolution`. So there is some inconsistency here since the user wants to compute the conditional SFRD (or `n_ion`, or X-ray emissivity) given a **conditional** value of the density field, but we ignore this conditional value when we pass the turnover masses to the C code. 

I am not sure there is a good solution around this. Previously, if users wanted to use these conditional functions, they had to either guess the conditional turnover masses, or taking them out of a lightcone simulation, as we do now. Since we only save the global turnover masses, we don't have lightcone boxes for them, and even if we did, it's pretty impossible to extract the "correct" turnover mass at the "correct" density value (because of the dependency on the non-local history of the density field).

Note that despite this limitation, there is still some density-conditionality in these functions, since the density value is used for evaluating the CHMF.

So it is true that we don't really compute the most-precise conditional SFRD/`n_ion` etc with these functions, but rather give an estimate. We could mention this briefly in the notes, but I wouldn't do much more than this. For example, I would not restore back the turnover masses as possible arguments. If users are absolutely sure of their conditional values for the turnover masses (which depend not only on the local environment where the SFRD is computed, but also on the history of how the density and other fields had been evolved), they are welcome to have it in their own scripts :)

src/py21cmfast/wrapper/cfuncs.py
jordanflitter: Done!

src/py21cmfast/wrapper/cfuncs.py
jordanflitter: Done!

src/py21cmfast/wrapper/cfuncs.py
jordanflitter: I also thought about doing something like this. While this is certainly doable, I fear that making a helper function for this purpose would make things worse because:
1. `compute_luminosity_function` uses linear masses, while the other functions use logarithmic masses.
2. The global functions can accept `redshifts` as an array, while the conditional ones only accept `redshift` as a float. So the dummy `log10mturns_mini` (when `USE_MINI_HALOS=False`) is either `np.zeros_like(redshifts)` or 0.
3. Remember that the C function `get_conditional_Nion` (called in the python function `evaluate_Nion_cond`) is the only function that requires having also the turnover mass for ACGs, so for example for that function there needs to be an extra output (`log10mturn_acg`), while `log10mturn_mcg` can never be a dummy.

These variations between the 6 relevant functions in `cfuncs.py` make it challenging to unify their logics into a single helper function. So while I believe it is possible, I'm not sure we would gain much from this effort (and even if we let copilot do this job, I fear that the resulting helper function would be messy in order to work right for all the different possibilities).

src/py21cmfast/wrapper/cfuncs.py
jordanflitter: Same reply as above :)
While the error logic in the 6 functions looks similar, the variations between them makes it annoying to put it in a unifying helper funciton.

src/py21cmfast/wrapper/cfuncs.py
jordanflitter: And same reply as above, again :)

tests/test_c_interpolation_tables.py
jordanflitter: You are right! But we actually need to have it there because we must have `inputs.node_redshifts == global_evolution.node_redshifts`, and if I try to move this line to be defined in `OPTIONS_HMF`, then `test_halo_sampler` segfaults for some reason :(

jordanflitter:
Thanks for your suggestions @steven-murray! The main changes since you last reviewed it are:
1. Functions return `output_mini = None` if `USE_MINI_HALOS=False`.
2. Functions can accept `global_evolution` as a possible input. This makes the tests MUCH faster now, since `run_global_evolution` (even though it's fast), doesn't need to be called dozens of times in `test_c_interpolation_tables.c`. Also, I figured that there is no need to have a `lightcone_mini` in `test_cfuncs.py`, so its removal from the module saves about 30 seconds :)

Like I wrote above, I don't think it's much beneficial to have a unifying helper. If you want to work on that or let copilot do it for us, I'm absolutely fine with this :)
You had a good point about the fact that we use the global turnover masses to compute conditional quantities. I don't think we could do things better than they currently are, besides of documenting this in the docstring (though I'm not sure how much I like this idea either :)

src/py21cmfast/src/integral_wrappers.c
steven-murray: OK, I think this is probably fair.

tests/test_c_interpolation_tables.py
steven-murray: errrr, well this sounds like a bug?

steven-murray:
Thanks @jordanflitter I think I'm basically good with this now. Only two things:

1. We should probably make an issue to identify the source of this segfault bug you mentioned.
2. I would write something about the fact that we're using global densities -- if not in the docstring to the python functions, at least as comments in the C functions that you modified.

tests/test_c_interpolation_tables.py
jordanflitter: I am not sure if that's a bug. All I can say is that because `test_halo_sampler.py` is connected to `test_c_interpolation_tables.py` through `OPTIONS_HMF`, if `ZPRIME_STEP_FACTOR=1.2`, then there is a segfault.

A step of 1.2 is quite big, I'm not sure if the halo sampler was meant to work with such large differences. But segfault is never a nice thing to have, so I'm okay with having a new issue on that :)

steven-murray: Yeah, in my opinion, a segfault is always a bug.

jordanflitter:
Alright, I will:
1. Make an issue for that potential segfault in `test_halo_sampler.py`.
2. Mention in the docstring that we use global turnover mass in the conditional functions.

We should also figure out what is causing the tests to segfault sometimes in `tests/test_global_initialization.py::test_initializations`. I don't think this has to do with this PR.

jordanflitter:
@steven-murray I addressed your two comments:
1. We have a new issue that describes the segfault I encountered: #723.
2. I added some notes in the last commit that clarify about the usage of global turnover masses in the conditional functions.

steven-murray (approved):

PR #725, opened 2026-06-12
Fix segfault in `test_global_initialization.py`

jordanflitter:
We have encountered some segfaults in `test_global_initialization.py`. These segfaults are not due to faulty design in `_global_initialization.py`, but rather because of how `test_initializations` was written. 
In short, when calling directly the initialization functions (not recommended!), a segfault could occur when two direct initialization calls with different `inputs` are executed consecutively. This happens because in using the direct initialization methods, `_GlobalInitManagerSingleton` does not know if the current `inputs` are the same ones that were used for previous initializations. 
In normal usage, users should not be concerned about initializations, since they are automatically performed via the `init_c_state` decorator, which calls `_GlobalInitManagerSingleton.init`, where `free()` is called if the current `inputs` don't match the saved `inputs`. 

While it is possible to keep `test_initializations` as it was written and avoid the segfaults, it would unnecessarily make the logic in `_global_initialization.py` more complicated, just to pass a test of an unrecommended usage of the code. An easier solution, that was made in this PR, was to call `free()` between the two direct initialization calls in `test_initializations`. This behavior mimics the logic in `_GlobalInitManagerSingleton.init`, and thus solves the segfault in `test_initializations`.

Apart from this, a few other minor changes were made:
1. Renamed `test_initializations` -> `test_direct_initializations`, to stress that direct initializations are tested in this test.
2. Because the segfault that we had was not deterministic (sometimes the test passed smoothly, sometimes it segfaulted), `test_direct_initializations` is performed now `N_REPEAT = 10` times. This could help catching similar segfaults in the future, if we change something in the logic of `_global_initialization.py`.
3. In `test_direct_initializations`, the second tested `inputs` now have `POWER_SPECTRUM="CLASS"` (+`K_MAX_FOR_CLASS=1`, to make the CLASS run short in time). This helps to test the extra allocations/initializations that are done when the CLASS settings are active.
4. Made the direct initializations private (with an underscore in the beginning of their names), to stress that these methods should not be called by users.

## Summary by Sourcery

Strengthen and clarify global initialization handling by privatizing low-level initialization methods and updating tests to prevent nondeterministic segfaults when invoking them directly.

Enhancements:
- Make direct initialization methods on the global initialization manager private and route the public init path through them to better encapsulate low-level behavior.

Tests:
- Rename the global initialization test to focus on direct initializations, ensure resources are freed between different input configurations, and repeat it multiple times to expose nondeterministic segfaults.
- Extend direct initialization tests to cover CLASS-based power spectrum settings and to use the new private initialization API.

### DIFF
--- a/tests/test_global_initialization.py
+++ b/tests/test_global_initialization.py
@@ -69,11 +69,11 @@ def test_direct_initializations(_run):
         )
     )
 
-    _GlobalInitManagerSingleton.broadcast_input_struct()
-    _GlobalInitManagerSingleton.initialize_power_spectrum()
-    _GlobalInitManagerSingleton.initialize_sigma_tables()
-    _GlobalInitManagerSingleton.initialize_heat()
-    _GlobalInitManagerSingleton.initialize_recombination_rate()
+    _GlobalInitManagerSingleton._broadcast_input_struct()
+    _GlobalInitManagerSingleton._initialize_power_spectrum()
+    _GlobalInitManagerSingleton._initialize_sigma_tables()
+    _GlobalInitManagerSingleton._initialize_heat()
+    _GlobalInitManagerSingleton._initialize_recombination_rate()
 
     assert _GlobalInitManagerSingleton.inputs_are_broadcast
     assert _GlobalInitManagerSingleton.ps_inited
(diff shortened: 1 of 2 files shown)


tests/test_global_initialization.py
```diff
@@ -46,12 +48,18 @@ def test_init():
...
 
 
-def test_initializations():
-    """Test that initializations work as expected."""
+@pytest.mark.parametrize("_run", range(N_REPEAT))
+def test_direct_initializations(_run):
+    """
+    Test that direct initializations work as expected.
+
+    We run this test several times because segfaults can still occur unexpectedly, even if one test passes smoothly.
```
steven-murray, line 56: I want a big note in here that calling these private functions is NOT a good idea, neither for users OR for ourselves in tests in general (because of potential segfaults).

tests/test_global_initialization.py
```diff
@@ -61,28 +69,35 @@ def test_initializations():
...
+    _GlobalInitManagerSingleton._initialize_recombination_rate()
 
     assert _GlobalInitManagerSingleton.inputs_are_broadcast
     assert _GlobalInitManagerSingleton.ps_inited
     assert not _GlobalInitManagerSingleton.sigma_inited
     assert _GlobalInitManagerSingleton.heat_inited
     assert not _GlobalInitManagerSingleton.recomb_inited
 
+    # NOTE: before initializing again below with different inputs, it is very important to free everything that was initialized,
+    # otherwise segfaults could occur!
```
steven-murray: Also clarify that these segfaults are not a problem with the user-facing logic of the code but rather due to us calling the private functions directly, which should never be done in real code.

steven-murray (commented):
Thanks @jordanflitter, I just have two very simple comments that I think will make it more clear what is happening.

steven-murray (approved):

PR #731, opened 2026-06-19
New enum for `v_cb`

### LINKED ISSUE #730: Change `v_cb` logic to a new enum called `V_CB_MODEL`
When `USE_MINI_HALOS=True`, the logic in the code is to determine the local `v_cb` value in the following way.
```python
if not inputs.matter_options.USE_RELATIVE_VELOCITIES and not inputs.astro_options.FIX_VCB_AVG:
   v_cb = 0
elif not inputs.matter_options.USE_RELATIVE_VELOCITIES:
   v_cb = inputs.astro_params.FIXED_VAVG
else:
   v_cb = v_cb_local #local fluctuating v_cb from initial conditions
```
It would be nicer if we could have an enum in `MatterOptions` called `V_CB_MODEL` with the following options: `“none”`, `“avg-auto”`, `“flucts”`, `“avg-debug”`. Also, it would be better to rename `FIXED_VAVG -> V_CB_AVG_DEBUG`. Then, the above logic would be replaced with
```python
if inputs.matter_options.V_CB_MODEL == "none":
   v_cb = 0
elif inputs.matter_options.V_CB_MODEL == "avg-auto":
   v_cb = inputs.cosmo_tables.V_CB_AVG
elif inputs.matter_options.V_CB_MODEL == “flucts”:
   v_cb = v_cb_local #local fluctuating v_cb from initial conditions
elif inputs.matter_options.V_CB_MODEL == “avg-debug”:
   v_cb = inputs.astro_params.V_CB_AVG_DEBUG
else:
   raise ValueError("V_CB_MODEL must be one of the following: “none”, “avg-auto”, “flucts”, “avg-debug”)
```

Note that `inputs.cosmo_tables.V_CB_AVG` is a new field in `CosmoTables`, which could easily be computed, given the `v_cb` transfer function from `CLASS`.

We could then throw some deprecation warning for the old parameters `USE_RELATIVE_VELOCITIES`, `FIX_VCB_AVG` and `FIXED_VAVG`.

jordanflitter:
Fix #730

This PR does the following.

* It made `USE_RELATIVE_VELOCITIES`, `FIX_VCB_AVG` and `FIXED_VAVG` deprecated. Deprecation warnings are displayed if the user sets these parameters to something other than `None`.
* The two booleans (`USE_RELATIVE_VELOCITIES`, `FIX_VCB_AVG`) are replaced by a new enum called `V_CB_MODEL` which has the following options: `”NONE”`, `”AVG-AUTO”`, “`FLUCTS”` and `”AVG-DEBUG”`. 
   - `”NONE”` corresponds to having `USE_RELATIVE_VELOCITIES=False` and `FIX_VCB_AVG=False`. 
   - `”FLUCTS”` corresponds to having `USE_RELATIVE_VELOCITIES=True`. 
   - `”AVG-DEBUG”` corresponds to having `USE_RELATIVE_VELOCITIES=False` and `FIX_VCB_AVG=True`, whereas a new float parameter called `V_CB_AVG_DEBUG` has replaced the deprecated `FIXED_VAVG` parameter. 
   - `”AVG-AUTO”` is a new feature, where the mean `v_cb` value that is used in the C code is determined on the fly from the CLASS output, consistently with the user’s chosen cosmological parameters.
* In the C code, `vcb_norel` was renamed to `vcb_const`. Moreover, `vcb_const` is defined only once and the following logic is defined only once in the C code (in `set_scaling_constants` in `scaling_relations.c`): `vcb_const=0` / `vcb_const=cosmo_tables->V_CB_AVG` / `vcb_const=astro_params->V_CB_AVG_DEBUG` if `V_CB_MODEL` is `”NONE”` / `”AVG-AUTO”` / `”AVG-DEBUG”`.
* The transfer function of `v_cb` is computed and allocated only if `V_CB_MODEL=“FLUCTS”`.
* The suppression in the matter power spectrum at high wavenumbers is applied if `V_CB_MODEL!=“NONE”` (unlike before, where it was applied only when `USE_RELATIVE_VELOCITY=True`, or `V_CB_MODEL=“FLUCTS”`, which I think it was a mistake)
* `SIGMAVCB` that normalizes `v_cb` in the MCG turnover mass calculations was replaced from a hard-coded 29 (km/sec) and is now computed on the fly, consistently with the user’s chosen cosmological parameters. This changes a bit the output (since for LCDM cosmology with Planck 2018 parameters we get SIGMAVCB=29.3 km/sec), so I had to reproduce some of the files in the database.
* New warning is raised if `V_CB_MODEL != "NONE"` and `USE_MINI_HALOS=False` (since `v_cb` only has impact on mini-halos). 
* Docstrings were updated to reflect the above logics.
* Tests were added to check the above logic works fine.
* Also, I shortened a bit the time it takes for `test_evolve` in `test_input_structs.py`, by explicitly setting `K_MAX_FOR_CLASS=0.01` (CLASS runs in that test multiple times since we inspect the content of `cosmo_tables` after they had been updated)

## Summary by Sourcery

Introduce a configurable v_cb modelling framework and modernize how relative-velocity effects are handled across the Python and C backends.

New Features:
- Add V_CB_MODEL option with NONE, AVG-AUTO, FLUCTS, and AVG-DEBUG modes to control the dark matter–baryon relative velocity treatment.
- Compute the mean and RMS v_cb from CLASS on the fly and expose V_CB_AVG via CosmoTables for use in simulations.
- Add V_CB_AVG_DEBUG astro parameter for fixed-debug v_cb configurations and wire it into turnover mass and global reionization calculations.

Bug Fixes:
- Ensure small-scale power-spectrum suppression from relative velocities is applied for all non-trivial v_cb models instead of only for fluctuating modes.
- Guard CLASS v_cb transfer function allocation and spline setup so it is only done when fluctuations are actually used, preventing invalid access when V_CB_MODEL disables them.

Enhancements:
- Deprecate USE_RELATIVE_VELOCITIES, FIX_VCB_AVG, and FIXED_VAVG in favor of V_CB_MODEL and V_CB_AVG_DEBUG, with compatibility shims and deprecation warnings.
- Refactor scaling and halo/minihalo logic to use a single vcb_const value derived from the chosen v_cb model and cosmological tables.
- Replace the hard-coded SIGMAVCB constant with an on-the-fly computation consistent with the chosen cosmology.
- Tighten validation and cross-struct consistency checks between V_CB_MODEL, FIX_VCB_AVG, USE_MINI_HALOS, and power spectrum settings, including a new warning when v_cb is enabled without mini-halos.
- Update template configurations, debug output, references, and I/O/serialization to reflect the new v_cb modelling interface and stored cosmological quantities.
- Slightly speed up evolve-related tests by lowering K_MAX_FOR_CLASS in the default test input parameters.

Documentation:
- Refresh parameter docstrings and high-level documentation for MatterOptions, AstroOptions, AstroParams, and CosmoTables to describe the new v_cb model, deprecations, and CLASS behaviour.

Tests:
- Expand and adjust unit tests to cover V_CB_MODEL behaviour, deprecated parameter warnings and removal guards, new validation paths, v_cb power spectrum access, serialization of V_CB_AVG, and mturns round-tripping under the new model.


src/py21cmfast/wrapper/inputs.py
```diff
@@ -1550,9 +1656,11 @@ class AstroParams(InputStruct):
...
     T_RE: float = field(default=2e4, converter=float)
-    FIXED_VAVG: float = field(
-        default=25.86, converter=float, validator=validators.gt(0)
+    _FIXED_VAVG: bool | None = field(
+        default=None, converter=attrs.converters.optional(float)
     )
+    V_CB_AVG_DEBUG: float = field(converter=float, validator=validators.gt(0))
+
     POP2_ION: float = field(default=5000.0, converter=float)
     POP3_ION: float = field(default=44021.0, converter=float)
```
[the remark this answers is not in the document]
jordanflitter: Thanks sourcery! You are right! I have changed the type of `FIXED_VAVG` from `bool | None` to `float | None`. I have also changed the type that `FIXED_VAVG` returns from `bool` to `float | None`.

src/py21cmfast/templates/Qin20.toml
```diff
@@ -6,7 +6,7 @@
 
 [MatterOptions]
 SOURCE_MODEL = 'E-INTEGRAL'
-USE_RELATIVE_VELOCITIES = false
```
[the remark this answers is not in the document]
jordanflitter: Thanks sourcery for this catch! You are absolutely right! I have changed V_CB_MODEL in Qin20/Park19 templates from `FLUCTS` to `NONE`.

tests/test_input_structs.py
```diff
@@ -197,6 +197,24 @@ def test_bad_input(self):
...
+        with pytest.warns(
+            deprecation.DeprecatedWarning, match="FIXED_VAVG is deprecated"
+        ):
+            astro_params = AstroParams(FIXED_VAVG=fixed_vavg)
+        assert fixed_vavg == astro_params.FIXED_VAVG
+        assert fixed_vavg == astro_params.V_CB_AVG_DEBUG
+
+    def test_fixed_vavg_is_removed(self):
+        """Fails when the removed_in version is reached, reminding you to delete FIXED_VAVG."""
+        AstroParams(FIXED_VAVG=1.0)
```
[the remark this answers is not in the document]
jordanflitter: Thanks sourcery! You are right! I have added `@deprecation.fail_if_not_removed` above `test_fixed_vavg_is_removed`

tests/test_input_structs.py
```diff
@@ -646,10 +726,49 @@ def test_validation_warnings(self, msg, kw):
...
+        assert v_cb_model == inputs.matter_options.V_CB_MODEL
+        assert fix_vcb_avg == inputs.astro_options.FIX_VCB_AVG
+
+    def test_fix_vcb_avg_is_removed(self):
+        """Fails when the removed_in version is reached, reminding you to delete FIX_VCB_AVG."""
+        InputParameters(
+            random_seed=1,
+            astro_options=AstroOptions(FIX_VCB_AVG=True),
+            matter_options=MatterOptions(V_CB_MODEL="AVG-AUTO"),
+        )
```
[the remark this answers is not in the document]
jordanflitter: Thanks sourcery! You are right! I have added `@deprecation.fail_if_not_removed` above `test_fix_vcb_avg_is_removed`

src/py21cmfast/src/thermochem.c
```diff
@@ -289,6 +288,11 @@ double lyman_werner_threshold(float z, float J_21_LW, float vcb) {
     double mcrit_noLW = 3.314e7 * pow(1. + z, -1.5);
     double f_LW = 1.0 + astro_params_global->A_LW * pow(J_21_LW, astro_params_global->BETA_LW);
 
+    // v_cb is normalized by the rms value of the DM-b relative velocity at kinematic decoupling,
+    // which, under the assumption of Maxwell-Boltzmann distribution, can be achieved by multiplying
+    // the mean v_cb by sqrt(3pi / 8). This gives about 29.3 km/s for LCDM cosmology with Planck
+    // 2018 parameters
+    double SIGMAVCB = cosmo_tables_global->V_CB_AVG * sqrt(3 * M_PI / 8);
```
steven-murray, line 295: So, in the past, if V_CB_AVG was zero, this still would have been 29.0, whereas now it will be zero. THis seems more consistent, but is it what we want?

src/py21cmfast/wrapper/inputs.py
```diff
@@ -132,6 +133,10 @@ def vld(inst, att, val):
     name="Planck18",
 )
 
+# The mean value of the amplitude of the relative velocity between baryons and cold dark matter at kinematic decoupling, in km/s,
+# for LCDM Planck 2018 cosmology
+V_CB_AVG_DEFAULT = 27.0
```
steven-murray, line 138: I'm confused, should this be 29.3 km/s, or should we leave it at the previous default (25.9 or so)? Why change it to a new value here?

src/py21cmfast/wrapper/inputs.py
```diff
@@ -699,6 +724,18 @@ class MatterOptions(InputStruct):
     KEEP_3D_VELOCITIES
         Whether to keep the 3D velocities in the ICs.
         If False, only the z velocity is kept.
+    V_CB_MODEL
+        The model to use for the relative velocity field between the (cold) dark matter and the baryons,
+        at the moment of kinematic decoupling. Options are:
+
+        * ``NONE``: Zero relative velocity is assumed.
+        * ``AVG-AUTO``: The mean of the amplitude of the relative velocity is used in every cell in the box. It computed from linear theory,
```
steven-murray:
```suggestion
        * ``AVG-AUTO``: The mean of the amplitude of the relative velocity is used in every cell in the box. It is computed from linear theory,
```

src/py21cmfast/wrapper/inputs.py
```diff
@@ -762,13 +802,59 @@ class MatterOptions(InputStruct):
...
-        return "CLASS" if self.USE_RELATIVE_VELOCITIES else "EH"
+        return "CLASS" if self.V_CB_MODEL == "FLUCTS" else "EH"
 
     @POWER_SPECTRUM.validator
     def _POWER_SPECTRUM_vld(self, att, val):
-        if self.USE_RELATIVE_VELOCITIES and val != "CLASS":
+        if self.V_CB_MODEL == "FLUCTS" and val != "CLASS":
             raise ValueError(
-                "Can only use 'CLASS' power spectrum with relative velocities"
+                "Can only use 'CLASS' power spectrum with V_CB_MODEL='FLUCTS'! "
```
steven-murray:
```suggestion
                "When using V_CB_MODEL='FLUCTS', you must use POWER_SPECTRUM = 'CLASS'! "
```

src/py21cmfast/wrapper/inputs.py
```diff
@@ -762,13 +802,59 @@ class MatterOptions(InputStruct):
...
+                "Please set POWER_SPECTRUM to 'CLASS' or change V_CB_MODEL to something else."
+            )
+
+    @cached_property
+    def USE_RELATIVE_VELOCITIES(self) -> bool:
+        """Whether to use the fluctuating field of the relative velocity between (cold) dark matter and baryons.
+
+        This is a deprecated property, and will be removed in a future version. Please use V_CB_MODEL instead.
```
steven-murray:
```suggestion
        This is a deprecated property, and will be removed in v5. Please use V_CB_MODEL instead.
```

src/py21cmfast/wrapper/inputs.py
```diff
@@ -1240,6 +1329,17 @@ class AstroOptions(InputStruct):
...
+    @cached_property
+    def FIX_VCB_AVG(self) -> bool:
+        """Whether to fix the amplitude of the relative velocity between (cold) dark matter and baryons on a constant mean value from linear perturbation theory.
+
+        This is a deprecated property, and will be removed in a future version. Please use V_CB_MODEL instead.
+        """
+        if self._FIX_VCB_AVG is None:
+            return None
+        else:
+            return self._FIX_VCB_AVG
```
steven-murray:
```suggestion
        return self._FIX_VCB_AVG
```

src/py21cmfast/wrapper/inputs.py
```diff
@@ -1808,15 +1975,52 @@ def _astro_options_validator(self, att, val):
...
+                    "FIX_VCB_AVG",
+                    deprecated_in="4.3.0",
+                    removed_in="5.0.0",
+                    details=(
+                        "FIX_VCB_AVG is deprecated and will be removed in a future version. "
+                        "Please use V_CB_MODEL directly instead."
+                    ),
+                ),
+                stacklevel=2,
+            )
```
steven-murray, line 2004: Didn't you already raise a warning about this in tihe AstroOptions class?

src/py21cmfast/wrapper/inputs.py
```diff
@@ -1808,15 +1975,52 @@ def _astro_options_validator(self, att, val):
...
+                not self.astro_options.FIX_VCB_AVG
+                and self.matter_options.V_CB_MODEL == "AVG-AUTO"
+            ):
+                raise ValueError(
+                    "FIX_VCB_AVG=False is not compatible with V_CB_MODEL == 'AVG-AUTO'! "
+                    "Either change V_CB_MODEL or set FIX_VCB_AVG to True."
+                )
+            elif (
+                self.astro_options.FIX_VCB_AVG
+                and self.matter_options.V_CB_MODEL != "AVG-AUTO"
```
steven-murray: Shouldn't `FIX_VCB_AVG` also be compatible with `AVG-DEBUG`?

steven-murray (commented):
Thanks @jordanflitter. Not bad for a day's work! Just a few comments.

src/py21cmfast/src/thermochem.c
jordanflitter: No, `SIGMAVCB` can't be zero. This is because `SIGMAVCB` is proportional to `cosmo_tables.V_CB_AVG` and this quantity is either `V_CB_AVG_DEFAULT=27 km/s` if `POWER_SPECTRUM != "CLASS"`, or it is computed directly from the CLASS output.
Note that `cosmo_tables.V_CB_AVG` can never be zero, unlike `astro_params.V_CB_AVG_DEBUG`.

src/py21cmfast/wrapper/inputs.py
jordanflitter: Do not confuse between the RMS of the Maxwell-Boltzmann distribution, which is 29.3 km/s, and the mean, which is the RMS times `sqrt(8/3pi)=0.92`, making it roughly 27 km/s.

src/py21cmfast/wrapper/inputs.py
jordanflitter: Thanks! Done!

src/py21cmfast/wrapper/inputs.py
jordanflitter: Done!

src/py21cmfast/wrapper/inputs.py
jordanflitter: Done!

src/py21cmfast/wrapper/inputs.py
jordanflitter: Of course :) thanks! Done!

src/py21cmfast/wrapper/inputs.py
jordanflitter: No, but the docstring of `FIX_VCB_AVG` in `AstroOptions` does mention it is a deprecated property. Maybe this is what you meant?
The reason why this deprecation warning is found under `_astro_options_validator` in `InputParameters` (and not under the cached property `FIX_VCB_AVG` in `AstroOptions`) is because we have some logics to check the compatibility of this flag with `V_CB_MODEL`, but because the latter is found in `MatterOptions`, these checks must be under `InputParameters`. Also, in these checks it is assumed that `FIX_VCB_AVG is not None`, and since python have this weird behavior that `not None = True`, I found it convenient to put these checks under the deprecation warning, which is why I put this warning under `InputParameters`.

src/py21cmfast/wrapper/inputs.py
jordanflitter: You are right! It was a bug! `FIX_VCB_AVG` should be compatible with ONLY `AVG-DEBUG`. Thanks! Done!

jordanflitter:
Thanks @steven-murray for your quick comments! I think it's now ready for round 2 :)

src/py21cmfast/wrapper/inputs.py
steven-murray: Got it.

steven-murray (approved):

Issue #732, opened 2026-06-26
Turnover mess, part 2


jordanflitter:
As part of cleaning the turnover mess (see #729), I found that the signature of `initialise_Nion_Conditional_spline` is different than the signature of `initialise_SFRD_Conditional_table`, even though both integrals are very similar and differ by only one extra power-law in the former to account for the mass-dependent escape fraction:
```c
void initialise_Nion_Conditional_spline(double z, double min_density, double max_density,
                                        double Mmin, double Mmax, double Mcond,
                                        double log10Mturn_min, double log10Mturn_max,
                                        double log10Mturn_min_MINI, double log10Mturn_max_MINI,
                                        ScalingConstants *sc, bool prev);
void initialise_SFRD_Conditional_table(double z, double min_density, double max_density,
                                       double Mmin, double Mmax, double Mcond,
                                       ScalingConstants *sc);
```
Firstly, there is an extra boolean called `prev` in `initialise_Nion_Conditional_spline`. The reason for it is to allow to initialize the previous conditional Nion table as part of the trapezoidal integration that is done in `IonisationBox.c` in the old v3 source models (see #147).

But more crucially, `initialise_Nion_Conditional_spline` expects to receive several more `double` variables, which represent the minimum and maximum turnover masses for ACGs and MCGs. This is required because when `USE_MINI_HALOS=True`, the MCG interpolation table depends also on the MCG turnover mass, since it can fluctuate due to `v_cb` and LW flux, but also due to reionization feedback, and the latter is never taken into account in the SFRD integrals, even in our most modern source models.  In other words, **even if we compute correctly the ACG turnover mass using the inhomogeneous reionization feedback, it is never taken into account when we compute the conditional SFRD in ACGs**, regardless if we use interpolation tables or not. This is also manifest from comparing the signatures of `EvaluateNion_Conditional` and `EvaluateSFRD_Conditional`, and realizing that there is no `log10Mturn` input in the latter:
```c
double EvaluateNion_Conditional(double delta, double log10Mturn, double growthf, double M_min,
                                double M_max, double M_cond, double sigma_max, ScalingConstants *sc,
                                bool prev)
double EvaluateSFRD_Conditional(double delta, double growthf, double M_min, double M_max,
                                double M_cond, double sigma_max, ScalingConstants *sc)
```
 **This is a bug!** There's even a comment in `get_cell_integrals` in `HaloBox.c` that says:
```c
[...]

Issue #470, opened 2025-03-17
Reionization feedback is ignored during spin-temperature and when USE_MINI_HALOS is False


daviesje:
Currently in 21cmfast, reionization feedback is totally ignored in two cases:

1. when calculating the spin temperatures.
2. if minihalos are turned off.

I am guessing the rationale behind (1) is since the spin temperature is only calculated in neutral regions, we don't need it. It does impact the global neutral fraction used for `tauX`. But this should only have an effect long after reheating finishes so I don't think theres a big impact there.

However, the halo model has allowed outputs of stellar mass/SFR directly, and the integrated (low-mass) components use the same tables as spin temperature. This means that the integrated ACG component of the output SFR is inconsistent with the rest.

I don't understand (2) so much. I guess the USE_MINI_HALOS flag adds the 2D (delta,Mturn) tables but I'm not 100% sure if this was intentional. So if the flag is off, the turnover mass for all halos is constant at `astro_params.M_TURN`

jordanflitter:
Note that this issue will be fixed when #668 is fixed (though it won't introduce the new astro option that will control whether reionization feedback is taken into account, we could keep discussing if we still want to have it together with the fix of #668).

jordanflitter:
It's important to clarify that while fixing #668 will help to solve this issue, it will only use the correct ACG turnover mass (that may or may not account for the reionization feedback) in evaluating the SFRD. However, there are two more places in `SpinTemperatureBox.c` that needs the ACG turnover mass:
1. In `global_reion_properties`, in determining the neutral volume filling factor (`Q_HI`) for the `nu_tau_one` (or `nu_tau_one_MINI`) calculations and for the `NO_LIGHT` condition.
2. In `nu_tau_one` (or `nu_tau_one_MINI`), in order to determine the lower limit for the frequency integral.

These places would still need to be modified, even after #668 is fixed. I made some comments in `SpinTemperatureBox.c` and `interp_tables.c` to remind us to still modify these places, even after #668 is fixed.

PR #737, opened 2026-07-02
Cleaning the turnover mess

### LINKED ISSUE #729: Turnover mess
We use turnover masses to regulate star formation in too massive/light halos ("duty cycle"). This the mathematical model we use for the duty cycle:

$$
f_\mathrm{duty}^\mathrm{ACG}\left(M_h\right)=\exp\left(-M_\text{turn-lower}^\mathrm{ACG}/M_h\right)
$$
$$
f_\mathrm{duty}^\mathrm{MCG}\left(M_h\right)=\exp\left(-M_\text{turn-lower}^\mathrm{MCG}/M_h\right)\exp\left(-M_h/M_\text{turn-upper}^\mathrm{MCG}\right)
$$

Apparently, as was noted in #688, there have been some changes in the logic in how to determine the turnover masses between v3 and v4, at least when mini-halos are turned on (this creates some confusion, like in #726).

### Turnover masses in spin temperature calculations in v4 (`SOURCE_MODEL=“E-INTEGRAL”`, `USE_MINI_HALOS=True`)
* ACG lower turnover mass is given by the maximum between `M_TURN` and the output of `atomic_cooling_threshold(z’’)`
* MCG upper turnover mass is given by the output of `atomic_cooling_threshold(z’’)`
* MCG lower turnover mass is given by (the filtered version of) the maximum between `M_TURN` and the output of `lyman_werner_threshold(z’)` (note the latter function is evaluated at `z'`, not `z''`)

### Turnover masses in spin temperature calculations in v3 (`USE_MINI_HALOS=True`)
* ACG lower turnover mass is given by the output of `atomic_cooling_threshold(z’’)` - **This is a bit different than what we have in v4!** (since there is no `max` operation between that value and `M_TURN`)
* MCG upper turnover mass is given by the output of `atomic_cooling_threshold(z’’)`
* MCG lower turnover mass is given by (the filtered version of) the maximum between between the output of `lyman_werner_threshold(z’)` with the local LW flux and `v_cb`, and the output of `lyman_werner_threshold(z’)` with zero LW flux and `VAVG` (if we have `FIX_VCB_AVG=True`). There is also protection that this value will not be greater than `pow(10, LOG10_MTURN_MAX)=1e10.` - **This is different than what we have in v4!**

Moreover, the logic for the turnover masses in v4 depends on the chosen `SOURCE_MODEL`. For example, if `SOURCE_MODEL="L-INTEGRAL"` (or has discrete halos) and `USE_MINI_HALOS=True`, then the ACG (MCG) lower turnover mass is taken to be the maximum between (1) `M_TURN`, (2) reionization feedback mass and (3) output of `atomic_cooling_threshold(z)` (`lyman_werner_threshold(z)`). The difference in the logics between different source models is confusing and makes it much more difficult to compare between them!

Finally, in both v3 and v4, if `USE_MINI_HALOS=False`, then the ACG lower turnover mass is just `M_TURN`, while completely ignoring the atomic cooling threshold and the reionization feedback. It's very confusing that ACG properties (e.g. emissivities/fluxes) depend on whether or not MCGs/mini-halos are present in the simulation. 

After discussing with @andreimesinger and @steven-murray we agreed that we should:

- [x] Rename the astro param `M_TURN` -> `M_TURN_STELLAR_FEEDBACK`, since the current name is very confusing! (we should have a deprecation warning that tells users that `M_TURN` is deprecated)
- [ ] Set `M_TURN_STELLAR_FEEDBACK` default to 5 if `USE_MINI_HALOS=True`, or 8.7 otherwise. **Update**: this is not possible since `M_TURN_STELLAR_FEEDBACK` lives in `AstroParams`, while `USE_MINI_HALOS` lives in `AstroOptions`!
- [x] Raise a warning if the user sets `M_TURN_STELLAR_FEEDBACK` on a value larger than 6 with `USE_MINI_HALOS=True`. **Update**: apparently we already had this warning (with the lower threshold being 8). I did however change the warning message to stress that such configuration doesn't make sense!
- [x] Fix the above logic in v4! The ACG lower turnover mass should be identical to the MCG upper turnover mass!
- [x] Fix the above logic in v4! Make the logic for the turnover masses to be the same, regardless the chosen `SOURCE_MODEL` (up to the known bug in `SpinTemperatureBox.c` where the reionization feedback cannot be accounted since that module does not receive the previous `IonizedBox` - see #470).
- [x] Introduce a new `REIONIZATION_FEEDBACK_MODEL` enum with the options `"NONE"`, "`ACG"`, `"MCG"`, and `"BOTH"`. This would make it easier to determine whether the ACG turnover mass should be inhomogeneous or not.
- [x] Fix the above logic in v4! The ACG lower turnover mass should be the same, regardless if `USE_MINI_HALOS` is True or False.
- [x] In a new markdown file called `M_TURN.rst`, document the current logic of the turnover masses in v4 + their logic history in previous versions. Specify any current inconsistencies we have in the code at the moment (e.g. #470).
- [x] Also mention the logic change between v3 and v4 in `updates_from_v3.rst`.

jordanflitter:
Fix #729 and #732.

This PR changes the following:

- Replaced `M_TURN` with `M_TURN_STELLAR_FEEDBACK` and made `M_TURN` as a deprecated parameter with the appropriate deprecation warning.
- Introduced a new enum called `REIONIZATION_FEEDBACK_MODEL` with the options (1) `"NONE"`, (2) `"ACG"`, (3) `"MCG"` and (4) `"BOTH"`. This new enum helps in breaking the coupling between `USE_MINI_HALOS` and the application of the reionization feedback on the ACG and MCG turnover masses.
- Renamed `lyman_werner_threshold` to `molecular_cooling_threshold_with_feedbacks` and `mturn_a_nofb` to `mturn_acg_homogeneous`.
- **Main change**: Changed the logic for the ACG turnover mass, it is now taken from the maximum between (1) `M_TURN_STELLAR_FEEDBACK`, (2) output of `atomic_cooling_threshold` and (3) `reionization_feedback` (when applicable). This is now the logic in all source models, except `CONST-ION-EFF`, and regardless the value of `USE_MINI_HALOS`. An exception still occurs for `E-INTEGRAL` in `SpinTemperatureBox.c`, where the reionization feedback is not applied due to #470.
- **Main change**: Changed the logic for the ACG turnover mass, it is now taken from the maximum between (1) `M_TURN_STELLAR_FEEDBACK`, (2) output of `molecular_cooling_threshold_with_feedbacks` and (3) `reionization_feedback` (when applicable). This is now the logic in all source models, except `CONST-ION-EFF`, and regardless the value of `USE_MINI_HALOS`. An exception still occurs for `E-INTEGRAL` in `SpinTemperatureBox.c`, where the reionization feedback is not applied due to #470.
- If reionization feedback is not applied on ACGs (regardless the value of `USE_MINI_HALOS`), their turnover mass is homogeneous and is taken from `mturn_acg_homogeneous`, which is now always the maximum between (1) `M_TURN_STELLAR_FEEDBACK`, and (2) output of `atomic_cooling_threshold`. The MCG turnover mass is always inhomogeneous.
- **Main change**: The MCG upper turnover mass is now always the ACG lower turnover mass.
- **Main change**: if the MCG turnover mass is greater than the ACG turnover mass, the MCG contribution is zero. 
- Refactored the calculations of the inhomogeneous turnover masses in the code such that they will be obtained from a single place - `compute_mturns_inhomogeneous` in `thermochem.c` (exception still occurs for `E-INTEGRAL` in `SpinTemperatureBox.c` due to #470).
- Also modified `ComuteLF` to reflect the above changes in the turnover mass logic.
- Standardized the signatures and calls to `EvaluateSomething` (conditional/unconditional SFRD/Nion) to follow the above logic as well as the initialization of the interpolation tables.
- In `cfuncs.py`, the following functions were added: `get_atomic_cooling_mass_threshold`, `get_molecular_cooling_threshold_with_feedbacks` and `get_reionization_feedback`. These are helpful functions for testing that the way the turnover masses are computed in the C code follow the exact same logic that is described above (in `test_cfuncs.py::test_compute_mturns_model`, this test would fail if we ever change the turnover masses model in the C code).
- Also refactored similar code lines from various functions in `cfuncs.py` into a new function called `get_log10mturns_helper`.
- Fixed some tests in `test_c_interpolation_tables.c` and added a few more, due to the introduction of `REIONIZATION_FEEDBACK_MODEL` in the code.
- Add a comprehensive review that documents how the turnover masses were computed in previous and current versions of `21cmFAST`. This review is found in a new markdown file called `M_TURN.rst` and is referred in the docs from `updates_from_v3.rst`. The latter markdown file was also modified to include other changes that were made between v3 and v4, and is now also mentioned in `misc.rst` under a new section called "I get different results compared to older versions of 21cmFAST, why?".

Note that all the integration tests keep passing without changing the referenced data files. This is because in our tests:
1. We use the default `M_TURN_STELLAR_FEEDBACK=8.7` when `USE_MINI_HALOS=False`. This is a very high value, much higher than the atomic cooling threshold.
2. We were careful in setting `M_TURN_STELLAR_FEEDBACK=5` when `USE_MINI_HALOS=True`, so the molecular cooling threshold with feedbacks is always more dominant in our tests in this configuration.
3. We always finish the tests at z=18, which is a redshift too high for the reionization feedback to become dominant.
So despite all the above changes in the logics, our tests keep using effectively the same turnover mass as before.


docs/faqs/misc.rst
```diff
@@ -9,6 +9,17 @@ Historically, ``21cmFAST`` was initially designed to simulate very fast reioniza
 
 In general, before running a full lightcone simulation via ``run_lightcone``, it's a good practice to make a quick calculation with ``run_global_evolution`` (find more information and limitations of this feature in `the global evolution tutorial <../tutorials/global_evolution.ipynb>`_).
 
+I get different results compared to older versions of 21cmFAST, why?
+--------------------------------------------------------------------
+
+``21cmFAST`` has undergone significant changes since its initial release, and the results of the code
+have changed accordingly. Specifically, for new v4 users, it is important to note that a new
+``SOURCE_MODEL`` parameter was introduced in ``MatterOptions``. In order to mimic the behavior of
+the code in v3, you can either set ``SOURCE_MODEL`` on ``"CONST-ION-EFF"`` or ``"E-INTEGRAL"``,
```
steven-murray:
```suggestion
the code in v3, you can either set ``SOURCE_MODEL`` to ``"CONST-ION-EFF"`` or ``"E-INTEGRAL"``,
```

docs/updates_from_v3.rst
```diff
@@ -14,8 +14,9 @@ discrete source field which is then used in the spin temperature and ionization
 calculations. This not only includes the effects of stochasticity in the IGM observables,
 but also creates several new outputs which can be further used in forecasting galaxy
 survey, line intensity mapping, and cosmic background statistics. The sampler is
-activated with the flag ``HALO_STOCHASTICITY`` and serves as a faster replacement to the
-previous excursion-set halo finder, with greatly increased functionality. Halos are
+activated with setting ``SOURCE_MODEL`` on ``"CHMF-SAMPLER"`` and serves as a faster
```
steven-murray:
```suggestion
activated with setting ``SOURCE_MODEL`` to ``"CHMF-SAMPLER"`` and serves as a faster
```

docs/updates_from_v3.rst
```diff
@@ -14,8 +14,9 @@ discrete source field which is then used in the spin temperature and ionization
...
+replacement to the previous excursion-set halo finder (which is still usable by
+``SOURCE_MODEL`` on ``"DEXM-ESF"``), with greatly increased functionality. Halos are
```
steven-murray:
```suggestion
``SOURCE_MODEL="DEXM-ESF"``), with greatly increased functionality. Halos are
```

docs/updates_from_v3.rst
```diff
@@ -25,11 +26,20 @@ their galaxy properties. Converting from the RNG to the properties can be done w
 directly stored in these objects for efficiency and so we can correctly account for
 feedback in the forward time-loop.
 
+A similar source model to the one descrobed above, albeit without discrete halos, is
+named ``"L-INTEGRAL"``. ``21cmFAST`` v4 also includes source models that mimic the
+behavior of the code in v3, which are ``"CONST-ION-EFF"`` and ``"E-INTEGRAL"``. The
+former corresponds to setting ``USE_MASS_DEPENDENT_ZETA`` on ``False`` in v3, while
```
steven-murray:
```suggestion
former corresponds to setting ``USE_MASS_DEPENDENT_ZETA`` to ``False`` in v3, while
```

docs/updates_from_v3.rst
```diff
@@ -25,11 +26,20 @@ their galaxy properties. Converting from the RNG to the properties can be done w
...
+the latter corresponds to setting ``USE_MASS_DEPENDENT_ZETA`` on ``True`` in v3.
```
steven-murray:
```suggestion
the latter corresponds to setting ``USE_MASS_DEPENDENT_ZETA`` to ``True`` in v3.
```

docs/updates_from_v3.rst
```diff
@@ -25,11 +26,20 @@ their galaxy properties. Converting from the RNG to the properties can be done w
...
+
 The conditional mass functions used to perform integrals have been extended with the
 Sheth-Tormen CHMF (Sheth+2002) which has been applied to ``21cmFAST`` in both halo and
 grid based source models, when the user sets ``HMF=='ST'``. All other mass functions
 rescale the Extended Press-Schechter (EPS) conditional mass function.
 
+It is also important to note that the logics for determining the turnover masses
```
steven-murray:
```suggestion
It is also important to note that the logic for determining the turnover masses
```

docs/updates_from_v3.rst
```diff
@@ -94,10 +104,68 @@ templates, but they are briefly described below:
...
 * ``AstroOptions``: previously ``FlagOptions``, these enable different modules which
   *only* affect the spin temperature and ionisation fields.
 
+Changes in defaults
+====================
+
+Some of the default values for parameters have changed between v3 and v4. For example:
+
+* ``HII_DIM``: 200 in v3, 256 in v4
+
```
steven-murray:
```suggestion
* ``HII_DIM``: 200 in v3, 256 in v4
```

(and so on for the rest of the list)

docs/updates_from_v3.rst
```diff
@@ -94,10 +104,68 @@ templates, but they are briefly described below:
...
+The default value of some of the astrophysical parameters have also changed, for example:
+
+* ``F_STAR7_MINI``: -2.0 in v3, :math:`\text{F\_STAR10} - 3 \times \text{ALPHA\_STAR} = -2.5` in v4
+
+* ``L_X``: 40.0 in v3, 40.5 in v4
+
+* ``L_X_MINI``: 40.0 in v3, ``L_X`` = 40.5 in v4
+
+* ``R_BUBBLE_MAX``: 15.0 or 50.0 depending on ``INHOMO_RECO`` in v3, 15.0 in v4
```
steven-murray, line 131:
```suggestion
* ``F_STAR7_MINI``: -2.0 in v3, :math:`\text{F\_STAR10} - 3 \times \text{ALPHA\_STAR} = -2.5` in v4
* ``L_X``: 40.0 in v3, 40.5 in v4
* ``L_X_MINI``: 40.0 in v3, ``L_X`` = 40.5 in v4
* ``R_BUBBLE_MAX``: 15.0 or 50.0 depending on ``INHOMO_RECO`` in v3, 15.0 in v4
```

docs/updates_from_v3.rst
```diff
@@ -94,10 +104,68 @@ templates, but they are briefly described below:
...
+Users can now quickly compute the global evolution of the fields in ``21cmFAST``
+by calling the function :func:`~py21cmfast.run_global_evolution`.
+See :ref:`tutorials/global_evolution` for more details.
+
+Redshift space distortions
+===========================
+In v3, the redshift space distortions were applied to the brightness temperature
+coeval box, by setting on True the flag ``SUBCELL_RSD``. In v4, redshift space distortions and
+velocity effects are applied to the brightness temperature lightcone box, by setting on True the keywrods
+arguments ``apply_rsds`` and ``include_dvdr_in_tau21`` in the function :func:`~py21cmfast.run_lightcone`.
```
steven-murray:
```suggestion
velocity effects are applied to the brightness temperature lightcone box, by setting the keyword
arguments ``apply_rsds`` and ``include_dvdr_in_tau21`` to ``True`` in the function :func:`~py21cmfast.run_lightcone`.
```

src/py21cmfast/src/HaloBox.c
```diff
@@ -134,10 +134,10 @@ int get_uhmf_averages(double M_min, double M_max, double M_turn_a, double M_turn
     intgrl_stars_only = Nion_General(consts->redshift, lnMmin, lnMmax, M_turn_a, &consts_sfrd);
     if (astro_options_global->USE_MINI_HALOS) {
         intgrl_fesc_weighted_mini =
-            Nion_General_MINI(consts->redshift, lnMmin, lnMmax, M_turn_m, consts);
+            Nion_General_MINI(consts->redshift, lnMmin, lnMmax, M_turn_a, M_turn_m, consts);
```
steven-murray: I'd prefer `M_turn_acg` over `M_turn_a`

src/py21cmfast/src/HaloBox.c
```diff
@@ -263,23 +263,21 @@ void get_cell_integrals(double dens, double l10_mturn_a, double l10_mturn_m,
...
-        EvaluateSFRD_Conditional(dens, growth_z, M_min, M_max, M_cell, sigma_cell, consts);
-    // TODO: SFRD tables still assume no reion feedback, this should be fixed
-    //   although it doesn't affect the histories (only used in Ts) it makes outputs wrong
-    //   for post-processing
+    properties->stellar_mass = EvaluateSFRD_Conditional(dens, l10_mturn_a, growth_z, M_min, M_max,
+                                                        M_cell, sigma_cell, consts);
     if (astro_options_global->USE_MINI_HALOS) {
         properties->stellar_mass_mini = EvaluateSFRD_Conditional_MINI(
-            dens, l10_mturn_m, growth_z, M_min, M_max, M_cell, sigma_cell, consts);
+            dens, l10_mturn_a, l10_mturn_m, growth_z, M_min, M_max, M_cell, sigma_cell, consts);
```
steven-murray: same comment here

src/py21cmfast/src/HaloBox.c
```diff
@@ -463,61 +437,78 @@ void halobox_debug_print_avg(HaloBox *halobox, ScalingConstants *consts, double
...
-    // for mean fixing. It is the volume-weighted mean of LOG10 Mturn, although we could do another
-    // weighting or use Mturn directly None of these are a perfect representation due to the
-    // nonlinear way turnover mass affects N_ion
-    log10_mturn_a_avg /= HII_TOT_NUM_PIXELS;
-    log10_mturn_m_avg /= HII_TOT_NUM_PIXELS;
-    averages[0] = log10_mturn_a_avg;
-    averages[1] = log10_mturn_m_avg;
+    if (uses_reionization_feedback_in_acgs(astro_options_global->REIONIZATION_FEEDBACK_MODEL)) {
+        // NOTE: This average log10 Mturn will be passed onto the spin temperature calculations
+        // where It is used to perform the frequency integrals (over tau, dependent on <XHI>), and
```
steven-murray:
```suggestion
        // where it is used to perform the frequency integrals (over tau, dependent on <XHI>), and
```

src/py21cmfast/src/interp_tables.c
```diff
@@ -341,163 +352,228 @@ void initialise_Nion_Conditional_spline(double z, double min_density, double max
...
-        if (!astro_options_global->USE_MINI_HALOS) continue;
 
-        for (k = 0; k < NMTURN; k++) {
-            if (isfinite(SFRD_conditional_table_MINI.z_arr[i][k]) == 0) {
-                LOG_ERROR("Detected either an infinite or NaN value in MCG SFRD conditional table");
-                Throw(TableGenerationError);
+            if (astro_options_global->USE_MINI_HALOS) {
+                for (j = 0; j < NMTURN; j++) {
+                    // NOTE: we use below homogeneous (feedback-free) ACG turnover mass, because if
+                    // the reionization reionization feedback dominates, then the the turnover
```
steven-murray:
```suggestion
                    // the reionization feedback dominates, then the turnover
```

src/py21cmfast/src/photoncons.c
```diff
@@ -165,10 +165,14 @@ int InitialisePhotonCons() {
                 // We Force QAG due to the changing limits and messy implementation which I will fix
                 // later (hopefully move the whole thing to python)
                 if (source_model_is_mass_dependent(matter_options_global->SOURCE_MODEL)) {
+                    // TODO: the passed ACG turnover mass that we give below has to be updated to
```
steven-murray, line 168: Flagging this as a thign we ought to actually do

src/py21cmfast/wrapper/cfuncs.py
```diff
@@ -79,6 +79,118 @@ def get_halo_catalog_buffer_size(
...
+            )
+
+    vfunc = np.vectorize(
+        lib.molecular_cooling_threshold_with_feedbacks, otypes=[np.float64]
+    )
+    return vfunc(redshifts, J_LW_21, v_cb)
+
+
+@init_c_state(broadcast_inputs=True)
+def get_reionization_feedback(
```
steven-murray:
```suggestion
def get_reionization_feedback_mass(
```

src/py21cmfast/wrapper/cfuncs.py
```diff
@@ -1257,3 +1327,79 @@ def return_chmf_value(
         sigma[None, :, None],
         inputs.matter_options.cdict["HMF"],
     )
+
+
+def get_log10mturns_helper(
```
steven-murray: Why don't we just call this `compute_logmturns`?

steven-murray (commented):
This is really great, thank you so much @jordanflitter.

I really like the new document outlining the whole M_TURN evolution over releases, well done!

One thing I would say is that we should make it clear that f_duty should be read as the probability of a halo of mass m_h hosting a galaxy (of either ACG or MCG type). The explicit reason that the MCG's turnover for m>M_TURN_ACG is because it is more probable for them to form ACGs than MCGs. We should also make sure that f_ACG+f_MCG is never greater than unity...

steven-murray:
Quick update:

It is true that the probability of hosting a galaxy (whether MCG or ACG) is never greater than unity in this model. 

Furthermore, the conditional probability of hosting an MCG, given that a galaxy is hosted at all, in this model is

```
p(mcg|g) = exp((Macg-Mmcg)/m - m/Macg)
```

When Mmcg=Macg, this has the form `p(mcg|g) = e^(-m/Macg)` which I think is very reasonable (i.e. the probability of the galaxy being an MCG above the ACG turnover mass falls exponentially with increasing mass). 

So I think this model makes sense. The only thing that probably doesn't really make sense is to force there to be no MCG's when Mmcg>Macg. Without forcing it, the number of MCG's will naturally die off exponentially as Mmcg grows above Macg anyway, but will do so smoothly.

docs/faqs/misc.rst
jordanflitter: Done

docs/updates_from_v3.rst
jordanflitter: Done

docs/updates_from_v3.rst
jordanflitter: Done

docs/updates_from_v3.rst
jordanflitter: Done

docs/updates_from_v3.rst
jordanflitter: Done

docs/updates_from_v3.rst
jordanflitter: Done

docs/updates_from_v3.rst
jordanflitter: Done

docs/updates_from_v3.rst
jordanflitter: Done

docs/updates_from_v3.rst
jordanflitter: Done

src/py21cmfast/src/HaloBox.c
jordanflitter: I agree, I changed that throughout the codespace.

src/py21cmfast/src/HaloBox.c
jordanflitter: I agree, I changed that throughout the codespace.

src/py21cmfast/src/HaloBox.c
jordanflitter: Done

src/py21cmfast/src/photoncons.c
jordanflitter: We agreed to keep that comment there, because the photon non-conservation correction will be completely rewritten in the future (#364)

src/py21cmfast/wrapper/cfuncs.py
jordanflitter: Done

src/py21cmfast/wrapper/cfuncs.py
jordanflitter: I changed it to `_ get_log10mturns_helper`, since this function should not be called by users (as it helps to extract the turnover masses from the inputs `lightcone` or `global_evolution`)

jordanflitter:
Thanks @steven-murray for your comments, I made some changes according to them. 

From what I was able to tell, I think you are satisfied with how we model the duty fraction for both ACGs and MCGs in the code, but you'd like to remove the constraint that `M_ACG > M_MCG` in order to make `f_duty^MCG` non-zero. I think I'm fine with that, but I do have to raise a concern that we already discussed. There are some functions (e.g. `EvaluateSFRD_Conditional_MINI`) that work with 2D interpolation tables since they evaluate quantities that depend on two local variables, `delta` and `log10Mturn_mcg`. For that reason, when we build the interpolation table, we assume that the ACG turnover mass is homogeneous and depends only on redshift (we also have now comments about this, e.g. in `initialise_SFRD_Conditional_table`). If we remove the `M_ACG > M_MCG` constraint, then it means that during reionization, the reionization feedback will be ignored for ACGs if we call `EvaluateSFRD_Conditional_MINI` and use interpolation tables! I think that the only way to overcome this is either to use a 3D interpolation table, or use an approximated analytic solution (which I'm not it exists), these are two solutions that are not ideal.

Plus, before I make any changes in the code's logic (and update the docs), we should see what @andreimesinger thinks, just to be sure that we are all on the same page.

steven-murray (approved):
woohoo, I like it!

jordanflitter:
Update for what was changed, after a discussion with @andreimesinger:

1. The MCG duty fraction was restored to be proportional to `exp(-M_h/M_atom)`, instead of `exp(-M_h/Mturn_ACG)`, that was a mistake that was incorporated in this PR and was now fixed to the previous form.
2. The zero MCG logic changed from `if Mturn_MCG > Mturn_ACG` to `if Mturn_ACG > M_atom`.
3. The enum `REIONIZATION_FEEDBACK_MODEL` was replaced with a boolean `USE_REIONIZATION_PHOTOHEATING_FEEDBACK`.
4. The metallicity is computed separately for each channel (ACG, MCG), rather than for their combined sum. This was a mistake in the previous code whose fix was necessary for the implementation of the above item 2 (in brief, the non-linear function of the metallicity + the said mistake prevented from separating the ACG and MCG contributions to the X-ray emissivity, thus the ACG turnover mass was assumed to be always homogeneous in the construction of the interpolation table for the total X-ray emissivity, otherwise the interpolation would have been 3D for the three fluctuating fields `delta`, `Mturn_ACG`, `Mturn_MCG`).
5. The `M_TURN.rst` in the docs was updated to reflect the above logics in v4.3 - except for the one mentioned in item 4, it will be described in `Radiation_Fields.rst` that is introduced in #755.

docs/M_TURN.rst
```diff
@@ -26,11 +26,7 @@ cutoff at the turnover mass,
...
 
 
-where :math:`H(x)` is the Heaviside step function. Meanwhile, the turnover mass was redshift-dependent
-and was defined as the halo mass that corresponds to a virial temperature of :math:`10^4` K, which is
-the threshold mass for atomic cooling, :math:`M_{\rm atom}(z)`. This mass scale was obtained by inverting
-Eq. 26 in Barkana & Loeb 2001 (https://arxiv.org/pdf/astro-ph/0010468) with :math:`T_{\rm vir}=10^4` K
-and a mean molecular weight of :math:`\mu=0.59`, corresponding to a fully ionized IGM. Approximately,
+where :math:`H(x)` is the Heaviside step function. As the time, the common name for the "turnover mass" was "minimum mass" or :math:`M_{\rm min}`. The user could define the turnover mass in terms of the halo mass or the corresponding virial temperature, with the latter implying a redshift dependence to the halo mass (c.f. Eq. 26 in Barkana & Loeb 2001 https://arxiv.org/pdf/astro-ph/0010468).  A common choice was a virial temperature of :math:`10^4` K, which is the threshold mass above which cooling through recombing hydrogen radiation is efficient (i.e. assuming ionized IGM with a mean molecular weight of :math:`\mu=0.59`). Approximately,
```
steven-murray: This should be split up into multiple lines. Also, typo in "recombing"

docs/M_TURN.rst
```diff
@@ -209,13 +204,13 @@ model with stochastic discrete halos was introduced. The source model that was u
 simulation was controlled by a new parameter called ``SOURCE_MODEL``. This parameter had several options:
 
 * ``"CONST-ION-EFF"``:
-  This option was equivalent to setting ``USE_MASS_DEPNDENT_ZETA`` on False
+  This option was equivalent to setting ``USE_MASS_DEPENDENT_ZETA`` on False
```
steven-murray: "on False" --> "to False"

docs/M_TURN.rst
```diff
@@ -209,13 +204,13 @@ model with stochastic discrete halos was introduced. The source model that was u
...
   in previous versions, namely the ionization efficiency was constant for all halo masses, the ACG
   turnover mass was set according to the ``M_MIN_in_Mass`` flag, and the ACG duty fraction function
   was a sharp cutoff at the ACG turnover mass.
 
 * ``"E-INTEGRAL"``:
-  This option was equivalent to setting ``USE_MASS_DEPNDENT_ZETA`` on True in
+  This option was equivalent to setting ``USE_MASS_DEPENDENT_ZETA`` on True in
```
steven-murray: "on True" --> "to True"

docs/M_TURN.rst
```diff
@@ -278,7 +273,12 @@ and ``"CHMF-SAMPLER"``, the logic for the turnover masses was as follows:
 
 In ``21cmFAST`` v4.3.0, the logic for the turnover masses was simplified and unified across most
 source models. This became possible by introducing two new parameters, ``V_CB_MODEL`` and
-``REIONIZATION_FEEDBACK_MODEL``. In addition, for better clarity, the free astrophysical
+``USE_REIONIZATION_PHOTOHEATING_FEEDBACK``. The latter was a flag that allowed to apply the inhomogeneous
+reionization feedback effect on the ACG turnover mass, regardless if MCGs were present or not in the
```
steven-murray: I think I would phrase this as "The latter was a flag that allowed applying inhomogeneous reionization feedback to all galaxies (both ACGs and MCGs). Note that this is implemented by modifying the ACG turnover mass inhomgeneously per-cell, which directly affects the ACGs, and indirectly affects the MCGs by extinguishing them since `M_TURN_ACG > M_ATOM`. Regardless, operationally it suppresses star formation for all species below the inhomgeneous photoheating mass."

docs/M_TURN.rst
```diff
@@ -384,14 +360,10 @@ same as in v4.0.0 (or v3.0.0):
...
-   avoid intensive computations for little reduction in fidelity.
+    f_{\rm duty}^{\rm (MCG)}(M_h;M_{\rm turn}^{\rm (MCG)},M_{\rm atom}) = \exp\left(-M_{\rm turn}^{\rm (MCG)}/M_h\right)\exp\left(-M_h/M_{\rm atom}\right)H(M_{\rm atom} - M_{\rm turn}^{\rm (ACG)}).
 
-2. The MCG **upper** turnover mass was defined to be the ACG **lower** turnover mass. This change
-   was made to ensure that there is no mass gap between MCGs and ACGs where galaxies cannot form.
+One subtle difference was made in the MCG duty fraction function compared to previous versions: MCGs can
+exist only if the ACG turnover mass is below the atomic cooling threshold mass,
+:math:`M_{\rm turn}^{\rm (ACG)} < M_{\rm atom}`, which is enforced by the Heaviside step
+function. This introduces a sharp cutoff in the MCG abundance (for a particular voxel) at this transition
+point, but the amplitude of this effect is small.
```
steven-murray: I think it would be best to add some physical insight here: the reason for this is that anything that raises M_TURN_ACG above M_ATOM is to do with some feedback suppression, which would also inhibit star formation in MCGs equally well. Therefore, given that there is very little star formation in MCGs above M_ATOM, this suppression beginning at even higher mass will result in *very* few MCGs below M_ATOM. This heaviside condition is a computationally efficient way of modeling this suppression.

steven-murray (commented):
Thanks @jordanflitter. I have a few comments about providing a bit more explanation, but otherwise I think it's good.

docs/M_TURN.rst
jordanflitter: I didn't write that part but I agree :) 
Done!

docs/M_TURN.rst
jordanflitter: You are right, I keep doing that. Done!

docs/M_TURN.rst
jordanflitter: Done!

docs/M_TURN.rst
jordanflitter: You are right, I wasn't careful in editing that part. I placed your suggested text.

docs/M_TURN.rst
jordanflitter: Yeah that sounds like a good explanation. I used your exact text there.

steven-murray (approved):

docs/M_TURN.rst
```diff
@@ -379,14 +360,10 @@ same as in v4.0.0 (or v3.0.0):
 
 .. math::
 
-    f_{\rm duty}^{\rm (MCG)}(M_h;M_{\rm turn}^{\rm (MCG)},M_{\rm atom}) = \exp\left(-M_{\rm turn}^{\rm (MCG)}/M_h\right)\exp\left(-M_h/M_{\rm turn}^{\rm (ACG)}\right)H(M_{\rm turn}^{\rm (ACG)} - M_{\rm turn}^{\rm (MCG)}).
-
-Two subtle differences were made in the MCG duty fraction function compared to previous versions:
-
-1. MCGs can exist only if :math:`M_{\rm turn}^{\rm (MCG)} < M_{\rm turn}^{\rm (ACG)}`, which is enforced
```
andreimesinger: I would change this to "if Mturn

docs/M_TURN.rst
andreimesinger: I would rephrase "ACG turnover mass is below the atomic cooling threshold mass" --> "ACG turnover mass is equal to the atomic cooling threshold (i.e. stellar and photo-heating feedback are inefficient)."   the ACG turnover mass by definition cannot be below the cooling threshold.

docs/updates_from_v3.rst
```diff
@@ -14,8 +14,9 @@ discrete source field which is then used in the spin temperature and ionization
 calculations. This not only includes the effects of stochasticity in the IGM observables,
 but also creates several new outputs which can be further used in forecasting galaxy
 survey, line intensity mapping, and cosmic background statistics. The sampler is
-activated with the flag ``HALO_STOCHASTICITY`` and serves as a faster replacement to the
-previous excursion-set halo finder, with greatly increased functionality. Halos are
+activated with setting ``SOURCE_MODEL`` to ``"CHMF-SAMPLER"`` and serves as a faster
```
andreimesinger: turning on the CHMF sampling does not replace DEXM.  It extends it to smaller scales so that one does not need to resolve the halos with the grid, thus allowing for an extended dynamical range.  The CHMF sampling still needs DEXM to get the largest halos above the grid scale.

src/py21cmfast/drivers/global_evolution.py
```diff
@@ -60,7 +60,20 @@ def compute_global_reionization_at_z(
...
         #       When that happens, note that the call below to evaluate_Nion_z calls run_global_evolution,
         #       so be careful to avoid infinite recursion!
-        nion, _ = evaluate_Nion_z(inputs=inputs, redshifts=np.asarray(redshift))
+        # TODO: we compute Nion below with no reionization feedback, since reionization feedback would make the ACG turnover mass
+        # to depend on the simulation's fields (specifically, on ionisation_rate_G12 and z_reion). If we hadn't done this, the ACG turnover mass will
+        # be evaluated from running a global evolution, causing an inifinite recursion! This might be fixed by passing the "correct" turnover mass
+        # to evaluate_Nion_z, but I am not sure it worths the effort, since this tricky situation happens only when the user runs a global evolution
+        # with USE_TS_FLUCT=False, which is kinda pointless because it would yield the wrong global 21-cm signal. Having said that, it should be noted
+        # that also with USE_TS_FLUCT=True, the ACG turnover mass is currently ALWAYS evaluated in SpinTemperatureBox.c with no reionization feedback,
```
andreimesinger, line 68: but this ACG turnover mass is only for Ts, which is usually in the regime of Ts >> Tcmb by the time photoheating feedback has a significant impact so it doesn't make much of a difference on the results.  Your comment makes it sound like photoheating feedback cannot be used if Ts is being calculated

src/py21cmfast/drivers/single_field.py
```diff
@@ -587,8 +587,20 @@ def compute_xray_source_field(
...
+
+                mturn_MCG = cfuncs.get_molecular_cooling_threshold_with_feedbacks(
+                    inputs=inputs,
+                    redshifts=zpp_avg[i],
+                    J_LW_21=0.0,
+                    v_cb=inputs.cosmo_tables.V_CB_AVG,
+                )
                 box.filtered_sfr_mini.value[i] = 0
-                box.mean_log10_Mcrit_LW.value[i] = inputs.astro_params.M_TURN  # minimum
+                box.mean_log10_Mcrit_LW.value[i] = np.log10(
```
andreimesinger, line 601: this variable name is somewhat misleading because it is a general one not Mcrit LW. but not a big deal i guess...

andreimesinger (approved):

docs/M_TURN.rst
jordanflitter: Thanks, that's a good comment! I negated the sentence so now it says "MCGs cannot
exist if the ACG turnover mass is above the atomic cooling threshold mass,
(i.e. if :math:`M_{\rm turn}^{\rm (ACG)} > M_{\rm atom}`), which is enforced by the Heaviside step
function."

docs/M_TURN.rst
jordanflitter: I think this was also addressed with the previous comment.

docs/updates_from_v3.rst
jordanflitter: Yes, that's correct. I changed it to: "The sampler is
activated with setting ``SOURCE_MODEL`` to ``"CHMF-SAMPLER"`` and serves as a complementary
algorithm to the excursion-set halo finder (which can still be usable without the sampler by
setting ``SOURCE_MODEL`` to ``"DEXM-ESF"``), with greatly increased functionality."

src/py21cmfast/drivers/single_field.py
jordanflitter: I agree. We leave these field names at the moment in order not to break API, but we are planning to change them soon in a future PR.

src/py21cmfast/drivers/global_evolution.py
jordanflitter: Note that this comment is found under the `if spin_temp is None` clause, which happens only if `USE_TS_FLUCT=False` and the spin temperature is not calculated. I give a bit more details for context below.

When we do `run_global_evolution`, we do a quick "single-zone" simulation, just for evaluating the global signal quickly. Since it's a single zone, we don't have ionized bubbles, and we need to estimate the neutral fraction, or volume filling factor, in an alternative method. If `USE_TS_FLUCT=True`, we take it from `spin_temp.Q_HI`, otherwise we need to compute it similarly to what the C code does when `USE_TS_FLUCT=True`: this is done by calling `evaluate_Nion_z`, but that function internally calls `run_global_evolution` if `USE_REIONIZATION_PHOTOHEATING_FEEDBACK=True`, so in order to avoid infinite recursion we just switch the reionization feedback off for this purpose.

PR #740, opened 2026-07-07
Use hmf's Yung24 fitting function directly in HMF test

### LINKED ISSUE #710: Update Yung+24 test to use hmf directly
Currently `hmf` does not have the Yung+24 function to test against, but it is being merged in https://github.com/halomod/hmf/pull/296. When that is merged, we should update the test that matches this function.

steven-murray:
## Summary
- hmf v3.6.2 added the Yung+24 halo mass function fit (previously only available on our end via hand-copied coefficients). This replaces the duplicated `YUNG24_PHYSICAL_PARAMS` dict and inlined polynomial formula in `test_new_hmf_matches_reference` with a direct call to `hmf.mass_function.fitting_functions.Yung24`, matching the approach already used for REED07.
- Bumps the `hmf` dependency to `>=3.6.2` in `setup.py`, since that's the first version with the Yung24 fit.

Fixes #710

## Test plan
- [x] `pytest tests/test_cfuncs.py -k test_new_hmf_matches_reference` passes for both `REED07` and `YUNG24` branches (`EH`/`BBKS` power spectra).
- [x] Full `pytest tests/test_cfuncs.py` passes (84 passed, 4 pre-existing xfails).
- [x] `ruff check`/pre-commit hooks pass.

## Summary by Sourcery

Replace hand-coded Yung24 halo mass function implementation in tests with direct use of hmf's Yung24 fitting function and bump the hmf dependency version accordingly.

Enhancements:
- Use hmf.mass_function.fitting_functions.Yung24 directly in the halo mass function reference test instead of duplicated polynomial parameters.

Build:
- Increase required hmf version to >=3.6.2 in setup.py to ensure availability of the Yung24 fitting function.


steven-murray:
Addressed the duplicated `hmf>=3.6.2` requirement — factored it into a single `HMF_REQ` constant used by both `test_req` and `install_requires` in `setup.py` (e33e4fc4).

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

steven-murray:
Re the Codecov \"JUnit XML file not found\" warning: this is a downstream symptom, not something introduced by this PR. The \"Run Unit Tests\" CI step is failing, which causes \"Run Integration Tests\" (where `junit.xml` is generated) to be skipped.

I compared the failing tests on this branch against the tip of `main` (3c61b224) and the failure set is identical — same 34 tests, same `astropy.units.errors.UnitConversionError` ('m / s' vs dimensionless). This is a pre-existing, repo-wide CI breakage unrelated to the files this PR touches (`setup.py`, `tests/test_cfuncs.py`); none of the failures are in the Yung24/hmf test changes here. Leaving it out of scope for this PR — it looks like a separate astropy-related regression across several unrelated test files (`test_h5.py`, `test_classy_interface.py`, `test_global_initialization.py`, `test_rsds.py`, etc.).

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

tests/test_cfuncs.py
```diff
@@ -372,31 +358,13 @@ def ps_difference(growth):
...
-        c_z = (
-            YUNG24_PHYSICAL_PARAMS["c_0"]
-            + YUNG24_PHYSICAL_PARAMS["c_1"] * z
-            + YUNG24_PHYSICAL_PARAMS["c_2"] * z**2
-        )
-        A_z = (
-            YUNG24_PHYSICAL_PARAMS["A_0"]
-            + YUNG24_PHYSICAL_PARAMS["A_1"] * z
-            + YUNG24_PHYSICAL_PARAMS["A_2"] * z**2
+        yung24 = Yung24(
```
jordanflitter: Not a critical comment, but is there a way to unify this with the `REED07` branch in `test_new_hmf_matches_reference`? That test still has two branches, which seems a bit odd considering that the `hmf` packages is now used for both `REED07` and `YUNG24`.

jordanflitter (approved):
Thanks @steven-murray, I made a minor comment.

The tests are failing because of astropy v8. I fixed that in #737 by modifying line 288 in `classy_interfrace.py`, see https://github.com/21cmfast/21cmFAST/pull/737/changes#diff-010c965001daef1fe0c45b5b8e24e3c87b2c1168c638a8846c97948d9a0e9765.

Also, shouldn't this branch be rebased on top of `release-v4.3`? I think it should.

I approved it anyways :)

tests/test_cfuncs.py
steven-murray: Good call — unified both branches into a single `MassFunction`-based comparison (matched transfer function, growth model, and cosmology), since Yung24 now supports the same `units="physical"` hmf_params approach as Reed07. Removed the separate PS-growth-calibration codepath and the now-unused `Yung24`/`optimize` imports (7538c0b1).

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

steven-murray:
@jordanflitter no I think this should go in to main -- it's just an update of a test.

Issue #668, opened 2026-04-14
Filter emissivity fields in all `SOURCE_MODEL` options


jordanflitter:
Today we have 5 options for `SOURCE_MODEL`. There is some inconsistency in the way we compute the radiation fields (X-ray heating rate, ionization rate, Lyman alpha and Lyman-Werner fluxes) between the different source models in the spin temperature module.

- In the old v3 source models (`CONST-ION-EFF` and `E-INTEGRAL`), we filter the Eulerian over-density field with a top-hat filter whose filter radius corresponds to the integrated shell radius. Given the filtered density field, we compute the local emissivity fields by integrating over the CHMF, and upon trapezoidal integration over the shell radii we end up with the radiation fields.
- In the new v4 source models (`L-INTEGRAL`, `DEXM-ESF` and `CHMF-SAMPLER`) we compute the local emissivity fields given the local, unfiltered Lagrangian density field + contributions from the discrete halo catalog if there is one. We then filter the emissivity fields, and upon trapezoidal integration over the shell radii we end up with the radiation fields.

We should use the v4 formalism also in the old v3 models. In my opinion, `E-INTEGRAL` should work very similarly to `L-INTEGRAL`, except that we use the density field on the Eulerian grid (and a top-hat filter in the reionization module, instead of the exponential filter. This should probably go into a different issue...).

There are many upsides to do that:

-  **Better modeling of the radiation fields**. From a quick chat that I had with @andreimesinger, old versions of 21cmFAST didn't have the emissivity fields on a grid, only the density field. Now that we also have the gridded emissivity fields (e.g. the fields of `HaloBox`), we should definitely use them, also for the older source models.
- **Removes limitations**. To relate with the previous point, we currently limit ourselves in using `LYA_MULTIPLE_SCATTERING=True` only in the new v4 source models. Having a unified formalism will allow us to relax this limitation and use `LYA_MULTIPLE_SCATTERING=True` also in the old source models. I think we have even more similar logics that constrain us to use either the old v3 or the new v4 source models in some scenarios, and these limitations could be removed as well once the unified formalism is implemented.
- **Much cleaner, more compact and organized code**. The current situation is that there is some duplication in the code, there are MANY if statements that go like `if SOURCE_MODEL < 2 do something, else, do something else`. Working with the same unified formalism will remove these ugly and lengthy if statements.
[...]

Issue #659, opened 2026-04-02
Improve precision of `tau_X`, mostly after reionization has started


jordanflitter:
This sub-issue follows #122. The idea in that issue is to use the information from `IonizedBox` at higher redshifts in order to evaluate `Q_HI`. 

In the easier implementation, only the **global** `neutral_fraction` from the higher redshifts is saved, such that `Q_HI = <x_HI>`, for each cell in the box. In the more difficult implementation, `Q_HI=x_HI`, namely the volume filling factor **varies spatially**. 

Note that `Q_HI` is evaluated at `z_hat`, which is an intermediate redshift between `zp` (the redshift of the snapshot) and `zpp` (the dummy redshift that corresponds to the integrated shell radius, from which the X-ray photon was emitted), and therefore requires not only to save the last `neutral_fraction` box, but several of those and interpolate the field at `z_hat`. 

In my opinion, taking the global `neutral_fraction` would be the better approach, not only because it is easier (and requires less memory), but also because it is somewhat more consistent: the X-ray photon traverse across the box, so it doesn't "see" the same cell at each `z_hat`. It is well understood that 21cmFAST is a FAST semi-numerical code, not RT, and as such, it uses the global volume filling factor in the evaluation of `tau_X`, as stated above Eq. 16 and in footnote 22 in [Mesinger et al. 2010](https://arxiv.org/pdf/1003.3878).

PR #749, opened 2026-07-23
Removal of 4D arrays in `XraySourceBox`

### LINKED ISSUE #552: Can we get rid of the 4D arrays in XraySourceBox?
## The problem

Each `XraySourceBox` contains now up to three 4D arrays (`filtered_xray`, `filtered_sfr` and `filtered_sfr_mini`) which take a lot of memory (in the case of Ly $\alpha$ multiple scattering, we have two more of these, `filtered_sfr_lw` and `filtered_sfr_mini_lw` to distinguish the straight line propagation of LW photons compared to the scattered trajectory of Ly $\alpha$ photons). Memorywise, I think this implementation could be improved.

## Proposed Solution (Ly $\alpha$ flux)

Let us look for example on how `filtered_sfr` contributes to `sfr_term` [here](https://github.com/21cmfast/21cmFAST/blob/main/src/py21cmfast/src/SpinTemperatureBox.c#L1567) and how `sfr_term` eventually is used to compute $J_\alpha$. In that line, we have

$$
\text{sfr term}=\dot\rho_\*^{\mathrm{(II)},R}\left(\mathbf x,z''\right)\times\Delta z''\left|\frac{dt''}{dz''}\right|
$$

while the Ly $\alpha$ flux in the code is eventually given by

$$
J_{\alpha,\*}\left(\mathbf x,z'\right)=\sum_{i=\mathrm{II,III}}\int_{z'}^{z_\mathrm{max}}\dot\rho_\*^{i,R}\left(\mathbf x,z''\right)\Delta z''\left|\frac{dt''}{dz''}\right|\left(1+z'\right)^2\left(1+z''\right)
$$

$$
\times\sum_{n=2}^{23}f_\mathrm{recycle}\left(n\right)N_\*^i\left(\nu''_n\right)I^i\left(\nu''_n\right)\Theta\left(z{max}\left(z',n\right)-z''\right)\frac{c}{4\pi}
$$

Here, the integral is computed numerically via the trapezoidal integration rule, and can be thought as a weighted sum of the filtered SFRD $\dot\rho_\*^{i,R}$. Instead of saving the entire 4D array of $\dot\rho_\*^{i,R}\left(\mathbf x,z''\right)$, we could write $J_{\alpha,\*}$  as

$$
J_{\alpha,\*}\left(\mathbf x,z'\right)=\sum_{i=\mathrm{II,III}}\sum_{R=R_\mathrm{min}}^{R_\mathrm{max}}\dot\rho_\*^{i,R}\left(\mathbf x,z''\right)\times w_i\left(z',R\right) = \sum_{R=R_\mathrm{min}}^{R_\mathrm{max}}\sum_{i=\mathrm{II,III}}\dot\rho_\*^{i,R}\left(\mathbf x,z''\right)\times w_i\left(z',R\right)
$$

where $w_i\left(z',R\right)$ is a "global" quantity (doesn't depend on the cell $x$) that depends only on the the stellar population $i$, the snapshot redshift $z'$ and the filter radius $R$ (note that the dummy redshift $z''$ is determined from $z'$ and $R$). From here, it is easy to see that we don't really need to save the entire 4D array of $\dot\rho_\*^{i,R}\left(\mathbf x,z''\right)$ (`filtered_sfr` in the code) but only save an "evolving" 3D box that evolves when we account higher filter radii. At the final step, the evolved 3D box is $J_{\alpha,\*}\left(\mathbf x,z'\right)$.

## Proposed Solution (X-ray flux)

The expression for X-ray flux is more complicated as it depends also on frequency and thus the quantities of interest involve integrals over frequency. Here, `filtered_xray` contributes to `xray_sfr` [here](https://github.com/21cmfast/21cmFAST/blob/main/src/py21cmfast/src/SpinTemperatureBox.c#L1570). This line reads (note that $L_X$, which is `halo_xray` in the code, contains both contributions from popII and popIII stars)

$$
\text{xray sfr}=\frac{10^{-38}L_X^{R}\left(\mathbf x,z''\right)}{V_\mathrm{cell}}\times\Delta z''\left|\frac{dt''}{dz''}\right|\times\left(1+z''\right)^{-\alpha_X}\times10^{38}
$$

This quantity is then used to compute more physical quantities that involve integration over frequency. For example, the X-ray heating is then given by

$$
\Gamma_X\left(\mathbf x,z'\right)=\int_{z'}^{z_\mathrm{max}}\left(1+z''\right)^{-\alpha_X}\frac{L_X^{R}\left(\mathbf x,z''\right)}{V_\mathrm{cell}}\Delta z''\left|\frac{dt''}{dz''}\right|
$$

$$
\times \int_{\max{[\nu_0,\nu_{\tau_X=1}}]}^{\nu_\mathrm{max}} d\nu\sum_i f_\mathrm{heat}\left(x_\mathrm{e}\left(\mathbf x,z'\right),\nu-\nu^\mathrm{th}\right)h_\mathrm{P}\left(\nu-\nu^\mathrm{th}_i\right)f_ix_i\left(\mathbf x,z'\right)\sigma_i\left(\nu\right)\left(\frac{\nu}{\nu_0}\right)^{-\alpha_X-1}
$$

$$
\times \frac{\left[\int_{\nu_0}^{\nu_\mathrm{max}} d\nu\left(\frac{\nu}{\nu_0}\right)^{-\alpha_X}\right]^{-1}}{h_\mathrm{P}\,\nu_0}c\left(1+z'\right)^{\alpha_X+3}
$$

Here, the frequency integral is computed via the trapezoidal integration rule. Similarly as above, we may write this expression as 

$$
\Gamma_X\left(\mathbf x,z'\right)= \sum_{R=R_\mathrm{min}}^{R_\mathrm{max}}\sum_{\nu=\nu_\mathrm{min}}^{\nu_\mathrm{max}}L_X^{R}\left(\mathbf x,z''\right)\times w\left(\mathbf x,z',R,\nu\right)=\sum_{R=R_\mathrm{min}}^{R_\mathrm{max}}L_X^{R}\left(\mathbf x,z''\right)\times\tilde w\left(\mathbf x,z',R\right)
$$

where $\tilde w\left(\mathbf x,z',R\right)$ is just as a summed version of $w\left(\mathbf x,z',R,\nu\right)$. Note that we cannot switch the order between the $\nu$ summation and the $R$ summation, since the former depends on the latter through the minimum frequency we consider: $\nu_\mathrm{min}=\max{[\nu_0,\nu_{\tau_X=1}}]$ and $\tau_X$ depends on $R$. Regardless, here we ended with a local weight (that depends on the cell $\mathbf x$). But still, we see that we don't need to save the entire 4D array $L_X^{R}\left(\mathbf x,z''\right)$ (`filtered_xray` in the code), but rather the "evolving" 3D box of it, just like in the case of the Ly $\alpha$ flux.

## Conclusions

I think it is indeed possible to come up with an implementation that doesn't require saving 4D arrays, only 3D boxes, thus reducing the amount of used memory. I also believe that this proposed implementation will not harm the runtime of the calculation, and I also think it could be done at the python wrapper for that purpose (instead the C backend), without slowing down things too much, as all we need to do is basically sum and multiply 3D boxes. I imagine replacing the `XraySourceBox` with something like `RadiationFields` that includes 3D boxes of physical radiation fields like $J_{\alpha,*}$, $J_\mathrm{LW}$, $J_{\alpha,X}$, $\Gamma_X$ and $\Lambda_X$ (the last one is X-ray ionization rate). Then, the computation of the spin temperature becomes much simplified as it would only require promoting $T_k$ and $x_e$ to the next redshift iteration and use the promoted values together with $J_\alpha$ to compute the spin temperature. This kind of implementation could also be generalized for `USE_HALO_FIELD=False` (though I need to give it more thought), which would make `ComputeTsBox` much simplified.

This is a rather big change because it requires modifying both wrapper and backend. Before making a PR for it, I'd like to hear @daviesje and @steven-murray thoughts on it.

jordanflitter:
Fix #552.

This PR introduces major refactoring for improving memory optimization and modularity of the code. It does the following:

1. Refactored `ComputeTsBox` in `SpinTemperatureBox.c` into new functions called `setup_radiation_fields`, `accumulate_radiation_shell` and `multiply_radiation_fields_by_constants`.
2. Moved the new functions from the previous item + `one_annular_filter` and `UpdateRadiationFields` to a new module called `RadiationFields.c`. This leaves `SpinTemperatureBox.c` pretty compact, as it now only evaluates the spin temperature and evolves the thermal fields (`T_k`, `x_e`) to the next redshift snapshot. Thus, the heavy calculations for computing the radiation fields (X-ray heating rate, photoionization rate from X-rays, Lyman alpha and Lyman Werner fluxes) are all found in the new module `RadiationFields.c`.
3. Renamed `XraySourceBox` to `RadiationFields`, `compute_xray_source_field` to `compute_radiation_fields`, etc.
4. **Main change**: removal of the 4D arrays from `RadiationFields` (formerly known as `XraySourceBox`). This is done by introducing new 3D fields in `RadiationFields` that have physical meaning (`xray_heating_rate`, `xray_ionization_rate`, `lya_flux_continuum_injected`, `lyw_flux`, and more). The idea is the following: instead of saving a-priori 4D arrays of the filtered emissivity fields for all filter radii, then for a given snapshot, the radiation fields evolve by integrating over the shells via the trapezoidal rule, so we only need the relevant filtered emissivity fields which are 3D arrays. As before, once all shells have been integrated, the radiation fields are multiplied by constants - this makes them physically meaningful quantities. It's important to stress that while `RadiationFields` (formerly known as `XraySourceBox`) now contains new 3D arrays, the allocations of these 3D arrays replaces the allocations of the 3D arrays that were used to be done internally in the C code, so the consumed memory is definitely smaller now.

Apart from this:
1. A bug in `ComputeTsBox` was fixed - number of threads was not specified within a pragma loop.
2. Another memory optimization was made. Until now, when `USE_LYA_HEATING=True` (which is the default), three arrays were allocated. While these arrays had different names and meaning, they corresponded to `lya_flux_continuum`, `lya_flux_injected` and `lya_flux_continuum_injected`. The first two are used for the effect of lya heating, since the contributions from continuum and injected photons to the effect are different, and the third one is used for the WF coupling. However, this is wasteful memorywise since the third one is exactly the sum of the first two. Thus, we allocate `lya_flux_continuum_injected` only when `USE_LYA_HEATING=False`, while we allocate `lya_flux_continuum` and `lya_flux_injected` when `USE_LYA_HEATING=True`, which allows us to get for "free" `lya_flux_continuum_injected` by combining the two allocated arrays of continuum and injected photons.

Several notes about the implementation:
1. On the python side, the new implementation for `compute_radiation_fields` (formerly known as `compute_xray_source_field`) is quite hacky. As the comment inside the function says, fixing #659 and #668 could make things a lot better, since we would not need to compute a-priori `log10_Mcrit_MCG_ave` before going into the C code. This would still leave some "hackiness" in the implementation, since we currently don't have `RadiationFieldsSetup` (new C struct from this PR) as an `OutputStructZ` on the python level, this is something I certainly suggest to do, but maybe in a future PR.
2. All tests pass without modifying the test data files. This is important because the aim for this PR is to compute the exact same numbers, but with less memory.
3. Unlike what I said in #552, I wasn't able to remove the `allow_already_computed` hack in `_compute()`. This is because all the fields in `RadiationFields` are evolved as the shells are integrated (per snapshot). The only way to change that would be to introduce a new `RadiationFields` instance so we would have two instances, i.e. `prev` and `current`, similarly to how we work in `_redshift_loop_generator`. This however would increase the used memory, so I figured it would be better to keep the `allow_already_computed` hack.
4. As was mentioned in #668, unifying our filtering scheme for all source models would make the code MUCH simpler and more modular. While doing that will change the output of the code for the old Eulerian source models, I suggest we do that within the v4.3 project.

## Summary by Sourcery

Replace the four-dimensional X-ray source representation with memory-efficient, shell-integrated radiation fields while modularizing their computation and preserving spin-temperature results.

New Features:
- Expose `RadiationFields` and `compute_radiation_fields` as the replacement API for computing X-ray, Lyman-alpha, and Lyman-Werner radiation fields.

Bug Fixes:
- Specify the configured OpenMP thread count when computing spin-temperature fields.

Enhancements:
- Replace per-shell four-dimensional emissivity storage with shell-integrated, physically meaningful three-dimensional radiation fields to reduce memory usage.
- Separate radiation-field computation into a dedicated module and simplify spin-temperature evolution around its computed inputs.
- Avoid redundant Lyman-alpha storage by allocating either combined flux or separate continuum and injected fluxes according to the heating configuration.
- Use named constants for filtering modes and update source-model, caching, and driver interfaces to the new radiation-field representation.

Documentation:
- Update the public API and related references from `XraySourceBox` to `RadiationFields`.

Tests:
- Update radiation-field, output-structure, input-validation, and caching tests for the renamed API and reduced-storage representation.


src/py21cmfast/drivers/single_field.py
```diff
@@ -379,10 +385,13 @@ def compute_halo_grid(
...
 
 # TODO: make this more general and probably combine with the lightcone interp function
+# TODO: remove the need_c argument, this is currently required because we call this function once for just computing the history of
+# log10_Mcrit_MCG_ave - see other comment about this in need_c.
 def interp_halo_boxes(
     halo_boxes: list[HaloBox],
-    fields: list[str],
+    interp_fields: list[str],
     redshift: float,
+    need_c: bool,
```
steven-murray, line 394: If this parameter is _known_ to be going to change soon, let's flag it as private  (i.e. call it `_need_c`)

steven-murray (approved):
Single comment but otherwise pre-approved. This looks really great. Looking forward to getting those other PRs in to really improve this modularity!

Issue #715, opened 2026-06-03
Testing: add flags that fail when warnings are raised


steven-murray:
The justification for doing this is to catch potential breaking changes coming in dependencies sooner (before they actually break things).

Another motivation is that we get less noise in the CI test logs.

jordanflitter:
Yes, I like this idea. The tests don't necessarily have to fail if a warning is raised. For example, if the warning is expected, the tests could still pass `with pytest.warns`

PR #754, opened 2026-08-08
Add pytest warning filters to catch deprecation warnings and fix misclassified DeprecationWarning

hassamrajpoot:
Implements #715.

**Approach**

Rather than manually auditing warnings, I added the filterwarnings block to setup.cfg first and let the suite surface exactly which tests needed attention. This reduced 718 warnings in the CI log down to 5 specific failing tests.

ini
filterwarnings =
    error::DeprecationWarning
    error::PendingDeprecationWarning
    error::FutureWarning

**UserWarning audit**

I audited all UserWarnings in the suite comprehensively to decide whether to add ignore::UserWarning. Every UserWarning in the codebase is intentional: physics domain advisories, operational notifications, and test comparison reporting. I attempted ignore::UserWarning but it broke test_warn_formatting because UserWarnings are part of 21cmFAST's user-facing CLI interface. Conclusion: UserWarning cannot be globally suppressed.

**Bonus finding: misclassified DeprecationWarning**

During the audit, I found with_equal_cdist_slices in lightconers.py raising UserWarning for a deprecation message. Root cause: the warning category argument was omitted from warnings.warn(), causing Python to default to UserWarning. Fixed by adding DeprecationWarning as the second argument. Without this fix, the new filter would not catch this deprecation.

**The 5 failing tests**

All 5 were tests deliberately exercising deprecated backwards-compatible parameters. Each needed a pytest.warns(DeprecationWarning) wrapper:

- test_inhomo_reco_false_sets_none: split into two tests; deprecated variant wrapped
- test_use_relative_velocities_false_sets_none: same pattern, split and wrapped
- test_fix_vcb_avg_conflict: pytest.warns combined with existing pytest.raises
- test_compatability_with_database: old database file triggers deprecation warnings when deserializing old parameter names; wrapped with pytest.warns(DeprecationWarning, match="is deprecated")
- Three with_equal_cdist_slices calls in test_drivers_lc.py: wrapped after fixing the source classification

**Test results**

192 tests across all affected files: 192 passed, 0 failed. Full suite: 968 passed, 1 failed (pre-existing test_hyper_2F3[0.0], unrelated to this change).

**Open questions for reviewers**

A few decisions I made that I'd welcome feedback on:

- ignore::UserWarning: I initially added this to reduce CI log noise (one of the stated motivations in #715) but it broke test_warn_formatting since UserWarnings are part of the CLI interface. I removed it. If there's a more targeted way to reduce UserWarning noise in CI logs without suppressing them globally, I'm open to implementing it.
- test_compatability_with_database: I used pytest.warns(DeprecationWarning, match="is deprecated") rather than matching the exact parameter names, since the old database file triggers two different deprecation warnings (USE_RELATIVE_VELOCITIES and INHOMO_RECO). Happy to make this more explicit if preferred.
- FutureWarning: I included this alongside DeprecationWarning and PendingDeprecationWarning since NumPy and other scientific libraries use it for upcoming breaking changes. Let me know if this should be scoped differently.

## Summary by Sourcery

Enforce strict handling of deprecation-related warnings and correctly classify the deprecated lightconer API warning.

Bug Fixes:
- Correct the classification of the `with_equal_cdist_slices` deprecation warning so it is reported as a deprecation rather than a user warning.

Enhancements:
- Add regression coverage for the deprecated lightconer API, including verification that it will eventually be removed.

Build:
- Configure pytest warning filters to treat deprecation, pending deprecation, and future warnings as errors.

Tests:
- Update affected tests to explicitly expect warnings from deprecated compatibility paths.

### DIFF
--- a/src/py21cmfast/lightconers.py
+++ b/src/py21cmfast/lightconers.py
@@ -8,6 +8,7 @@ from collections.abc import Sequence
 from functools import cached_property, partial
 
 import attrs
+import deprecation
 import numpy as np
 from astropy import units
 from astropy.cosmology import FLRW, z_at_value
@@ -147,8 +148,15 @@ class Lightconer(ABC):
         slices in comoving distance.
         """
         warnings.warn(
-            "with_equal_cdist_slices is deprecated and will be removed in future versions. "
-            "Call between_redshifts instead to silence this warning.",
+            deprecation.DeprecatedWarning(
+                "with_equal_cdist_slices",
+                deprecated_in="4.2.0",
+                removed_in="5.0.0",
+                details=(
+                    "with_equal_cdist_slices is deprecated and will be removed in a future version. "
+                    "Call between_redshifts instead to silence this warning."
+                ),
+            ),
             stacklevel=2,
         )
         return cls.between_redshifts(
--- a/tests/test_lightconer.py
+++ b/tests/test_lightconer.py
@@ -3,6 +3,7 @@
 import re
 from dataclasses import dataclass
 
+import deprecation
 import numpy as np
 import pytest
 from astropy import units as un
@@ -166,6 +167,28 @@ def test_equal_cdist_endpoint():
         resolution=res,
     )
     assert np.isclose(lc.lc_distances.max(), d1, atol=d1 / 20)
+
+
+def test_with_equal_cdist_slices_deprecated_warning():
+    """Test that with_equal_cdist_slices raises a deprecation warning."""
+    with pytest.warns(
+        deprecation.DeprecatedWarning, match="with_equal_cdist_slices is deprecated"
+    ):
+        lcn.RectilinearLightconer.with_equal_cdist_slices(
+            min_redshift=6.0,
+            max_redshift=7.0,
+            resolution=2 * un.Mpc,
+        )
+
+
+@deprecation.fail_if_not_removed
+def test_with_equal_cdist_slices_is_removed():
+    """Fails when removed_in version is reached, reminding you to delete with_equal_cdist_slices."""
+    lcn.RectilinearLightconer.with_equal_cdist_slices(
+        min_redshift=6.0,
+        max_redshift=7.0,
+        resolution=2 * un.Mpc,
+    )
 
 
 def test_angular_lightconer_bad_instantiation():


hassamrajpoot:
Fixed the CI failure on macOS , the filterwarnings were catching pytest's 
own internal PytestRemovedIn10Warning in test_perturb.py, causing 5 test 
errors. Scoped the filters to py21cmfast specifically as Sourcery suggested, 
which catches deprecations from our own code only and leaves third-party 
warnings alone.

Tradeoff: this means cases where py21cmfast calls a deprecated dependency 
API won't be caught (the warning would come from the dependency's code, 
not py21cmfast's). Happy to use a targeted ignore list instead if preferred.

jordanflitter:
Hi @hassamrajpoot, thanks for contributing to the code!

I went over your changes and I have some comments, but mostly questions:

1. I’m failing to understand the overall goal of this PR. If I look at the CI log, I still see at the bottom of “Run Unit Tests” the following message: "922 passed, 26 skipped, 30 xfailed, 588 warnings”. Meanwhile, I see in #748 the following message: "922 passed, 26 skipped, 30 xfailed, 599 warnings”. So I’m a bit confused about the numbers cited in the description. Is this PR intended only for the three modified test modules? I might be missing something fundamental here.
2. Note that #715 is about mapping ALL the warnings in the test suite, such that no unexpected warning is raised during the test (which would also silent the warning notification in the log as far as I understand). I think this is beyond the scope of this PR.
3. We have been lately using `deprecation.DeprecatedWarning`, which I think is more informative than `DeprecationWarning` as it allows to date the versions where the deprecation warning was introduced and is expected to be removed. This is also helpful because we can test that the deprecation warnings are removed when they should be. What is the advantage of using `DeprecationWarning` instead?
4. I don’t understand the new `filterwarnings` line in `setup.cfg` (and whether it should be defined in that module). Could you please elaborate on that?
5. Please address sourcery's comment (match="is deprecated" filter is quite broad). Why was this change implemented in `test_global_evolution::test_compatability_with_database`, and only there?
6. Since we are having more and more deprecation warning (the code evolves together with the names in it, and we want to keep the same API until v5 is out), I think that it would be better to test and manage all the deprecation warnings in a dedicated test module, something like `test_deprecated.py`, rather than having these tests spread in other test modules. What do you think @steven-murray?
7. This is just my quick list of comments/questions, but I genuinely think it would be better for @steven-murray to approve these changes, as he’s our python wizard.

hassamrajpoot:
Hi @jordanflitter, thanks for the detailed review.

I want to be upfront first:  I interpreted it as "make deprecation warnings fail CI" rather than "map all warnings so none appear unexpectedly." 

My reasoning was based on steven-murray's opening comment in #715: "The justification for doing this is to catch potential breaking changes coming in dependencies sooner (before they actually break things)." That framing led me toward catching deprecation warnings as errors , so that if a dependency deprecates an API we're using, CI fails immediately rather than the warning accumulating silently. I wasn't thinking about silencing all warnings from the log, I was thinking about catching future breakage early.

I understand now that the full vision is broader 

1. On points 1 and 2: Correct  , the 588 remaining warnings confirm this PR only partially addresses #715. The filterwarnings approach catches deprecations from py21cmfast code but doesn't map all warnings or silence log noise.

2. On point 3: I wasn't aware the team uses deprecation.DeprecatedWarning. To answer your question directly : I used DeprecationWarning simply because it's Python's built-in standard and I wasn't aware of the team's convention. deprecation. DeprecatedWarning is clearly better here: it carries version information, allows testing that warnings are removed when they should be, and is consistent with the rest of the codebase. There's no advantage to using the bare DeprecationWarning in this context. I'll update the wrappers accordingly.
 
3. On point 4: The filterwarnings block tells pytest to treat those warning categories as errors rather than just printing them. Given your broader scope clarification, this may be the wrong mechanism entirely.

4. On point 5: The broad match="is deprecated" was deliberate because the legacy file triggers two deprecation warnings (USE_RELATIVE_VELOCITIES and INHOMO_RECO). I flagged this as an open question in the PR description. Happy to make it more explicit.

5. On point 6: test_deprecated.py makes a lot of sense given deprecations will keep accumulating until v5. I'd actually be happy to take on the full scope of #715 if that's useful , mapping all warnings across the suite and consolidating deprecation tests into test_deprecated.py. That's a bigger piece of work but I'm willing to do it properly if it's the direction you want to go.
 

Proposed path forward: A few options as I see it:

- I take on the full scope of #715 , map all warnings, consolidate into test_deprecated.py, drop the filterwarnings block or rethink it
- Narrow this PR to just the lightconers.py fix (genuinely self-contained) and open a fresh PR for the full #715 work
- Close this entirely and start fresh with the correct scope

Happy to follow whatever direction you and @steven-murray think is best.

jordanflitter:
Okay, thanks for the reply, now everything makes more sense :)

I think that #715 indeed meant to clean (eventually) all the warning messages from the log by using `pytest.warns`, that was at least my interpretation, but I could have been wrong. It would be best to let @steven-murray clarify what he meant :)

Based on your reply, I think we are more or less on the same page for the long-term. There should be a dedicated `test_deprecated.py` module that tracks all the deprecation warnings, while in the source code we use `deprecation.DeprecatedWarning` for more descriptive warning messages and better testing (e.g. the deprecation warning is gone once the code's version is high enough). Meanwhile, in the other test modules, we use `pytest.warns` for (1) clear the log from expected warnings and (2) make sure that all the warnings that are triggered in the tests are expected, as unexpected warnings would cause a failure (this will be super helpful!).

So that's the planned roadmap for handling the warnings in the test-suite, which I think you agree with me, but let me know if you have any objections. Like I said, I would really love to hear @steven-murray's input on this, as very often he has good advice from his perspective. If he's cool with the above overall plan, we should update #715 with more details of the agreed plan.

As for what you should do next, provided we are all agree on the above long-term plan, we would definitely appreciate any help in achieving that goal, even partially. So I leave it up to you how much effort you would like to spend on that, as long as you stick with the agreed plan :)

One more thing that I should mention. We have an old generic issue of "improving testing suite", #450. Our tests desperately require a revision, especially the integration tests which I plan to completely overhaul. To me this is even a bigger issue than #715 that requires more planning and effort, and I was actually thinking of working on that some time in the upcoming year, after I finish refactoring the C code to make it more modular. In my mind, I thought we would deal first with #450 to increase the coverage of the C code and test better its modular parts, and then proceed to #715 to make sure that all the raised warnings in the tests are expected. Having said that, I don't think the ordering here is very crucial, and I guess we could start working on #715 before #450. But let me know if you think otherwise.

hassamrajpoot:
Agreed on the roadmap . let's wait for @steven-murray's input before moving things around.

Thanks for the context on #450 as well. The ordering isn't strictly crucial, though getting #715 done first might provide a useful baseline , a strict "no unexpected warnings" policy could make the integration test overhaul easier to monitor as tests get restructured.

For this PR specifically: while we wait for @steven-murray, I can strip it down to just the lightconers.py bug fix and update it to use deprecation.DeprecatedWarning as per the project convention. That way the isolated fix gets merged cleanly, and I can open a fresh PR for the test_deprecated.py rollout once the direction is confirmed.

Let me know if that works.

steven-murray:
Hi!

Let me clarify a bit my thoughts on #715, though I think we're on the same page.

The idea with #715 is that we should have one CI job where we run `pytest -Werror`, i.e. any warnings that are raised are flagged as errors, making them much more visible. We do not have to block on this CI job, but it is good to have it go red so we can keep any eye on when we might expect things to fail in the future. Right now, this would fail instantly for MANY of the tests. Thus, to implement this PR fully we would need to:

1. Fix any code where warnings are currently being raised and we didn't intend to have a warning raised (e.g. deprecated code in third-party libraries)
2. Explicitly expect warnings from our own code when appropriate, and assert that they do get raised. This can come in two main forms:
   a) Deprecation warnings, where we should provide an explicit test that we properly raise this warning, and will also fail if we do not remove that code at a certain version.
   b) runtime/user warnings where we should ensure that these warnings are raised at the appropriate places for appropriate inputs.
3. Filter warnings explicitly that we know will be raised, and we are OK with them being raised, but don't need to test them being raised (e.g. warnings from numpy for zero-division where we know we have properly accounted for this). Generally this will be for third-party library warnings. In this last case we should almost always prefer more local filters than global filters, i.e. it is better to filter the warning per-test-function rather than globally in the `conftest`, because we *want* to be alerted when things go awry. Sometimes, it is appropriate to filter warnings at the test-module level, when you know a lot of the tests in that module use the same pattern and all will raise the same warning, and we are OK with that warning being raised.

Ultimately, with a combination of the above, we should be able to run through the entire test suite without any warnings triggering errors, which keeps our logs generally clean so that when real errors happen, they are easier to identify, and also so that new warnings (usually from newly-deprecated third-party code) are caught early, before they become breakages.

On other points:

* Yes I approve of the `deprecation` (or similar) packages to handle our own deprecations
* I agree, having a single module for deprecation tests would be a good idea
* Note that there is a newer (perhaps better) package for managing deprecations we could consider: https://borda.github.io/pyDeprecate/stable/
* I agree that we can do #715 before #450 (either way would work but #715 would make #450 easier)

hassamrajpoot:
Thanks @steven-murray, that breakdown is clear and the three-part approach makes sense. The non-blocking pytest -Werror CI job is a good way to monitor without blocking the main pipeline, and the preference for local over global filters is well taken.

I'll definitely look into pyDeprecate as well, the interface looks really clean.

Since the true scope of #715 is a comprehensive overhaul, here is what I propose for immediate next steps to keep things moving:

1. This PR: I'll strip this down to just the isolated lightconers.py bug fix. Quick question on this: should I use the existing deprecation package Jordan mentioned for this quick fix, and we can evaluate migrating to pyDeprecate in the larger #715 PR? Or would you prefer I use pyDeprecate here right away?

2. Next PR: I'll open a fresh draft PR specifically for #715 and begin working through the checklist you outlined above (starting with the non-blocking CI job and setting up test_deprecated.py).
 

Happy to proceed once you and @jordanflitter confirm that direction.

jordanflitter:
I think that it doesn't matter much how much goals each PR contains, as long as they align with our long-term plans. So I think that the plan you suggested @hassamrajpoot is solid though a bit unbalanced as the first PR is much shorter, while the second PR is much more comprehensive and potentially touching also source code and not just tests. So I'm also okay with unifying the two PRs that you suggested.

Having said that, after a second thought, it might be a better idea to indeed follow your plan, as this PR attempts to merge a change from a forked version into the main branch. Usually, we prefer to merge new code from other branches (not forks) as this makes it easier for other contributors to edit the PR, and if the change is big enough we would rather merge it with the new release in development (now v4.3, #748), and not the main branch (now v4.2). So I actually think it would be better to limit the scope of this PR like you suggested and merge it with either the main branch or #748. As for the complete treatment for #715, I believe it would be better to make a new branch that would ultimately be merged with #748.

Thanks again for contributing to the code, each improvement is highly welcome and appreciated!

hassamrajpoot:
@jordanflitter I've updated this PR to focus strictly on the isolated lightconers.py fix, using the existing deprecation.DeprecatedWarning pattern and reverting all other configuration and test file changes.

Ready for review whenever you get a chance.

I'll branch off v4.3 (#748) for the comprehensive #715 overhaul once this is merged, starting with the non-blocking CI job setup as @steven-murray outlined.

jordanflitter (approved):
Thanks @hassamrajpoot this looks good!

I approved the PR but I think it would be better to add some tests for this new deprecation warning before merging, see for example `test_fixed_vavg_deprecated_warning` and `test_fixed_vavg_is_removed` in `test_input_structs.py`. These tests should ultimately (not in this PR) be in `deprecated.py` (and then removed when we get to v5 and we will remove the old names from the codebase)

hassamrajpoot:
@jordanflitter I've added the two requested unit tests to tests/test_lightconer.py (test_with_equal_cdist_slices_deprecated_warning and test_with_equal_cdist_slices_is_removed using @deprecation.fail_if_not_removed) to enforce cleanup at v5.

jordanflitter:
Thanks! Looks good to me! Feel free to merge (we usually give the pleasure of clicking on the big green button to the author of the PR)

hassamrajpoot:
@jordanflitter Thanks! I'd love to, but the merge option isn't available yet, likely because one check is still showing as cancelled (the Ubuntu apt-get build that hit a network error during dependency installation, unrelated to the code).

jordanflitter:
Try now :)

hassamrajpoot:
@jordanflitter All checks are green but the merge option isn't available on my end, looks like it's only available to organisation members.

steven-murray:
@hassamrajpoot I invited you as a member of the organization so you should be able to merge now!

Issue #669, opened 2026-04-14
Decide on a consistent mean fixing logic for the emissivity fields


jordanflitter:
Currently we do mean fixing of the emissivity fields if `SOURCE_MODEL` is either `CONST-ION-EFF` or `E-INTEGRAL`. This can be seen in the following lines in `SpinTemperatureBox.c`:

```c
mean_sfr_zpp[R_ct] = EvaluateSFRD(zpp, &sc);

....

calculate_sfrd_from_grid(R_ct, delta_box_input, Mcrit_box_input, del_fcoll_Rct,
                                         del_fcoll_Rct_MINI, &ave_fcoll, &ave_fcoll_MINI, &sc);
avg_fix_term = mean_sfr_zpp[R_ct] / ave_fcoll

...

sfr_term = del_fcoll_Rct[box_ct] * z_edge_factor * avg_fix_term *
                                   astro_params_global->F_STAR10;
```

Otherwise, we do mean fixing only if `HMF` is `Watson` or `Watson-z`, as can be seen from these lines in `HaloBox.c` and `scaling_relations.c`.
```c
ScalingConstants hbox_consts;
set_scaling_constants(redshift, &hbox_consts, true);

...
void set_scaling_constants(double redshift, ScalingConstants *consts, bool use_photoncons) {
    consts->redshift = redshift;

    // Set on for the fixed grid case since we are missing halos above the cell mass
    consts->fix_mean = matter_options_global->HMF == 2 || matter_options_global->HMF == 3;
...

set_fixed_grids(M_min, M_max_integral, ini_boxes, mturn_a_grid, mturn_m_grid,
                            &hbox_consts, grids);
...
int set_fixed_grids(double M_min, double M_max, InitialConditions *ini_boxes, float *mturn_a_grid,
                    float *mturn_m_grid, ScalingConstants *consts, HaloBox *grids) {
....
    ScalingConstants _ev_consts = *consts;
    ScalingConstants *ev_consts = &_ev_consts;
...
   if (ev_consts->fix_mean) mean_fix_grids(M_min, M_max, grids, ev_consts);

}
```

We should decide on a consistent logic for the mean fixing, and document it properly. Note that there was a similar discussion in #369.

jordanflitter:
For better documenting what has done until now, I am adding below some inputs that I received from the great @daviesje.

[...]

PR #755, opened 2026-08-10
Always filter emissivity fields

### LINKED ISSUE #668: Filter emissivity fields in all `SOURCE_MODEL` options

jordanflitter:
Fix #668.

Before this PR, the Eulerian source models did not compute the emissivity fields (in `HaloBox`) and filter them to get the radiation fields (in `RadiationFields`). Instead, the density field was filtered in order to compute the "effective emissivity field" for each shell, while the radiation fields were computed on the fly in the evaluation of the spin temperature. This PR changes that and unifies the prescription for computing the emissivity and radiation fields, regardless if the considered source models are Eulerian or Lagrangian. This change allowed to clean many redundant lines in `RadiationFields.c` that were used for the "old" Eulerian source models, making the code much cleaner and more modular (but there's still some work to do...).

In evaluating the spin temperature, the only difference between the two types of source models is that in the Eulerian source models we use the Eulerian density field to compute the emissivity fields on the Eulerian grid, while for the Lagrangian source models we use the Lagrangian field to compute the emissivity fields on the Lagrangian grid, and then we advect them to the Eulerian grid (there's also a multiplication by `1+delta` of the emissivity fields in the Eulerian source models, I am still not sure if it's really required...). There are however still differences in how the code computes the reionization field for the two types of source models, but I left this work for a future PR.

Because this PR changes the output (i.e. realizations of the spin/brightness temperature field) for the Eulerian source models, I have made several tests to confirm that I did not introduce a new bug, see my next comment below.

Other more minor changes that were made:

- There was a bug in `CONST-ION-EFF`. In this source model, it is assumed that the SFRD/emissivity fields follow the time derivative of the collapsed fraction. However, the mean fixing ratio `avg_fix_term` was computed for the plain collapsed fraction, but was applied on the derivative of the collapsed fraction. Compared to what we do in the other source models, it is more consistent to compute `avg_fix_term` for the derivative of the collapsed fraction (which is proportional to the SFRD). That was fixed in this PR.
- A new page, called `Radiation_Fields.rst` was added to the docs. This file describes the history of how emissivity fields and radiation fields were computed in 21cmFAST, including the changes that were made in v4.3 and were part of this PR.
- Previously, `MINIMIZE_MEMORY` affected how the radiation fields were computed in the Eulerian source models. After this PR, its effect on the code has been reduced significantly, which allowed to delete/optimize some tests.

## Summary by Sourcery

Unify emissivity generation and radiation-field filtering across Eulerian and Lagrangian source models.

New Features:
- Compute emissivity fields for Eulerian source models in HaloBox and process them through the same RadiationFields pipeline used by Lagrangian source models.
- Require explicit radiation-field inputs for spin-temperature calculations and expose HaloBox and RadiationFields consistently across simulation outputs and caching.

Bug Fixes:
- Correct CONST-ION-EFF normalization by deriving the mean-fix factor from the collapsed-fraction derivative used for emissivity.
- Fix Eulerian emissivity and X-ray scaling, density weighting, and source-grid handling.

Enhancements:
- Remove the legacy Eulerian radiation-field computation and substantially simplify radiation-field setup, memory management, and source-model branching.
- Reduce MINIMIZE_MEMORY-specific behavior and compute HaloBox fields only when required by enabled outputs or physics.
- Allow Lyman-alpha multiple scattering with Eulerian source models and refine halo-mass range validation.

Documentation:
- Add documentation describing the evolution and unified computation of emissivity and radiation fields.

Tests:
- Update spin-temperature, output-structure, caching, input-validation, and integration coverage for unified Eulerian and Lagrangian radiation-field workflows.
- Remove obsolete MINIMIZE_MEMORY tests and adapt expected outputs to the revised field lifecycle.


docs/Radiation_Fields.rst
```diff
@@ -0,0 +1,353 @@
...
+
+    \Gamma_X = 4 \pi \int_{\nu_{min}}^{\infty} d\nu \sum_i (h_{\rm P}\nu - E^i_{\rm th}) f_{\rm heat} f_i \sigma_i(\nu) J_X(\nu),
+
+where :math:`h_{\rm P}` is Planck's constant, :math:`\nu_{min}` is the minimum frequency for ionizing X-ray photons to escape the host
+galaxy, :math:`i` runs over the species in the IGM that are relevant to the process of energy transfer by X-ray photons (i.e. HI, HeI, HeII),
+:math:`E^i_{\rm th}` is the ionization threshold energy of species :math:`i`, :math:`f_{\rm heat}` is the fraction of the energy of an
+ionized electron of energy :math:`E_e = h_{\rm P}\nu - E^i_{\rm th}` that is deposited as heat in the IGM (a function of :math:`E_e` and
+the local ionization fraction), :math:`f_i` is the number fraction of species :math:`i` in the IGM, and :math:`\sigma_i(\nu)` is the
+cross-section for photo-ionization of a particle from species :math:`i` due to the interaction with an X-ray photon of frequency :math:`\nu`.
+To understand the above experssion, let us consider the following components:
```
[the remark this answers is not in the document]
jordanflitter: Thanks! Fixed in the last commit!

docs/Radiation_Fields.rst
```diff
@@ -0,0 +1,353 @@
...
+Given the heat transfer rate, the X-ray heating rate is given by
+
+.. math::
+
+    \left.\frac{dT_k}{dt}\right|_{X} = \frac{2}{3 k_B} \frac{\Gamma_X}{1 + x_e},
+
+where :math:`k_B` is Boltzmann's constant and :math:`x_e` is the local ionization fraction. The X-ray heating rate is thus a physical
+quantity that can be represented as a 3D box at a given redshift, and is thus considered a "radiation field" in ``21cmFAST``.
+Similarly, the photionization rate due to X-ray radiation, as well as the contribution to the Lyman-alpha flux from X-ray excitation of HI
+atoms, are also considered "radiation fields".
```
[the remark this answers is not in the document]
jordanflitter: Thanks! Fixed in the last commit!

docs/Radiation_Fields.rst
```diff
@@ -0,0 +1,353 @@
...
+
+    \tau_X(\nu, z') = \int_z^{z'} dz'' \frac{c}{(1+z'')H(z'')} \sum_i \bar n_i(z'') \sigma_i(\nu''),
+
+where :math:`\bar n_i(z'')` is the mean number density of species :math:`i` at redshift :math:`z''`. Note that the optical depth in
+``21cmFAST`` is homogeneous and isotropic as it is computed using the mean number density of the IGM, :math:`\bar n_i(z'')`,
+regardless of the actual density field that the photon encountered in its path, thus allowing ``21cmFAST`` to run much faster than
+traditional radiative transfer codes.
+
+The Lyman alpha flux from continuum photons (photons that were emitted with frequency below Lyman beta and redshifted into Lyman alpha) and
+injected photons (photons that were emitted with frequency above Lyman beta and went atomic cascades once they had redshifted into a Lyman
```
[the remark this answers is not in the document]
jordanflitter: Thanks! Fixed in the last commit!

docs/Radiation_Fields.rst
```diff
@@ -0,0 +1,353 @@
...
+.. math::
+
+    \epsilon_X(z, \delta) = \bar\rho_b(z=0) (1+\delta) \zeta_X f_* \frac{d}{dt}f_{\rm coll}(z, M_{\rm turn} ; \delta, M_{\rm cond}),
+
+where :math:`\bar\rho_b` is the mean baryon number density, :math:`f_*` is the star formation efficiency, :math:`\zeta_X` is the number of
+X-ray photons per unit stellar mass (the latter two are free parameters). Meanwhile, the Lyman-alpha emissivity field was given by
+
+.. math::
+
+    \epsilon_\alpha(z, \delta) = \bar\n_b(z=0) (1+\delta) N_{\gamma /{\rm b}} f_* \frac{d}{dt}f_{\rm coll}(z, M_{\rm turn} ; \delta, M_{\rm cond}),
```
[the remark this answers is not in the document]
jordanflitter: Thanks! Fixed in the last commit!

docs/Radiation_Fields.rst
```diff
@@ -0,0 +1,353 @@
...
+
+As for the conditional collapsed fraction, it is given by
+
+.. math::
+
+    f_{\rm coll}(z, M_{\rm turn} ; \delta, M_{\rm cond}) = \frac{1}{\bar\rho_m} \int_{M_{\rm turn}}^{\infty} dM_h M_h\frac{dn}{dM_h}(z, M_h ; \delta, M_{\rm cond}),
+
+where :math:`\bar\rho_m` is the mean matter density and :math:`dn/dM_h(z, M_h ; \delta_{\rm cond}, M_{\rm cond})` is the conditional halo
+mass function in a region of size :math:`R(M_{\rm cond}) = [3M_{\rm cond}/(4\pi \bar\rho_m)]^{1/3}` and overdensity :math:`\delta_{\rm cond}`.
+For the calculation of the conditionl collapsed fraction, ``21cmFAST`` v1.0.0 assumed that the conditional halo mass function was given
```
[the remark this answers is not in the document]
jordanflitter: Thanks! Fixed in the last commit!

docs/Radiation_Fields.rst
```diff
@@ -0,0 +1,353 @@
...
+
+    f_{\rm coll}(z, M_{\rm turn} ; \delta, M_{\rm cond}) = {\rm erfc}\left[\frac{\delta_c - \delta}{D(z)\sqrt{2(\sigma^2(M_{\rm turn}) - \sigma^2(M_{\rm cond}))}}\right],
+
+where :math:`{\rm erfc}` is the complementary error function, :math:`\delta_c=1.686` is the critical overdensity for collapse,
+and :math:`\sigma^2(M)` is the variance of the linear density field, smoothed with a top-hat filter of
+radius :math:`R(M) = [3M/(4\pi \bar\rho_m)]^{1/3}`.
+
+While ``21cmFAST`` v1.0.0 used the analytical result for the conditional collapsed fraction from the extended Press-Schechter formalism
+in order to assess the fluctuations in the emissivity fields, the code also normalized the mean collapsed fraction in the box to match
+the global collapsed fraction, as given by solving the above intergral numerically with the user's selected halo mass function
```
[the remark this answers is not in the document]
jordanflitter: Thanks! Fixed in the last commit!

docs/Radiation_Fields.rst
```diff
@@ -0,0 +1,353 @@
...
+(which by default was the Sheth-Tormen mass function).
+
+Effective Emissivity Fields
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Given the above definitions for the emissivity fields, the effective emissivity fields are achieved by interpolating in time/redshift and
+filtering in space. Since ``21cmFAST`` v1.0.0 did not have the emissivity fields on a grid, the interpolation and filtering was done in
+the following way. Firstly, given they Eulerian density field :math:`\delta_{\rm E}(z, \mathbf{x})` at redshift :math:`z` (the current
```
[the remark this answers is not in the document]
jordanflitter: Thanks! Fixed in the last commit!

docs/Radiation_Fields.rst
```diff
@@ -0,0 +1,353 @@
...
+the code as in previous versions. The LW flux is given by
+
+.. math::
+
+    J_{\rm LW}(z, \mathbf{x}) = h_{\rm P}\nu_\alpha\frac{(1+z)^2}{4\pi}\sum_{n=2}\int_z^{z_{\rm max}(n)} \frac{cdz''}{H(z'')} I_\alpha(\nu') \epsilon_\alpha^{\rm eff}(z', \mathbf{x}),
+
+where :math:`\nu_\alpha` is the Lyman-alpha frequency. Several notes:
+
+* Note that unlike previous radiation fields, the LW flux contains units of energy, owned by the factor of :math:`h_{\rm P}\nu_\alpha`.
+* The contribution to the LW flux came from both atomic cooling galaxies (ACGs) and MCGs, as the SFRD in both populations is modedled a bit
```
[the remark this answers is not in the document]
jordanflitter: Thanks! Fixed in the last commit!

docs/Radiation_Fields.rst
```diff
@@ -0,0 +1,353 @@
...
+differently, mostly in the modeling of the star formation efficiency, the turnover mass and the duty fraction (see more details on the
+latter two in :doc:`M_TURN`). Likewise, the contribution to the X-ray and Lyman-alpha fluxes also came from both ACGs and MCGs.
+* It was assumed that ACGs contained only popII stars, while MCGs contained only popIII stars. Therefore,
+the shape of the SEDs in ACGs and MCGs were also different in the evaluation of Lyman-alpha and LW fluxes, as well their amplitudes
```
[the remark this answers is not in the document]
jordanflitter: Thanks! Fixed in the last commit!

docs/Radiation_Fields.rst
```diff
@@ -0,0 +1,353 @@
...
+  This option was equivalent to setting ``USE_MASS_DEPNDENT_ZETA`` to True in previous versions.
+
+* ``"L-INTEGRAL"``:
+  This is a new source model that was introduced in v4.0.0. It was similar to ``"E-INTEGRAL"``, but the emissivity
+  fields were computed on the Lagrangian density grid, and then mapped to the Eulerian grid (see details below).
+
+* ``"DEXM-ESF"``:
+  This is a new source model that was introduced in v4.0.0, similar to ``"L-INTEGRAL"``, but the emissivity fields recieved
+  contributions from resolved discrete halos that were found via the excursion-set formalism (ESF), as was implemented by
+  the ``DEXM`` algorithm (see Mesinger & Furlanetto 2007, https://arxiv.org/pdf/0704.0946).
```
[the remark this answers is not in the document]
jordanflitter: Thanks! Fixed in the last commit!

docs/Radiation_Fields.rst
```diff
@@ -0,0 +1,353 @@
...
+of the code. Namely, these fields were an internal part of the spin temperature calculation, and the effective emissivity fields
+were computed by filtering the Eulerian density field with a top-hat filter and scaling by the linear growth factor for interpolating
+the fields at :math:`z'`:
+
+.. math::
+
+    \epsilon(z, \mathbf{x}) \to \epsilon^{\rm eff}(z', \mathbf{x})=\epsilon[z', \delta_{\rm E}^R(z,\mathbf{x}) D(z')/D(z)].
+
+In contrast, for the three "Lagrangian" source models, the emissivity fields were computed on the Lagrangian density grid, and were
+then advected to the Eulerian grid by an algorithm similar to 2LPT. Moroever, the emissivity fields in the "Lagrangian" source models
```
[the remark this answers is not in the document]
jordanflitter: Thanks! Fixed in the last commit!

docs/Radiation_Fields.rst
```diff
@@ -0,0 +1,353 @@
...
+
+In the fourth public release of ``21cmFAST`` v4.0.0 (Davies et al. 2025, https://arxiv.org/pdf/2504.17254), the evaluation of the
+emissivity fields and the radiation fields was changed for several new configurations that were introduced in that version. This was
+controlled by the new enum parameter ``SOURCE_MODEL``, which had five options:
+
+* ``"CONST-ION-EFF"``:
+  This option was equivalent to setting ``USE_MASS_DEPNDENT_ZETA`` to False in previous versions.
+
+* ``"E-INTEGRAL"``:
+  This option was equivalent to setting ``USE_MASS_DEPNDENT_ZETA`` to True in previous versions.
```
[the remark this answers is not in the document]
jordanflitter: Thanks! Fixed in the last commit!

docs/Radiation_Fields.rst
```diff
@@ -0,0 +1,353 @@
...
+
+    \epsilon(z, \mathbf{x}) \to \epsilon^{\rm eff}(z', \mathbf{x})=\epsilon^{R_{\rm i}, R_{\rm o}}(z', \mathbf{x}).
+
+The difference between "Eulerian" and "Lagrangian" source models in v4.3.0 (specifically, between ``"E-INTEGRAL"`` and ``"L-INTEGRAL"``) was
+that in the former, the emissivity fields were computed on the Eulerian density grid, while in the latter, the emissivity fields were
+computed on the Lagrangian density grid and then advected to the Eulerian grid.
+
+Furthermore, in v4.3.0, the effect of Lyman-alpha multiple scattering could have been applied via a new flag called ``LYA_MULTIPLE_SCATTERING``,
+regardless if the simulated source model was "Eulerian" or "Lagrangian". When this flag was set to True, the effective emissivity for
+Lyman-alpha photons was achieved by filtering the interpolated emissivity field with a generalization of the sphericall shell filter
```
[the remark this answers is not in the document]
jordanflitter: Thanks! Fixed in the last commit!

jordanflitter:
## Validation tests

In order to validate the new code, I introduced a temporary `USE_NEW_CODE` flag. I used that flag to run simulations with the old/new Eulerian source models. Since the old Eulerian source models did not have `HaloBox` and `RadiationFields` outputs, I could only study the fields from `TsBox`, and `spin_temperature` is the most relevant one because it depends on both X-ray heating (controlled by the `halo_xray` field in `HaloBox`) and WF coupling (controlled by the `halo_sfr` field in `HaloBox`).

I started first by running two quick `run_global_evolutions` simulations, and measuring the residuals in the global spin temperature for each Eulerian source model.

Old vs New code (global evolution, E-INTEGRAL, with mini-halos)

Old vs New code (global evolution, E-INTEGRAL, without mini-halos)

Old vs New code (global evolution, CONST-ION-EFF)

As can be seen, the residuals of the global spin temperature are almost sub-percent at all redshifts and Eulerian source models, and because all the integration tests for the Lagrangian source models have passed without modifying the datafiles, this means that there is no bug in the prefactors that multiply the integrals, or in the mean fixing.

Interestingly, if I repeat the same comparison, but for the global spin temperature from `run_lightcone`, there is a divergence between the old and new Eulerian source models at low redshifts. However, **this divergence is expected!** (as I explain in the comment below)
To convince myself that indeed there is no bug in the new Eulerian source models, I did the following test. In `HaloBox.c`, I used the temporary `USE_NEW_CODE` flag to force the new `E-INTEGRAL` source model to use the Lagrangian density field for the emissivity fields (instead of the Eulerian density field). Under this configuration, `E-INTEGRAL` and `L-INTEGRAL` shared the same input density field for `HaloBox`, so if there was a bug in the new code for the evaluation of the emissivity fields (given the input density field), then a mismatch between the global spin temperatures of the two source models would have been observed. Yet, there wasn't such a mismatch, the two global spin temperatures agreed up to sub-percent differences, which means that there is no bug in the new code for the evaluation of the emissivity fields (given the input density field). **Therefore, the divergence that is seen in the global spin temperature (from `run_lightcone`) between the old and new `E-INTEGRAL` codes is a real one!** I explain its origin in the comment below. 

Old vs New code (lightcone, CONST-ION-EFF)

Old vs New code (lightcone, E-INTEGRAL, with mini-halos)

Old vs New code (lightcone, E-INTEGRAL, without mini-halos)

jordanflitter:
## Origin of the discrepancy at low redshifts (Jensen's inequality!)

The main difference between the old and new codes for the Eulerian source models is flipping the order of operations:

- In the old code, the Eulerian density field was filtered, and then the emissivity function was applied on the filtered density field.
- In the new code, the emissivity function is applied on the Eulerian density field, and then the emissivity field is filtered.

Because the emissivity function is a convex function of the density field, and filtering can be viewed as a weighted average, we can apply the Jensen's inequality: for a convex function `f(delta)`, `<f(delta)> > f(<delta>)`. Thus, from the Jensen's inequality, we can predict that the new code should yield a larger effective emissivity, and stronger radiation fields! (i.e. stronger X-ray and Lya fluxes)

At sufficiently low redshifts, the WF coupling saturates, and the spin temperature follows the gas kinetic temperature. Because of the increased X-ray flux in the new code, X-ray heating is more efficient, which means that the gas heats up faster and the spin temperature is higher. We see exactly that when we compare the coeval boxes of the spin temperature between the old and new codes! Interestingly, it also seems that the spin temperature field is smoother in the new code.
 
Old vs New code (coeval, CONST-ION-EFF, spin temperature)

Old vs New code (coeval, E-INTEGRAL, without mini-halos, spin temperature)

Because the spin temperature is higher (but not much higher compared to CMB temperature), the brightness temperature is also higher. Because the spin temperature in the new code is smoother, also the brightness temperature becomes smoother, so power in small scales (high `k`) is expected to be suppressed in the new code compared to the old one!

Old vs New code (coeval, CONST-ION-EFF, brightness temperature)

Old vs New code (coeval, E-INTEGRAL, without mini-halos, brightness temperature)

src/py21cmfast/drivers/single_field.py
```diff
@@ -884,32 +915,9 @@ def compute_spin_temperature(
...
-                # TODO: the radiation arrays below are defined as np.float64, but should be np.float32 - see https://github.com/21cmfast/21cmFAST/issues/744
-                dtype = np.float32 if "filtered" in array else np.float64
-                setattr(
-                    radiation_fields,
-                    array,
-                    Array(shape=shape, dtype=dtype)
-                    .initialize()
-                    .with_value(val=np.zeros(shape)),
-                )
+        raise ValueError(
```
steven-murray: Wouldn't it be better just to make `radiation_fields` a required argument of the function now?

src/py21cmfast/io/caching.py
```diff
@@ -326,13 +326,12 @@ def from_inputs(cls, inputs: InputParameters, cache: OutputCache) -> Self:
...
             "PerturbedField": {},
+            "HaloBox": {},
             "IonizedBox": {},
             "BrightnessTemp": {},
         }
         if inputs.astro_options.USE_TS_FLUCT:
-            others |= {"TsBox": {}}
-        if inputs.matter_options.lagrangian_source_grid:
-            others |= {"RadiationFields": {}, "HaloBox": {}}
+            others |= {"RadiationFields": {}, "TsBox": {}}
```
steven-murray, line 334: We should probably be careful to be able to read old outputs (using `XraySourceBox` as well...)

src/py21cmfast/src/HaloBox.c
```diff
@@ -263,38 +398,83 @@ void get_cell_integrals(double dens, double l10_mturn_a, double l10_mturn_m,
...
+    if (source_model_uses_eulerian_grids(matter_options_global->SOURCE_MODEL)) {
+        // TODO: right now n_ion from HaloBox is not used at all in IonisationBox.c for the Eulerian
+        // source models, so whatever we compute here is not very relevant at the moment. This might
+        // change however in the future. When that happens, we will have to figure out whether the
+        // multiplication of n_ion by (1 + delta) is the right thing to do for the Eulerian source
+        // models, since this factor does not seem to appear in IonisationBox.c for the Eulerian
+        // source models. I wonder if it's because what we call n_ion is not a standard emissivity
+        // field, but rather a ratio: it is the number of ionizing photons that have escapted into
+        // the IGM over the number of baryons in the cell. It's possible that both the numerator and
+        // denominator are scaled by (1 + delta) on the Eulerian grid, and thus the ratio is not
```
steven-murray, line 466: Something to discuss with Boss Man @andreimesinger

src/py21cmfast/wrapper/inputs.py
```diff
@@ -2546,6 +2543,12 @@ def check_halomass_range(inputs: InputParameters) -> None:
     if it is above the turnover mass.
     """
     # There are no problems if we are not using halos
+    # TODO: the last warning below (that we also test), is only reachable if max_integral_mass is larger than
+    # the upper mass_limits, but if we have halos then the latter is max_dexm which is exactly the same as max_integral_mass.
+    # Thus the conclusion is that this warning is reachable only when we use L-INTEGRAL. Therefore, I am keeping the quick return
+    # if we don't use lagrangian_source_grid (instead of a quick return if we don't use has_discrete_halos), but I suggest to
+    # modify the code below to include also Eulerian source models, (and the comment above that says that "there are no problems
+    # if we are not using halos")
```
steven-murray: Trying to follow this... should we not implement that in this PR then?

steven-murray (commented):
Thanks @jordanflitter for the code and for the tests and plots. This is looking good to me, with the caveat of some comments.

docs/Radiation_Fields.rst
```diff
@@ -0,0 +1,354 @@
...
+* :math:`(h_{\rm P}\nu - E^i_{\rm th})f_{\rm heat}` is the energy deposited to the IGM as heat per electron with energy :math:`E_e`.
+* Once we apply :math:`\int_{\nu_{min}}^{\infty} d\nu \sum_i` we account for all the species and all the frequencies. The multiplication
+  by :math:`4\pi` is needed in order to account all the directions from which the X-ray photons come. Combining everything together we end
+  up with the desired heat transfer rate per gas particle.
+
+Given the heat transfer rate, the X-ray heating rate is given by
+
+.. math::
+
+    \left.\frac{dT_k}{dt}\right\Bigg|_X = \frac{2}{3 k_B} \frac{\Gamma_X}{1 + x_e},
```
steven-murray: This equation is broken

docs/Radiation_Fields.rst
```diff
@@ -0,0 +1,354 @@
...
+The X-ray optical depth is given by
+
+.. math::
+
+    \tau_X(\nu, z') = \int_z^{z'} dz'' \frac{c}{(1+z'')H(z'')} \sum_i \bar n_i(z'') \sigma_i(\nu''),
+
+where :math:`\bar n_i(z'')` is the mean number density of species :math:`i` at redshift :math:`z''`. Note that the optical depth in
+``21cmFAST`` is homogeneous and isotropic as it is computed using the mean number density of the IGM, :math:`\bar n_i(z'')`,
+regardless of the actual density field that the photon encountered in its path, thus allowing ``21cmFAST`` to run much faster than
+traditional radiative transfer codes.
```
steven-murray, line 79: I don't have time to think this through now, so just wanted to flag: is this approximation, where the mean emmissivity is used to get the optical depth, rather than the inhomogenous emmisivitiy, related to the reason why originally we were doing `emissivity(filtered density)` rather than `filtered(emissivity(density))`?

docs/Radiation_Fields.rst
```diff
@@ -0,0 +1,354 @@
...
+
+    \int_{\nu_{min}}^{\infty} d\nu \int_z^{\infty} dz' e^{-\tau_X(\nu, z')} \approx \int_z^{\infty} dz' \int_{\min[\nu_{min},\nu_{\tau_X=1}]}^{\infty} d\nu,
+
+where :math:`\nu_{\tau_X=1}` is the frequency at which the X-ray optical depth is unity. In other words, the code assumed that only photons
+with X-ray optical depth less than unity contribute to the X-ray radiation fields.
+
+Emissivity Fields
+~~~~~~~~~~~~~~~~~
+
+For the emissivity fields, the code assumed that they follow the derivative of the conditional collapsed fraction of halos above the
```
steven-murray, line 115: Not sure what's going on with the structure of the document here... is this "Emissivity Fields" section supposed to be nested under v1.0.0?

docs/Radiation_Fields.rst
```diff
@@ -0,0 +1,354 @@
...
+turnover mass :math:`M_{\rm turn}` (see more details about the turnover mass in :doc:`M_TURN`). The X-ray emissivity field was
+given by
+
+.. math::
+
+    \epsilon_X(z, \delta) = \bar\rho_b(z=0) (1+\delta) \zeta_X f_* \frac{d}{dt}f_{\rm coll}(z, M_{\rm turn} ; \delta, M_{\rm cond}),
```
steven-murray, line 121: For my own enrichment: this is saying that I generate X-ray photons at a rate proportional to the rate at which I form stars. I suppose this is empirically reasonable but it seems a bit strange: in my imagination if I had a galaxy that had stopped forming stars, it could still produce X-ray photons. Is there some empirical work that this assumption is based on?

docs/Radiation_Fields.rst
```diff
@@ -0,0 +1,354 @@
...
+As for the conditional collapsed fraction, it is given by
+
+.. math::
+
+    f_{\rm coll}(z, M_{\rm turn} ; \delta, M_{\rm cond}) = \frac{1}{\bar\rho_m} \int_{M_{\rm turn}}^{\infty} dM_h M_h\frac{dn}{dM_h}(z, M_h ; \delta, M_{\rm cond}),
+
+where :math:`\bar\rho_m` is the mean matter density and :math:`dn/dM_h(z, M_h ; \delta_{\rm cond}, M_{\rm cond})` is the conditional halo
+mass function in a region of size :math:`R(M_{\rm cond}) = [3M_{\rm cond}/(4\pi \bar\rho_m)]^{1/3}` and overdensity :math:`\delta_{\rm cond}`.
+For the calculation of the conditional collapsed fraction, ``21cmFAST`` v1.0.0 assumed that the conditional halo mass function was given
+by the Sheth-Tormen conditional mass function. Under this assumption, the conditional collapsed fraction has an analytical form,
```
steven-murray, line 145: Is this the ST CHMF or the PS CHMF?

docs/Radiation_Fields.rst
```diff
@@ -0,0 +1,354 @@
...
+    f_{\rm coll}(z, M_{\rm turn} ; \delta, M_{\rm cond}) = {\rm erfc}\left[\frac{\delta_c - \delta}{D(z)\left(2(\sigma^2(M_{\nu}) - \sigma^2(M_{\rm cond}))\right)^{1/2}}\right],
+
+where :math:`{\rm erfc}` is the complementary error function, :math:`\delta_c=1.686` is the critical overdensity for collapse,
+and :math:`\sigma^2(M)` is the variance of the linear density field, smoothed with a top-hat filter of
+radius :math:`R(M) = [3M/(4\pi \bar\rho_m)]^{1/3}`.
+
+While ``21cmFAST`` v1.0.0 used the analytical result for the conditional collapsed fraction from the extended Press-Schechter formalism
+in order to assess the fluctuations in the emissivity fields, the code also normalized the mean collapsed fraction in the box to match
+the global collapsed fraction, as given by solving the above integral numerically with the user's selected halo mass function
+(which by default was the Sheth-Tormen mass function).
```
steven-murray, line 158: So, I *think* what you are saying here is that the CHMF equation here, if integrated over  the PDF of delta, by construction gives the global Press-Schechter HMF. However, since the user is able to select a non-press-shechter global HMF, we still use this PS formula for the CHMF, but we apply a renormalization such that `\int d delta p(delta) CHMF == Custom_HMF(sigma)`. Is that right?

docs/Radiation_Fields.rst
```diff
@@ -0,0 +1,354 @@
...
+
+.. math::
+
+    J_{\rm LW}(z, \mathbf{x}) = h_{\rm P}\nu_\alpha\frac{(1+z)^2}{4\pi}\sum_{n=2}\int_z^{z_{\rm max}(n)} \frac{cdz''}{H(z'')} I_\alpha(\nu') \epsilon_\alpha^{\rm eff}(z', \mathbf{x}),
+
+where :math:`\nu_\alpha` is the Lyman-alpha frequency. Several notes:
+
+* Note that unlike previous radiation fields, the LW flux contains units of energy, owned by the factor of :math:`h_{\rm P}\nu_\alpha`.
+* The contribution to the LW flux came from both atomic cooling galaxies (ACGs) and MCGs, as the SFRD in both populations is modeled a bit
+differently, mostly in the modeling of the star formation efficiency, the turnover mass and the duty fraction (see more details on the
```
steven-murray: syntax error here which abruptly stops the dot-point list.

docs/Radiation_Fields.rst
```diff
@@ -0,0 +1,354 @@
...
+were computed by filtering the Eulerian density field with a top-hat filter and scaling by the linear growth factor for interpolating
+the fields at :math:`z'`:
+
+.. math::
+
+    \epsilon(z, \mathbf{x}) \to \epsilon^{\rm eff}(z', \mathbf{x})=\epsilon[z', \delta_{\rm E}^R(z,\mathbf{x}) D(z')/D(z)].
+
+In contrast, for the three "Lagrangian" source models, the emissivity fields were computed on the Lagrangian density grid, and were
+then advected to the Eulerian grid by an algorithm similar to 2LPT. Moreover, the emissivity fields in the "Lagrangian" source models
+were computed as an intermediate output of the code (stored in the ``HaloBox`` class). This enabled to interpolate the emissivity fields
```
steven-murray, line 310: Do you really mean "interpolate" here?

src/py21cmfast/drivers/single_field.py
jordanflitter: Yes, you are right! (especially now that we use `radiation_fields` for every source model).

However, we actually need `radiation_fields` only below `Z_HEAT_MAX` (because above it we assume that there is no astrophysics). So I adopted your suggestion in the meantime and I will deal with it in my next future PR (that would clean things a bit more).

src/py21cmfast/io/caching.py
jordanflitter: The name `XraySourceBox` essentially doesn't exist now in the codebase (up to the new file in the docs). Are you worried that old cache files that contained fields like `XraySourceBox` could not be read now? This is a good comment, but I think it's more related to #749 than here :)

src/py21cmfast/wrapper/inputs.py
jordanflitter: I am not sure. I am bit confused about this function. We test that the last warning in this function is triggered in `test_input_structs.py::test_halomass_ranges` with `SOURCE_MODEL="L-INTEGRAL"`, but I think it doesn't make sense because we are supposed to integrate over the entire mass-range in this source model, with no gaps. So I think this function/test should be revised, and I am not sure if it should be in this PR (because this check doesn't do anything with Eulerian source models, I just wonder if it's really relevant to `L-INTEGRAL`).

docs/Radiation_Fields.rst
jordanflitter: You are right, MathJax sucks 😝

Without `\Bigg`, the long vertical bar on the RHS is broken into 3 short vertical bars on top of each other, and mighty Gemini suggested me to use `\Bigg` which clearly doesn't work well. I guess that having 3 short vertical bars on top of each other is better, but I'm not sure how to fix it.

docs/Radiation_Fields.rst
jordanflitter: This comment has nothing to do with emissivity, or with the changes that were made in this PR.

The integrand of the optical depth is proportional to the number density that the photon sees during its propagation. Because using the actual number density along the path is REALLY computationally expensive (this would make the fluxes a function of also the incoming angle!), what we do in `21cmFAST` (and other semi-numerical codes) is to simply take the background homogeneous value for the number density.

docs/Radiation_Fields.rst
jordanflitter: Good question. Everything above the v1.0.0 section is about the plain theory (up to the optical depth approximation that is mentioned there).
The emissivity fields are model-dependent (for example, in v1 the code assumed that the emissivity fields are proportional to the derivative of the collapsed fraction. In the versions that came afterwards, the model for the emissivity fields was changed). So I think the modeling of the emissivity fields has to be described under the v1.0.0 section, but maybe with a better name for the title: "Emissivity fields" -> "Emissivity fields model".
In my opinion however, the computation of the **effective** emissivity fields (that are found in the expression for the fluxes) given the emissivity fields (that are computed from the density field) is **not** model-dependent. But sadly, in older versions of `21cmFAST` with Eulerian source models the effective emissivity fields were computed differently (and dare I say, wrongly) so they are also first described under the v1.0.0 section.

docs/Radiation_Fields.rst
jordanflitter: I completely agree with everything you said ("you hit the nail on the head!", as LLMs like to say). 

Up to this day, we still assume that the emissivity fields are proportional to the SFRD. This is very clear from the modeling of the X-ray emissivity, which is proportional to `L_X/SFR *SFRD`, where `L_X/SFR` is our astro parameter (bad naming!), `L_X/SFR ~ 10^40 erg/sec *yr/M_sun` (it actually accounts for photons up to energy 2 keV). Anyway, the v2 paper where this modeling is first introduced is https://arxiv.org/pdf/1809.08995, where it seems like they took the empiric relation from https://arxiv.org/pdf/1702.00409. But I totally agree with you, I always thought that it's weird to assume that the star-formation timescale is the relevant timescale for the calculation of the emissivity (but maybe I'm missing something...). Anyway, this is the justification for the X-ray emissivity. As for the emissivity that is used for the Lya flux, I don't know what is the justification to assume that it is proportional to the SFRD.

And note that while you reached the correct conclusion, that part you were commenting about was regarding the modeling in v1. In that paper/version, there was no reference to the SFRD in the model, but if we compare the equations with v2 (which was the first 21cmFAST paper to use the term "SFRD" in the context of modeling the emissivity), then we can see that the SFRD in v1 was proportional the derivative of the collapsed fraction.

docs/Radiation_Fields.rst
jordanflitter: So in theory, the conditional mass function that should have been used there is ST, but since the collapsed fraction only has analytic from with PS, then the PS mass function was used in computing the local conditional collapsed fraction. Then, as the text below explains, the ST mass function was used to normalize the collapsed fraction, such that the mean collapsed fraction in the box would match the theoretical global collapsed fraction (while assuming the ST mass function).

docs/Radiation_Fields.rst
jordanflitter: Yes, that's exactly correct ("you hit the nail on the head again!")

Today btw, we don't do mean fixing if we use ST mass function and Lagrangian source models. This is clear from the following line (in `scaling_relations.c`):
```c
consts->fix_mean =
        source_model_uses_eulerian_grids(matter_options_global->SOURCE_MODEL) ||
        matter_options_global->HMF == HMF_WATSON || matter_options_global->HMF == HMF_WATSON_Z ||
        matter_options_global->HMF == HMF_REED07 || matter_options_global->HMF == HMF_YUNG24;
```
I don't like this logic and indeed we have an open issue about it, #669.

docs/Radiation_Fields.rst
jordanflitter: Thanks! It should be fixed now.

docs/Radiation_Fields.rst
jordanflitter: Yes, we have a function in `single_fields.py` that does exactly that: `interp_halo_boxes`.

docs/Radiation_Fields.rst
jordanflitter: Oh I now realized what you meant! Yeah, I got mixed up between sections formatting and subsections formatting. I think it should be okay now.

jordanflitter: Very weird, I changed the formatting in the section and sub-section titles, but it continues to render the "Emissivity fields" title bigger than "21cmFAST v1.0.0"

src/py21cmfast/io/caching.py
jordanflitter: There was no `xray_soruce_box` field in `Coeval`, so it's not a big issue as it may seem.

src/py21cmfast/src/HaloBox.c
jordanflitter: Let's leave it as a comment in the code for the meantime.

docs/Radiation_Fields.rst
jordanflitter: Okay now the structure of the document is finally fixed! I'm resolving this comment.

src/py21cmfast/wrapper/inputs.py
jordanflitter: I fixed `check_halomass_range` and `test_halomass_ranges` in the last commit.

docs/Radiation_Fields.rst
steven-murray: Got it.

docs/Radiation_Fields.rst
steven-murray: Got it :-)

Issue #691, opened 2026-05-18
Final memory issues cleanup


jordanflitter:
I list here some small issues I encountered while working on #337. When we get to v4.3, we could turn them into sub-issues.

- [ ] **Understand remaining anomalies in the RAM curves**. There are still open questions in how the RAM curves look like (see last plots in #337). Most notably, it is not clear why the RAM curve is not dead flat during the execution of `_redshift_loop_generator` (when there are no `HaloBoxes`).
- [ ] **Understand why the `actualsize` of `InitialConditions` is not constant**. It seems like the actual memory size that `InitialConditions` takes is not constant, see script in #337 (which is based on this [article](https://towardsdatascience.com/the-strange-size-of-python-objects-in-memory-ce87bdfbb97f/)).
- [ ] **Allocate some boxes with `np.float32` instead of `np.float64`** - see #744.
- [ ] **Replace `MINIMIZE_MEMORY` flag with an enum**. We use today the `MINIMIZE_MEMORY` flag for three purposes; (1) for minimizing the allocated memory in the C code during the computation of spin temperature (when there are no `HaloBoxes`), (2) for saving the allocation of unnecessary boxes in the reionization module, and (3) for purging `perturbed_field` structs (if they are written to disk). The first purpose will be eliminated when we fix #552 and #668 as the future `RadiationFields` boxes will be evaluated from only allocating 3D boxes. The second purpose will also be eliminated when we fix #642, as users will directly control which boxes they want to allocate and compute. Therefore, only the third purpose, of controlling whether the `perturbed_field` structs are purged, will remain. However, I think we should extend this boolean logic to more structs, not just `PerturbedField`. For example, in order to reduce RAM even more, users might like to `purge` **current** instances of `TsBox` and `IonizedBox` if they are saved on disk (unlike the **previous** instances which we always purge when `_redshift_loop_generator` finishes an iteration). While purging the current boxes lowers the RAM, it could increase runtime due to reading and loading large boxes, so we should give the users more control on how memory is managed during the simulation. I suggest therefore that we extend the boolean `MINIMIZE_MEMORY` flag into an enum variable that gives more flexibility.
- [ ] **Initialization of current boxes in python vs. C.** See #687.
[...]

PR #759, opened 2026-08-24
Refactor of the calculation of the radiation fields

jordanflitter:
This PR does a lot of refactors to clean up the code after the merger of #749.

1. A new subclass of `OutputStructZ` was introduced - `RadiationFieldsSetup`. This object corresponds to the C-struct with the same name that was introduced in #749, but it now contains many more fields, as the new fields replace the global arrays that were used in `RadiationFields.c`. 
2. A new single field function was introduced - `setup_radiation_fields`.
3. The output of `setup_radiation_fields`, an object of type `RadiationFieldsSetup` can now be used as an input to `compute_radiation_fields` or be computed internally in this function. Together with flipping the order of the integrated shells (now from largest to smallest), this allowed to remove the hacky `mode` input in `UpdateRadiationFields` and the hacky function `_interpolate_and_evaluate_radiation_fields` that were introduced in #749. Now `UpdateRadiationFields` is called only once per shell, with no need to call it for initializations and cleanups like before.
4. In `_redshift_loop_generator`, we call `setup_radiation_fields` once before entering the redshift loop, and we use its output, `rad_setup`, as an input to `compute_radiation_fields`. This allowed to allocate many arrays only once **per simulation** (not per iteration), similarly as was done with the global arrays that were used in `RadiationFields.c`, for optimizing allocation efficiency. We also purge `rad_setup` at the end of `_redshift_loop_generator`, to free the arrays that were allocated in `rad_setup`. This new logic allowed to remove the hacky `TsInterpArraysInitialised` and `cleanup` logics that were used in `RadiationFields.c`.
5. The radii and redshifts of the integrated shells are set in python in `setup_shells`, a new method in `RadiationFieldsSetup`. This reduced the redundant calculations of the shell properties in setting up the radiation fields, as well as in accumulating the shells in the trapezoidal integration for the radiation fields. Moreover, this allowed to remove the call to `setup_z_edges` in the C code, since it was redundant to the calculations that were made in python, and we can now use the information that is contained in `rad_setup`. As a matter of fact, the exclusive usage of the shells (as defined in python) in C **revealed a bug**: In `setup_z_edges` the redshifts of the shells were computed by Taylor approximating the cosmological relation z(r) to convert from comoving distance to redshift. This was done in a progressive manner, such that the redshift of the next shell was inferred from the previous one. Due to the Taylor approximation, this created an accumulative error in the evaluation of the redshifts of the outermost shells. In contrast, in python we use the `astropy` built-in conversion from comoving distance to redshift, which is more exact. Due to this change, all the integration tests with `USE_TS_FLUCT=True` in their configuration failed: the global brightness temperature had relative difference (compared to what was stored in the datafiles) that grew with each iteration, with a final relative difference of about 7e-3, which is larger than the threshold of `rtol=1e-3`. In my opinion, had we continued the simulation in the tests to lower redshifts below `z=18`, such that the simulation would have contained more iterations, the relative difference would have been higher than one percent, due to the accumulative error. I therefore also updated the datafiles, in order to let the integration tests to pass smoothly.
6. `R_inner` and `R_outer` were removed from the signature of `UpdateRadiationFields` as they are now computed inside the function, using the information that the input `rad_setup` carries.
7. `pertrubed_field_redshift` was removed from the signature of `UpdateRadiationFields`, because we always have `pertrubed_field_redshift=redshift` (this wasn’t the case in old versions of the code). 
8. Only one `dummy_box` is allocated in `one_annular_filter`, since there is no need to allocate two boxes. This reduces the used peak memory (and possibly also runtime as `one_annular_filter` is a relatively hot function: for each snapshot, it is called for every shell between 2-5 times).
9. `mean_log10_Mcrit_LW ` and all the `filtered_` fields were moved from `RadiationFields` to `RadiationFieldsSetup`, making `RadiationFields` to hold only physically meaningful fields.
10. New C modules were introduced: `RadiationFieldsSetup.c` and `RadiationFieldsSetup.h`. This change was made in order to distinguish better between the calculations we do for setting up the calculations required for the radiation fields (in the new modules) and the calculations done to evaluate the radiation fields (in `RadiationFields.c`, as before).

## Summary by Sourcery

Refactor radiation-field setup and evaluation around reusable per-simulation state while correcting shell redshift integration and reducing memory overhead.

New Features:
- Introduce a reusable RadiationFieldsSetup output structure and setup_radiation_fields function for preparing radiation-field calculations.
- Allow callers to prepare radiation-field state explicitly or have it created automatically by compute_radiation_fields.

Bug Fixes:
- Use Python-computed shell redshifts for radiation-field integration, avoiding accumulated errors from the previous approximate C-side calculation.
- Update integration test data to reflect the corrected shell-redshift calculation.

Enhancements:
- Separate radiation-field setup responsibilities from field evaluation and remove the previous setup, evaluation, cleanup, and mode-handling flow.
- Reuse radiation-field setup allocations across simulation redshift iterations and reduce peak filtering memory usage.
- Keep RadiationFields focused on physical outputs while moving intermediate filtered fields and calculation tables into RadiationFieldsSetup.
- Simplify radiation-field APIs by deriving shell bounds and redshifts from setup state and removing obsolete arguments.

Documentation:
- Update the radiation-fields documentation to reflect the refactored calculation flow.

Tests:
- Update output-structure and radiation-field tests for RadiationFieldsSetup, automatic setup, zero-source behavior, and setup/input redshift validation.

Chores:
- Add dedicated C modules and shared indexing support for radiation-field setup calculations.


src/py21cmfast/drivers/coeval.py
```diff
@@ -773,6 +774,11 @@ def _redshift_loop_generator(
     this_spin_temp = None
     this_halofield = None
     this_radiation_fields = None
+    if inputs.astro_options.USE_TS_FLUCT:
+        this_rad_setup = RadiationFieldsSetup.new(redshift=-1.0, inputs=inputs)
```
steven-murray, line 778: why don't you set dummy=True in this call instead of below?

src/py21cmfast/drivers/single_field.py
```diff
@@ -508,139 +502,135 @@ def interp_halo_boxes(
...
-    zpp_avg: np.ndarray
-        The average redshifts for each shell.
-    z_max: float
-        The maximum redshift for the integration.
-        If a shell crosses this redshift, its contribution to the radiation fields is ignored.
-    need_c: bool
-        Whether to evaluate the 3D boxes via the C code.
-    previous_spin_temp: :class:`~TsBox`
+        The redshift at which to compute the radiation fields.
+    hboxes: Sequence of :class:`~HaloBox` instances
```
steven-murray, line 524: I'm gonna use this opportunity to get you to rename this argument to `halo_boxes`

src/py21cmfast/drivers/single_field.py
```diff
@@ -508,139 +502,135 @@ def interp_halo_boxes(
...
+    rad_setup.setup_shells(inputs=inputs, redshift=redshift)
+
+    # Let's figure out if we really need to go through the C code
+    sfr_allzero = np.all([np.all(hbox.get("halo_sfr") == 0) for hbox in hboxes])
+    lowest_shell_above_zmax = rad_setup.zpp_avg.value.min() >= rad_setup.source_z_max
+    need_c = not (sfr_allzero or lowest_shell_above_zmax)
+
+    if need_c:
+        # TODO: the whole code below is only required if we use mini-halos. It could be removed though, see comment below
+        if inputs.astro_options.USE_MINI_HALOS:
```
steven-murray, line 568: Could we then move this up to the line above, so:

```
if need_c and inputs.astro_options.USE_MINIHALOS:
```

src/py21cmfast/wrapper/outputs.py
```diff
@@ -372,6 +372,7 @@ def get_required_input_arrays(self, input_box: Self) -> list[str]:
 
     def ensure_input_computed(self, input_box: Self, load: bool = False) -> bool:
         """Ensure all the inputs have been computed."""
+        # TODO: This is a hack to avoid having ValueError because the RadiationFieldSetup doesn't have any arrays if mini-halos are not used. This should be fixed in the future.
```
steven-murray, line 375: I don't understand why this comment is here?

src/py21cmfast/wrapper/outputs.py
```diff
@@ -1238,25 +1239,247 @@ def prepare_for_next_snapshot(self, next_z, force: bool = False):
...
+    inverse_val_box = _arrayfield()
+    # arrays for R-dependent prefactors
+    lya_flux_continuum_prefactor = _arrayfield(optional=True)
+    lya_flux_injected_prefactor = _arrayfield(optional=True)
+    lya_flux_continuum_injected_prefactor = _arrayfield(optional=True)
+    lyw_flux_prefactor = _arrayfield(optional=True)
+    lyw_flux_prefactor_MINI = _arrayfield(optional=True)
+    lya_flux_continuum_prefactor_MINI = _arrayfield(optional=True)
+    lya_flux_injected_prefactor_MINI = _arrayfield(optional=True)
+    lya_flux_continuum_injected_prefactor_MINI = _arrayfield(optional=True)
```
steven-murray, line 1277: can we rename this to `_mcg` instead of `_MINI` now? Since they're moving classes anyway

src/py21cmfast/wrapper/outputs.py
```diff
@@ -1238,25 +1239,247 @@ def prepare_for_next_snapshot(self, next_z, force: bool = False):
...
+    # array and floats required for the X-ray optical depth calculation
+    ave_log10_MturnLW = _arrayfield(optional=True)
+    Q_HI_zp: float = attrs.field(default=1.0)
+    x_e_ave_zp: float = attrs.field(default=0.0)
+    # boolean to indicate whether there's enough light
+    NO_LIGHT: bool = attrs.field(default=True)
+    # maximum source redshift for the radiation fields calculation
+    source_z_max: float = attrs.field(default=0.0)
+    # redshifts of the input hboxes
+    hbox_redshifts: list[float] = attrs.field(factory=list)
```
steven-murray, line 1287: and this --> `halo_box_redshifts`

steven-murray (commented):
Thanks @jordanflitter, this really does clean up this part of the code! Only a few minor style and documentation comments I have. I would also love to see one full end-to-end test  that shows how much difference you get due to that approximation of z(r) at lower redshifts.

src/py21cmfast/drivers/coeval.py
jordanflitter: Good question! This is because I call `_init_arrays` in the line that comes after `RadiationFieldsSetup.new()`. If I had called `RadiationFieldsSetup.new(dummy=True)` (or `RadiationFieldsSetup.dummy()`), `_init_arrays` wouldn't have done anything, because it is returns early when `self.dummy=True`.

src/py21cmfast/drivers/single_field.py
jordanflitter: Yes, this is a good suggestion, but there are so many renames I would like to do (this is one of them), that I think it would be better all the renaming to another dedicated PR

src/py21cmfast/drivers/single_field.py
jordanflitter: Yes, of course! (I am not sure why I didn't do it before, probably because I got a bit confused with all the refactoring in this PR)

src/py21cmfast/wrapper/outputs.py
jordanflitter: You are right! I added this comment in an intermediate stage when `RadiationFieldSetup` didn't necessarily have `Array`s in its fields, so I had to modify the logic in that line in a somewhat hacky way. But now `RadiationFieldSetup` always have `Array`s in its fields, so that comment should be removed indeed.

src/py21cmfast/wrapper/outputs.py
jordanflitter: Again, great suggestion, but I'd leave all renamings to future PR

src/py21cmfast/wrapper/outputs.py
jordanflitter: Again, great suggestion, but I'd leave all renamings to future PR

src/py21cmfast/drivers/single_field.py
jordanflitter: Actually, I stand corrected. We must have this annoying structure of
```python
if need_c:
    # TODO: the whole code below is only required if we use mini-halos. It could be removed though, see comment below
    if inputs.astro_options.USE_MINI_HALOS:
        # many lines to compute log10_Mcrit_MCG_ave

    rad_setup.compute(
        redshift=redshift,
        previous_spin_temp=previous_spin_temp,
        allow_already_computed=True,
     )
```
This is because if `USE_MINI_HALOS=False`, we still want to call `rad_setup.compute` if `need_c=True`!

There is a TODO comment that says that we don't really need to compute `log10_Mcrit_MCG_ave` when `setup_radiation_fields` is called. Once that TODO comment is addressed, the structure would be much nicer and will look like
```python
if need_c:
    rad_setup.compute(
        redshift=redshift,
        previous_spin_temp=previous_spin_temp,
        allow_already_computed=True,
     )
```

jordanflitter:
Hi @steven-murray I made two more small commits: the first removes the unnecessary comment you mentioned and the second one fixes the docs (I forgot to mention the `F_H2_SHIELD` parameter that we use for the calculation of the LW flux). If you are okay with leaving the renaming for future PR, I think this one is ready to be merged!

src/py21cmfast/drivers/coeval.py
steven-murray: Ahh, I see. This is a bit hacky then. You need an output struct that is both initialised AND is a dummy?

src/py21cmfast/drivers/single_field.py
steven-murray: Ah, I missed that!

steven-murray (approved):
Let's go!

src/py21cmfast/drivers/coeval.py
jordanflitter: Yes! And it is indeed tricky!

The reason why we need this peculiar combination is because we do in `rad_setup` something that we don't do in the other `OutputStructZ`: we allocate beforehand memory that will be used for the entire simulation, not just for the specific snapshot (this actually replaces the previous hacky `TsInterpArraysInitialised` and `cleanup` logics, but we now implement these logics in python rather than in C). Therefore, we need to do the initialization of `rad_setup` before we call a single-field function (`compute_radiation_fields`). And we also need to define it as a `dummy`, since it used as an input for a single-field function, while its arrays are not computed, and we have an error that is triggered for this specific scenario, unless the input is `dummy`.

Btw, this is somewhat related to an item that we have in #691. We could discuss it in more detail once we get to that issue.

Issue #762, opened 2026-08-31
[BUG] hbox_arr is not computed properly upon resume


DanielaBreitman:
**Describe the bug:**
- It seems like XRSB is really not needed to be written ever and is calculated from the halo catalog. HOWEVER, it seems like it needs to be on disk for is_complete_at to return True and resume. In other words, there is no point in writing it EXCEPT if you want to resume. So I would say it's a bug to require XRSB for resuming. is_complete_at should return True even when the XRSB is not there.
- Suppose we bypass this first bug (e.g. by writing XRSB to disk), then in that case it seems like we get an incomplete halo history when resuming because the `hbox_arr` array starts out empty and does not get populated with the halo catalogues of previous redshifts. This then produces a wrong XRSB and consequently a wrong spin temperature. Here's a more detailed Claude explanation:
is_complete_at(index=idx) only checks completeness at that single redshift, not the whole history below it (loop in caching.py:403-410). So you don't need to keep all XraySourceBoxes — writing just the last one of each batch (highest completed node) is enough to make is_complete_at succeed and resume correctly at the right index.

However, I checked what happens to hbox_arr (the accumulated HaloBox list fed into compute_xray_source_field) if resume actually succeeds (coeval.py L767, L790-L816, L885):

```
hbox_arr = []
...
for iz, z in ...:
    if iz < start_idx:
        continue          # <- skips computing this_halobox AND appending to hbox_arr
    ...
    if z in inputs.node_redshifts:
        hbox_arr += [this_halobox]
```

hbox_arr starts empty every call and is only populated for redshifts actually iterated (iz >= start_idx). If resume jumps straight to start_idx, none of the skipped redshifts' HaloBoxes get reloaded into hbox_arr — there's no cache-reload path for it anywhere in _setup_ics_and_pfs_for_scrolling either. But compute_xray_source_field needs the halo history over a 500 cMpc look-back window (R_MAX_TS = 500.0, default in inputs.py:1564) to build its filtered X-ray shells — that's many node redshifts' worth of HaloBoxes, not just the current one.

**Details:**
 - 21cmFASTv4.2

PR #764, opened 2026-09-01
fix: hbox_arr bug and is_complete_at bug

DanielaBreitman:
Fix for issue #762.

## Summary by Sourcery

Fix coeval-run resumption and cache completeness handling for HaloBox histories and optional XraySourceBox outputs.

Bug Fixes:
- Preserve cached HaloBox history when resuming redshift scrolling so X-ray, spin-temperature, and brightness-temperature results remain correct.
- Correctly restore cached HaloBox data in the resume starting point.
- Allow caches without XraySourceBox files to be considered complete and resumable.

Enhancements:
- Centralize cache completeness and output loading around required fields while treating XraySourceBox as optional.

Tests:
- Add regression coverage for resumed simulations, cached HaloBox restoration, and caches that omit XraySourceBox.


jordanflitter:
Hi @DanielaBreitman thanks for this PR! I looked at the new code in `coeval.py` and it looks like the correct fix to me. The only reason why I am not approving this PR is because I think it would be better if @steven-murray could also give it a look, as he understands better than me the caching mechanism in 21cmFAST.
I also don't understand why the pre-commit failed, but I don't think it has anything to do with your changes.

steven-murray (approved):
Thank you @DanielaBreitman this is great! Note that pre commit is failing... You should set up pre commit on your dev environment and follow it's instructions!

jordanflitter:
@DanielaBreitman I don't know why the pre-commit check fails, but I think it's okay to merge the code now (we usually give the honor of clicking the big green button to the author of the PR).

steven-murray:
@jordanflitter @DanielaBreitman  I disagree! Fix the precommit failures first (literally just do what it tells you to do when you follow the link!). Make sure you set up pre commit on your environment so you can catch these things early as well!

PR #768, opened 2026-09-08
LW bug:

qyx268:
1. LW threshold now uses NU_LW / NU_Lya.
2. LW spectral integration is energy-weighted using a + 2.
3. J21 conversion now includes NU_Lya / (NU_HI − NU_LW).

## Summary by Sourcery

Correct LW spectral integration and intensity normalization to use the appropriate band limits and energy weighting.

Bug Fixes:
- Correct LW band threshold handling by applying the target-frame lower-frequency limit during spectral integration.
- Correct LW intensity conversion to account for the Lyman-alpha to LW-band frequency range.

Enhancements:
- Use energy-weighted spectral integration for LW radiation contributions.

### DIFF
--- a/src/py21cmfast/src/SpinTemperatureBox.c
+++ b/src/py21cmfast/src/SpinTemperatureBox.c
@@ -397,8 +397,8 @@ void calculate_spectral_factors(double zp) {
             if (astro_options_global->USE_MINI_HALOS) {
                 sum_ly2_val_MINI = frecycle(2) * spectral_emissivity(nuprime, 0, 3);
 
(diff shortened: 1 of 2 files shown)


qyx268:
4 tests using minihalo feature failed. please let me know if I should update the testing reference data

steven-murray:
Thanks @qyx268!

I think this all looks reasonable, though it would be good for @jordanflitter to have a look. 

When you update/fix physics that breaks existing tests, I think the best idea is to first make a plot showing the relative difference, mad post that plot in this PR. Then you can regenerate those data files.

PR #769, opened 2026-09-10
Refactor of physical constants into their integral functions

### LINKED ISSUE #705: Make `HaloBox->n_ion` have a physical meaning
The meaning of `HaloBox->n_ion` is quite strange. It holds the multiplication of two numbers.
1. Number of ionizing photons per stellar baryon (coming from both popII and popIII stars). This is what we normally we refer to as `n_ion`.
2. The mean baryon density today, $\rho_b=\Omega_b\rho_\mathrm{crit}$, in units of solar mass per Mpc cubed.

I believe this strange combination was made in order to have all the output fields of `HaloBox` to have some density units, though I am not sure why. This can be seen by the last lines in `move_halo_galprops` (in `map_mass.c`):

```c
for (unsigned long long int i_cell = 0; i_cell < HII_TOT_NUM_PIXELS; i_cell++) {
      boxes->n_ion[i_cell] *= cell_vol_inv;
      boxes->halo_sfr[i_cell] *= cell_vol_inv;
      if (astro_options_global->USE_TS_FLUCT) {
           boxes->halo_xray[i_cell] *= cell_vol_inv;
       }
       if (astro_options_global->INHOMO_RECO) {
           boxes->whalo_sfr[i_cell] *= cell_vol_inv;
       }
       if (astro_options_global->USE_MINI_HALOS) {
           boxes->halo_sfr_mini[i_cell] *= cell_vol_inv;
       }
       if (config_settings.EXTRA_HALOBOX_FIELDS) {
           boxes->halo_mass[i_cell] *= cell_vol_inv;
           boxes->halo_stars[i_cell] *= cell_vol_inv;
           if (astro_options_global->USE_MINI_HALOS) {
               boxes->halo_stars_mini[i_cell] *= cell_vol_inv;
           }
       }
}
```
For example, `halo_sfr` is actually the SFRD, while `halo_mass` and `halo_stars` denote the halo/stellar mass normalized by the cell volume (these fields should be renamed in my opinion...).

Anyway, I suggest to do the following. Instead of having:
```c
// at set_halo_properties (HaloBox.c)
n_ion_sample =
        stellar_mass * consts->pop2_ion * fesc + stellar_mass_mini * consts->pop3_ion * fesc_mini;
// at move_grid_galprops (map_mass.c)
double prefactor_nion = prefactor_stars * consts->fesc_10 * consts->pop2_ion;
double prefactor_nion_mini = prefactor_stars_mini * consts->fesc_7 * consts->pop3_ion;
// at get_uhmf_averages (HaloBox.c)
double prefactor_nion = prefactor_stars * consts->fesc_10 * consts->pop2_ion;
double prefactor_nion_mini = prefactor_stars_mini * consts->fesc_7 * consts->pop3_ion;
// at find_ionised_regions (IonisationBox.c)
if (consts->lagrangian_source_grids)
    curr_fcoll *= 1 / (RHOcrit * cosmo_params_global->OMb * (1 + curr_dens));
```

we should have
```c
// at set_halo_properties (HaloBox.c)
n_ion_sample =
        stellar_mass * consts->pop2_ion * fesc + stellar_mass_mini * consts->pop3_ion * fesc_mini / (RHOcrit * cosmo_params_global->OMb);
// at move_grid_galprops (map_mass.c)
double prefactor_nion = prefactor_stars * consts->fesc_10 * consts->pop2_ion / (RHOcrit * cosmo_params_global->OMb);
double prefactor_nion_mini = prefactor_stars_mini * consts->fesc_7 * consts->pop3_ion / (RHOcrit * cosmo_params_global->OMb);
// at get_uhmf_averages (HaloBox.c)
double prefactor_nion = prefactor_stars * consts->fesc_10 * consts->pop2_ion / (RHOcrit * cosmo_params_global->OMb);
double prefactor_nion_mini = prefactor_stars_mini * consts->fesc_7 * consts->pop3_ion / (RHOcrit * cosmo_params_global->OMb);
// at find_ionised_regions (IonisationBox.c)
if (consts->lagrangian_source_grids)
    curr_fcoll *= 1 / (1 + curr_dens);
```

This would make `HaloBox->n_ion` to be what we normally call `n_ion`, without any weird multiplication by $\rho_b$.

I labeled this as API break, since it is going to change the output values of `HaloBox->n_ion`, but I'm not sure we should wait till v5 to change that. Moreover, if users had been exploring this field, it is very likely they did not know it contains the multiplication by $\rho_b$, so their analysis is wrong if they had relied on this field.

Note: for the older source models that are defined on the Eulerian grid, we also have the fields `unnormalised_nion` and `unnormalised_nion_mini`, which, as the name suggest, contain `n_ion`, but without the constants (`POP2_ION*F_STAR10*F_ESC10` or `POP3_ION*F_STAR7_MINI*F_ESC7_MINI`). Again, users should not guess the factors they need to multiply the output fields in order to get a physically meaningful result. I see no reason why not to have the constants absorbed in those output fields.

jordanflitter:
The purpose of this PR is to clean the C code. Before this PR, functions that perform integration over the halo mass function (e.g. `Nion_General`) involved only the integral part, without the multiplication by physical constants/prefactors (e.g. `RHOcrit * OMm`). As a result, the physical prefactors were scattered across the code in multiple places, thus creating redundant calculations, while making the code less modular and less readable (especially in `get_uhmf_averages` in `HaloBox.c` and `move_grid_galprops` in `map_mass.c`, where the prefactors code was really nasty).

To solve this, this PR moved the multiplication by the constants/prefactors into the integral functions in `hmf.c`. Now, these functions return the physical SFRD/N_ion/X-ray emissivity, while their callers never call functions like `Fcoll_General` (the collapsed fraction, which is the integral part without the constants only for the mass-independent source model). As a result, the C code is now much more readable and modular.
In theory, the code is supposed to be also faster now when interpolation tables are used, because the multiplication by the prefactors is now done at each entry of the interpolation table, instead at every cell.
On top of that, some optimization has been done: When the new flag (see below) `USE_METALLICITY` is False, the X-ray emissivity is proportional to the SFRD. Hence, given the SFRD, the X-ray emissivity integral/interpolation is not performed, and also the CIC algorithm is not performed on the X-ray emissivity.

Few other bonuses:

1. New flag, `USE_METALLICITY`.  Before this PR, the metallicity was taken into account in the X-ray emissivity calculation by setting to True the astro flag `USE_UPPER_STELLAR_TURNOVER`, which was quite a hack. Now, with the new flag, the metallicity can also be applied in the `E-INTEGRAL` source model. It cannot be applied however in the `CONST-ION-EFF` source model as the X-ray emissivity integral assumes that the underlying source model has a mass-dependence (due to the usage of mass-dependent `F_star`, exponential duty fraction, but most importantly, usage of `t_STAR` for the SFR timescale, instead of using the time derivative).
2. The field `n_ion` in `HaloBox` is now really `n_ion` (i.e. the number of ionizing photons that escaped into the IGM, per baryon) without the multiplication by `RHOcrit * OMb`. This solves #705.
3. The struct `IntegralCondition` was removed from `HaloBox.h`, as after moving some of its fields to `ScalingConstants` it became unnecessary.

It is important to stress that this PR is a pure refactor of the code, so the output remains the same as before. Therefore, data files were not needed to be reproduced in order to make the integration tests to pass.


src/py21cmfast/src/HaloBox.c
```diff
@@ -109,178 +90,86 @@ int get_uhmf_averages(double M_min, double M_max, double M_turn_acg, double M_tu
...
+            averages_out->fescweighted_sfr =
+                averages_out->n_ion * RHOcrit * cosmo_params_global->OMb / consts->sfr_timescale;
+        } else {
+            // For the mass-independent source model, the weighted SFRD is proportional to the SFRD
+            // TODO: This is a dead code at the moment, but it's useful to keep it here since in the
+            // future we might want to use
+            //       the weighted SFRD for Eulerian source models as well.
+            //       Note that halo_sfr might not be evaluated in some scenarios.
+            //       Also note that currently in IonisationBox.c, t_STAR is used for the weighted
+            //       SFRD, which I think is a mistake
```
steven-murray, line 169: Is there an issue for this?

src/py21cmfast/src/HaloBox.c
```diff
@@ -467,15 +361,21 @@ void get_cell_integrals(double dens, double l10_mturn_acg, double l10_mturn_mcg,
...
         if (astro_options_global->USE_TS_FLUCT) {
-            properties->stellar_mass *= 1. + dens;
-            properties->halo_xray *= 1. + dens;
+            properties->halo_sfr *= 1. + dens;
             if (astro_options_global->USE_MINI_HALOS) {
-                properties->stellar_mass_mini *= 1. + dens;
+                properties->sfr_mini *= 1. + dens;
+            }
+            if (astro_options_global->USE_METALLICITY) {
+                properties->halo_xray *= 1. + dens;
```
steven-murray, line 369: I don't understand why we only multiply the xray emission by 1+ delta if we're using metallicity?

steven-murray (commented):
This is a great work @jordanflitter. I have only two small comments. To be fair, while the code looked pretty reasonable, I am trusting mostly in the integration tests to prove you did the right thing!

src/py21cmfast/src/HaloBox.c
jordanflitter: There is now! https://github.com/21cmfast/21cmFAST/issues/770

src/py21cmfast/src/HaloBox.c
jordanflitter: Good question! If you look a bit earlier in that function, you'll see that `properties->halo_xray` is also set only when we use metallicity! The reason for this is because of the optimization trick described in the PR's description: if `USE_METALLICITY=False`, then the non-stochastic X-ray emissivity is proportional to the SFRD: `Xray_emissivity = L_X/SFR * SFRD`, where `L_X/SFR` is our astro parameter `L_X` (which has a bad name!). So the relevant multiplication by `1 + dens` for the X-ray emissivity in this case is done for the SFRD, and then at the bottom of `move_grid_galprops` in `map_mass.c`, we set `Xray_emissivity = L_X/SFR * SFRD` if indeed `USE_METALLICITY=False`.

steven-murray: OK :-) I thought it was something like that.

steven-murray (approved):

PR #772, opened 2026-09-17
fix: memory leak in XRSB computation when gc is disabled

DanielaBreitman:
Description: In `compute_xray_source_field()` in `single_field.py`,  the `for i in range(N_STEP_TS)` loop calls `interp_halo_boxes()` once per filtering shell, building a brand-new throwaway `HaloBox` called `hbox_interp` each time, holding full-resolution halo_sfr/halo_xray/halo_sfr_mini arrays (~13.5GB each at HII_DIM=1500 → ~37-40GB per shell, which matches the log below). The code never purged `hbox_interp` after using it, and that wouldn't normally leak.
However, since we're running with `gc.disable()` for the entire job, and `gc.collect()` is only called once after each fully-completed coeval, these 40 disposable ~40GB objects aren't immediately, explicitly freed and just pile up for the entire duration of one redshift's spin-temperature computation. This corresponds to an extra memory overhead of up to ~1.6TB for a single coeval and it currently only gets freed only by my explicit `gc.collect()` after each coeval evaluation.

To reiterate, if users run with garbage collection enabled, this should not be a problem.

Fix is just to explicitly purge `hbox_interp` right after each use inside the shell loop (both the "all-zero SFR, skip shell" branch and the normal compute branch), making disposal explicit instead of relying on any garbage collection pass.

Here's [log showing the memory leak](https://github.com/user-attachments/files/32321361/EOS26_coeval_6_179127326.log) just in case.

## Summary by Sourcery

Bug Fixes:
- Prevent excessive memory accumulation during XRSB computations when garbage collection is disabled by explicitly releasing interpolated halo boxes after each filtering-shell computation.


steven-murray (approved):
Seems correct to me

Issue #565, opened 2025-09-04
Conveniently get arrays from OutputStruct after they had been purged (but still on disk)


jordanflitter:
I ran the following script:

```python
import py21cmfast as p21c
import numpy as np
from tempfile import mkdtemp

# Define inputs
inputs = p21c.InputParameters.from_template(
        name = 'latest-discrete',
        random_seed = 1
    ).with_logspaced_redshifts(
        zmin = 10.0,
        zmax = 35.0,
        zstep_factor = 1.02,
).evolve_input_structs(HII_DIM = 35, N_THREADS = 6.)

# Define lightconer
lightconer = p21c.RectilinearLightconer.between_redshifts(
    min_redshift=inputs.node_redshifts[-1]+0.5,
    max_redshift=inputs.node_redshifts[0]-0.5,
    resolution=inputs.simulation_options.cell_size,
    cosmo=inputs.cosmo_params.cosmo
)

# Define cache path
cache = p21c.OutputCache(mkdtemp())

# Define the redshifts of the output coeval boxes
out_redshifts = [20,]

# These lines find the closest redshifts to out_redshifts in node_redshifts
out_redshifts_node = []
node_redshifts_array = np.array(inputs.node_redshifts)
for out_redshift in out_redshifts:
    output_redshift = node_redshifts_array[np.argmin(np.abs(out_redshift - node_redshifts_array))]
    out_redshifts_node.append(output_redshift)

# Generate lightcone and coeval boxes
gen = p21c.generate_lightcone(
    lightconer = lightconer,
    inputs = inputs,
    cache = cache,
    write = p21c.CacheConfig.last_step_only(),
    progressbar = True,
)

# Append output coeval boxes
coevals = []
for iz, redshift, coeval, lightcone in gen:
    if redshift in out_redshifts_node:
        coevals.append(coeval)

# Print
coeval = coevals[0]
print(f"Mean of spin temperature: {coeval.ts_box.spin_temperature.value.mean()}")
print(f"Mean of SFRD: {coeval.halobox.halo_sfr.value.mean()}")
```

and I got the following output:

```
Mean of spin temperature: 56.73775100708008
Traceback (most recent call last):
  File "/Users/jordanflitter/Desktop/Postdoc/2.Research/Projects/Debug_script.py", line 223, in <module>
    print(f"Mean of SFRD: {coeval.halobox.halo_sfr.value.mean()}")
[...]

PR #773, opened 2026-09-18
Transparently load purged arrays on direct access

steven-murray:
## Summary

Closes discussion in #565. Direct attribute access on an `OutputStruct` field (e.g. `box.field.mean()`) now transparently loads the value from disk if it was purged, instead of requiring `.get()` — this is the behavior users kept expecting when hitting confusing `AttributeError`s after `purge()`.

- `Array` now supports the numpy array protocol (`np.asarray(...)`) and delegates unknown attribute access (`.mean()`, `.sum()`, etc.) to its resolved value, loading from disk if needed.
- By default, the loaded value is cached directly onto the `Array` instance so repeated access doesn't keep re-reading from disk (mirroring what `.get()` already does). This can be disabled via the new `CACHE_ARRAYS_ON_ACCESS` config option for callers who want strict control over when memory gets reallocated.
- Plain inspection (`repr`, `hasattr`, `copy.deepcopy`, `pickle`) never triggers a disk read or raises a surprising error.
- No API break: `struct.field` still returns an `Array`, not a bare `ndarray`; `.get()`/`.value`/`.state` are all unchanged.
- Fixed a pre-existing bug in `Array.purged_to_disk()` that left `state.computed_in_mem=True` after dropping the in-memory value (previously unused elsewhere in the codebase).
- Added a "Managing memory" section to the `caching.ipynb` tutorial documenting the tradeoff and the opt-out.

See the design discussion in [#565](https://github.com/21cmfast/21cmFAST/issues/565#issuecomment-5728949731) for the rationale, including why a simpler `__getattr__`-on-`OutputStruct` approach doesn't work.

## Test plan

- [x] New unit tests in `tests/test_arrays.py` covering duck-typing, uncomputed-array error semantics, transient vs. cached loading, the `CACHE_ARRAYS_ON_ACCESS` config flag, and pickle/deepcopy/`hasattr` safety.
- [x] New integration tests in `tests/test_output_structs.py` against a real disk-cached `InitialConditions` object.
- [x] Full non-slow test suite passes (`pytest tests/ -k "not slow"`, excluding modules with unrelated missing optional dev dependencies in this environment).
- [x] `caching.ipynb` tutorial re-executed end-to-end with real output.

## Summary by Sourcery

Allow direct access to purged array values while preserving Array objects and providing safe, configurable loading behavior.

New Features:
- Enable direct NumPy and ndarray-style access to purged arrays with transparent disk loading.
- Add configurable caching of values loaded through direct array access, enabled by default and opt-outable via CACHE_ARRAYS_ON_ACCESS.

Bug Fixes:
- Correct purged array state so in-memory computation is no longer reported after its value has been removed.

Enhancements:
- Make inspection, copying, and serialization of uncomputed or purged arrays safe without triggering disk loads.

Documentation:
- Document direct array access caching and memory-management tradeoffs in the caching tutorial.

Tests:
- Add unit and integration coverage for array duck typing, transparent loading, caching configuration, state handling, and inspection safety.


jordanflitter (approved):
Thanks Steven! (and Claude!)

Looks really great! I like the new transparent usage (`ionized_box.neutral_fraction.mean` is much smoother than `ionized_box.neutral_fraction.value.mean` or `ionized_box.get("neutral_fraction").mean`). And I also like the new `test_arrays.py` (I'm surprised that we didn't have it until now).

I have been thinking whether we really want/expect the users to know about the state of the arrays, if they are on RAM or not. While I believe that the vast majority of users really don't care about this detail ("just give me the numpy array!"), I don't think it's harmful to have the new section in the tutorial, and the extra functionality with `CACHE_ARRAYS_ON_ACCESS`. 

There's one thing that I think that is still missing though, and it's to describe the `get` method and what it is good for. The new section in the tutorial mentions it several times, but I don't think it describes in first place what it does!

Apart that comment on 'get` in tutorial, everything looks great, so I'm approving this :)

PR #774, opened 2026-09-18
feat: default summary plots for simulation outputs

### LINKED ISSUE #257: [Feature Req.] Option for CLI to output basic plots

Running the CLI produces a lightcone/coeval dataset, but no plots or anything tangible to look at. This can be a bit disconcerting, since by default the outputs are in the cache, which is kinda hidden, which means it might seem like the call did nothing at all. Outputting at least a lightcone/coeval default plot (and maybe xHI/global signal if appropriate) would at least assure the user that things ran OK, and give them a way to check what their options look like.

Just a new `--plot` option, potentially with string options that it can take, specifying different plots to make.

An alternative would be to have another CLI command `21cmfast plot` which can be pointed at a particular lightcone in the cache. At the end of the `21cmfast lightcone` command output, it could say something like "Use `21cmfast plot <file>` to get plots!"

steven-murray:
Closes #257.

Running the CLI produces data in a (fairly hidden) cache, with nothing tangible to look at. This adds simple default summary plots — enough to confirm the run worked and see roughly what the model does, without trying to anticipate the real plots a user will need.

## What the plots look like

All three were made with `21cmfast plot <file>` on a 92 Mpc, `HII_DIM=64` run of the `latest` template (spin temperature + inhomogeneous recombinations), except the global evolution which uses `EOS21`. They're also committed to `docs/images/summary_plots/` and shown in the CLI docs.

**Lightcone** — a slice through the lightcone, with the global panels beneath it plotted against the *same line-of-sight distance* axis, so features line up by eye:

Lightcone summary plot

**Coeval** — a slice through the box for each of the main fields:

Coeval summary plot

**Global evolution** — global signal (with a frequency axis), ionization history, temperature history:

Global evolution summary plot

## CLI

A new `21cmfast plot` command, pointed at any primary output file:

```bash
21cmfast plot lightcone.h5
```

It detects the file kind (coeval / lightcone / global evolution) and writes `lightcone_summary.png` next to the data. `--out` puts it elsewhere, `--show` opens it interactively.

`run coeval`, `run lightcone` and `run global` also take `--plot` to write the summary as part of the run. When `--plot` isn't given, the run now ends by telling the user how to make plots later:

> 📊 Use **`21cmfast plot lightcone.h5`** to get summary plots of this simulation!

## Plotting API

New public functions in `py21cmfast.plotting`:

- `plot_global_signal` — the global 21-cm signal vs redshift, with a secondary frequency axis (tick locations are chosen to be round numbers that are *also* well-separated in redshift; the naive `secondary_xaxis` ticks bunch up badly at low z).
- `plot_global_evolution` — multi-panel global signal / ionization history (`x_HI`, `x_e`) / temperature history (`T_S`, `T_K`, `T_CMB`). Panels with no data are dropped, so it works for both `LightCone` and `GlobalEvolution` objects regardless of which modules were run.
- `coeval_summary_plot` — a row of slices through the box for each of the main fields.
- `lightcone_summary_plot` — as above.
- `summary_plot` — dispatches on object type.
- `get_global_quantities` — small helper that pulls `(redshifts, quantities)` off either a `LightCone` or a `GlobalEvolution`.

`plot_global_history` now also works with `GlobalEvolution` objects, and `coeval_sliceplot` puts the colorbar label on the right axis when the colorbar is horizontal.

### Judgement calls worth a look

- **Temperature slices are log-scaled and clipped.** `spin_temperature` and `kinetic_temp_neutral` slices are dominated by a handful of very hot cells near sources — in the example above the box spans 106–15800 K but 99% of cells are below 300 K, so on a linear scale the panel is uniformly dark. They're plotted on a log scale clipped to the 1st–99th percentile of the slice. That does mean the extremes are saturated; `log=False` or explicit `vmin`/`vmax` turns it off.
- **`brightness_temp` keeps the existing fixed `(-150, 30)` mK EoR normalization**, so the color scale is comparable between runs rather than auto-scaled.
- **The lightcone panel has a minimum height** (1.1") even when the true aspect ratio would make it thinner, because the axis labels need somewhere to live.

## Notes

- Tested end-to-end by running coeval/lightcone/global simulations through the CLI and eyeballing the output; 5 new CLI tests and 14 new plotting tests.
- `run ics` deliberately has no `--plot`; use `coeval_sliceplot` directly for initial conditions.
- jordanflitter's suggestion on the issue (calling `global_evolution` from the CLI to plot global quantities rather than saving a `GlobalEvolution` object) is a larger refactor and isn't attempted here — `run global --plot` gets the plots without changing what's saved.

## Summary by Sourcery

Add default simulation summary plots and expose them through the CLI and plotting API.

New Features:
- Add a `21cmfast plot` CLI command that automatically detects saved coeval, lightcone, and global-evolution outputs and writes summary images.
- Add optional `--plot` support to coeval, lightcone, and global simulation commands, with guidance for creating plots after runs.
- Provide public plotting APIs for global signals and histories, coeval summaries, lightcone summaries, and type-based summary dispatch.

Bug Fixes:
- Make global-history plotting compatible with both `LightCone` and `GlobalEvolution` objects.
- Correct horizontal coeval colorbars to display their labels on the appropriate axis.

Enhancements:
- Add default visualizations for key simulation fields, including frequency-aware global histories and robust scaling for temperature slices.
- Align lightcone global-history panels with the same line-of-sight distance axis for easier visual comparison.

Documentation:
- Document the new CLI plotting workflow and include example summary plots in the CLI documentation.

Tests:
- Add CLI and plotting coverage for output detection, plot generation, plot options, global-evolution support, field scaling, and error handling.


docs/tutorials/cli_usage.rst
```diff
@@ -355,6 +355,51 @@ for example::
     ``node_redshifts``.
 
 
+Making Plots
+------------
+
+A simulation that writes everything into a (rather hidden) cache directory can feel a
+bit like it did nothing at all. To get a quick look at what you just made, use::
```
steven-murray:
```suggestion
To get a quick look at a simulation you just made (whether a lightcone, coeval or global signal), use::
```

src/py21cmfast/cli.py
```diff
@@ -719,6 +741,91 @@ def global_evolution(
...
+        ("coeval", Coeval),
+        ("global_evolution", GlobalEvolution),
+    ):
+        if markers[marker]:
+            return cls.from_file(path)
+
+    raise ValueError(
+        f"The file {path} is not a recognized 21cmFAST output file "
+        "(expected a coeval, lightcone or global-evolution file)."
+    )
```
steven-murray: Are you sure there is not such a function already existing in the `io` sub-module? If there is not, we should move this functionality to that module and import it from there. I think the function name should be something like `load_simulation` or `load_high_level_simulation` (to distinguish it from loading the lower-level cache files)

steven-murray:
We should add a flag to the run scripts to *show* the plot rather than write it to file (or both).

docs/tutorials/cli_usage.rst
steven-murray: Applied, thanks.

src/py21cmfast/cli.py
steven-murray: Good catch — there wasn't an existing one, so I've moved it to `py21cmfast.io.h5` as `load_high_level_simulation` (exported from both `py21cmfast` and `py21cmfast.io`). Went with the longer name for exactly the reason you gave: it sits right next to `read_output_struct`, which reads the lower-level cache files, so the distinction is worth spelling out. The CLI now just imports it, and it's tested in `tests/io/test_h5.py` rather than via the CLI.

steven-murray:
Added `--show` to `run coeval`, `run lightcone` and `run global` (16e58712).

It's independent of `--plot`, so all three combinations work:

```bash
21cmfast run lightcone --param-file custom.toml --redshift-range 6 12 --plot   # write only
21cmfast run coeval --param-file custom.toml -z 8 --show                       # show only
21cmfast run global --param-file custom.toml --plot --show                     # both
```

The "use `21cmfast plot <file>`" hint at the end of a run is now suppressed when either flag is given, not just `--plot`.

Same commit also covers your other two comments — see the inline replies.

steven-murray:
Merged `release-v4.3` in (this PR was still based on `main`, which is well behind the base branch after the #775 rename). Clean merge, and all 141 tests in `test_plotting.py`, `test_cli.py` and `tests/io/test_h5.py` still pass — none of the renamed fields are ones the plotting code uses.

``75f0e9d7`` also clears six ruff errors that were failing pre-commit.ci. **All six are on `release-v4.3`, not from this PR** — the three files involved are byte-identical to the base branch. pre-commit.ci is green again now. Happy to split that commit into its own PR if you'd rather keep this one to plotting.

One of them was a real bug rather than a style nit: in `_redshift_loop_generator`, the resume path still said `hbox_arr.append(...)` and read a `"HaloBox"` struct from the cache, both renamed everywhere else — so any resumed run with a Lagrangian source grid would hit a `NameError`.

⚠️ **That path is still broken after my fix**, and needs whoever did the rename. `test_coeval_resume_reconstructs_hbox_history` was failing on the base branch too (it constructs `CacheConfig(xray_source_box=False)`, renamed to `radiation_fields`). With that kwarg corrected it gets further and then fails with:

```
TypeError: radiation_fields should be of type RadiationFields, got <class 'NoneType'>
  src/py21cmfast/drivers/coeval.py:899 in _redshift_loop_generator
```

The resume path needs to rebuild the radiation fields, not just the emissivity fields, and I didn't want to guess at the intended behaviour.

steven-murray:
@jordanflitter your main suggestion has been applied here (allowing `--show`) to bring the plot up interactively.

However, it looks like we got the merge from main a little wrong, and I tried to get Claude to fix things, but it is stuck on one particular case. Maybe you could fix this most easily (see above comment from "me").

PR #775, opened 2026-09-19
Rename for better readability

jordanflitter:
This PR renames many variables/functions/classes in order for their names to reflect better what they really stand for, and make the code a lot more readable.

- The term “halo box” was renamed with the term “emissivity fields”:
   - `HaloBox` -> `EmissivityFields`
   - `compute_halo_grid` -> `compute_emissivity_fields`
   - `EXTRA_HALOBOX_FIELDS` -> `EXTRA_EMISSIVITY_FIELDS`
   - `halobox`, `halo_box`, hbox` etc -> `emissivity_fields`

- The term “mini halo” was replaced with “mcgs”:
  - `USE_MINI_HALOS` -> `USE_MCGS`
  - `INTEGRATION_METHOD_MINI` -> `INTEGRATION_METHOD_MCGS` (as well as `INTEGRATION_METHOD_ATOMIC` -> `INTEGRATION_METHOD_ACGS`)
  - `F_STAR7_MINI` -> `F_STAR7_MCG` (as well as `F_STAR10` ->  `F_STAR10_ACG`)
  - `F_ESC7_MINI` -> `F_ESC7_MCG` (as well as `F_ESC10` ->  `F_ESC10_ACG`)
  - `ALPHA_STAR_MINI` -> `ALPHA_STAR_MCG` (as well as `ALPHA_STAR` ->  `ALPHA_STAR_ACG`)
  - `L_X_MINI` -> `LX_OVER_SFR_MCG` (as well as `L_X` -> `LX_OVER_SFR_ACG`)
  - In addition, all the `_mini` (or `_MINI`) variables in the code were renamed with `_mcg` suffix, while their non-mini partners were renamed with `_acg` suffix.

- The term “fcoll” or “f_coll” was replaced with “n_ion” in the context of reionization:
  - Some fields in `IonizedBox` were renamed:
     - `mean_f_coll` -> `nion_unconditional_acg`
     - `mean_f_coll_MINI` -> `nion_unconditional_mcg`
     - `log10_Mturnover_ave` -> `log10_mturn_ave_acg`
     - `log10_Mturnover_MINI_ave` -> `log10_mturn_ave_mcg`
     - `unnormalised_nion` -> `nion_conditional_filtered_acg` (now the prefactors are included in the output!)
     - `unnormalised_nion_mini` -> `nion_conditional_filtered_mcg` (now the prefactors are included in the output!)
[...]

PR #776, opened 2026-09-19
CONST-ION-EFF photoionisation rate keeps dfcoll/dt, not t_STAR

### LINKED ISSUE #770: Inconsistent `ionisation_rate_G12` for `SOURCE_MODEL="CONST-ION-EFF"`
There seems to be some inconsistency in how the photoionization rate is computed in `IonisationBox.c` for the mass-independent source model (`"CONST-ION-EFF"`), this is what we call `ionisation_rate_G12` in the code.

Note that in `IonisationBox.c` we always use `astro_params_global->t_STAR * t_hubble(redshift)` in setting `gamma_prefactor` for the Eulerian source models, including the mass-independent source model:
```c
if (consts->lagrangian_source_grids)
        consts->gamma_prefactor /= RHOcrit * cosmo_params_global->OMb;
else
        consts->gamma_prefactor = consts->gamma_prefactor / (sc.t_h * sc.t_star);
```
However, for the mass-independent source model, that seems to be a mistake because:
1. We don't use `t_STAR` is computing the SFRD for the mass-independent source model, instead the SFRD timescale is obtained by taking the time-derivative of the collapsed fraction, see Eq. (13) in https://arxiv.org/pdf/1003.3878.
2. The relevant paper in the literature was written when there was only a mass-independent source model, and that paper indeed says that the photoionization rate is proportional to the time-derivative of the collapsed fraction, without `t_STAR` never being mentioned, see Eq. (4) in https://arxiv.org/pdf/1402.2298.

shaneraphel:
`set_ionbox_constants` leftover divides every Eulerian `gamma_prefactor` by `t_h * t_star`. That factor is licensed for E-INTEGRAL, whose SFRD is stars / (t_STAR t_hubble). CONST-ION-EFF is the mass-independent Eulerian model (`source_model_is_mass_dependent` is false; leftover already stores that bit as `mass_dep_zeta`). Its SFRD is dfcoll/dt (Zahn et al. 2011 eq.13; Sobacchi & Mesinger 2014 eq.4) and never multiplies by t_STAR.

The mass-independent Eulerian arm now divides by Hubble time only. Lagrangian RHOcrit*OMb and E-INTEGRAL t_STAR stay. Gamma is not zeroed.

Fixes #770

Reproducible dest sha256 00b1707ec6a35fe10dfd0b9268f74dee7f96d35773f291f23802c7ec006acf4a (690 B) against 21cmFAST 40c5fcb42f7f. git apply is clean. Zero SGD; n_parameters=0.

## Summary by Sourcery

Bug Fixes:
- Correct the CONST-ION-EFF photoionisation-rate normalization to use the collapse-fraction growth rate rather than the stellar timescale.


jordanflitter:
Hi @shaneraphel, thanks for addressing the issue in #770. 

Unfortunately, the solution to this issue is more complicated than changing one line. This is because, as you said, the photoionization rate in the `CONST-ION-EFF` source model is proportional to `dfcoll/dz`. However, the change that you made only strips away the timescale from the photoionization calculation, so the computed quantity has the wrong unit in the mass-independent source model!

The correct solution should be the following.
1. First, express the photoionization rate as a time-derivative of `n_ion`, not as the time-derivative of the collapsed fraction.
2. For the mass-dependent source model, the time derivative of `n_ion` is approximated by dividing `n_ion` with `t_STAR * t_hubble(z)`, this what gives the correct units for the photoionization rate. For the mass-independent source model however, this is evaluated like the following: if the spatially varying `n_ion` is `zeta * f_coll`, then `dn_ion/dt = zeta * dfcoll/dt = zeta * dz/dt * dfcoll/dz`. This could be computed semi-analytically by evaluating the spatially varying part of `dfcoll/dz` with EPS, and then normalizing by the theoretical value from ST (or whatever HMF the user chose to work with) over the numerical mean of the `dfcoll/dz_eps` box.
3. Note that unlike the mass-dependent source model, for `CONST-ION-EFF` we will have to allocate two boxes, one for `n_ion`, and one for `dn_ion/dt`, since now these quantities are no longer connected by a simple scalar multiplication.

This solution is certainly doable, but I think that it should be done not from the main/v4.2 branch, but rather from the v4.3 branch, especially from the `rename` branch (see #775) that is soon going to be merged into v4.3. The advantage of doing it from v4.3 (once the `rename` branch is merged):
1. The relevant functions have been renamed properly to reflect exactly what they do, so it becomes easier for a human reviewer to read/edit the code and understand exactly what's going on.
2. This would also make it easier to solve future conflicts in merging the fix into v4.3, since v4.3 includes major refactoring and renaming of the code.

If you are still interested in solving #770, I suggest you to remove this PR, and make a new one that is based on v4.3, after the `rename` branch had been merged (this should happen sometime next week). Then, please follow the above guidelines to implement the correct solution.

shaneraphel:
Thanks. Closing this PR. Dividing out t_STAR leaves CONST-ION-EFF gamma with the wrong unit; the rate has to be written as dn_ion/dt = zeta * dz/dt * dfcoll/dz (EPS box, HMF-normalized), with a second box, on v4.3 after #775.

